PR #38, opened 2022-05-18
BHBC-1739: Push/search data to/from elastic search

anissa-agahchen:
# Overview

## Links to Jira tickets

- https://quartech.atlassian.net/browse/BHBC-1738

- Ability to send a JSON Object to Elastic Search
- Ability to do a simple test search, and have the system return all fields and all records that contain the term
- New and updated tests


docker-compose.yml
```diff
@@ -301,7 +300,7 @@ services:
     depends_on:
       db:
         condition: service_healthy
-    command: ["npm", "run", "migrate-rollback"]
+    command: ["npm", "run", "migrate:rollback"]
```
NickPhura: These two migrate commands will need to be reverted back, as they were changed in backbone (sims uses a colon, backbone uses a dash) (Will eventually update sims to use a dash too as its more consistent with other similar commands).

api/src/paths/dwc/elastic-search.ts
```diff
@@ -0,0 +1,137 @@
+import { RequestHandler } from 'express';
```
NickPhura: (Not his line specifically)

I would change the name of this file to not include `elastic`, as our use of elastic search isn't a detail the outside world needs to know about.

NickPhura: They just care what the endpoint will do for them (performing some kind of search function on dwc data).

api/openshift/api.dc.yaml
```diff
@@ -172,10 +174,10 @@ objects:
                       name: ${DB_SERVICE_NAME}
                 - name: DB_PORT
                   value: '5432'
+                - name: KEYCLOAK_URL
+                  value: ${CERTIFICATE_URL}
```
NickPhura: `CERTIFICATE_URL` isnt used anymore (or shouldnt be)

api/src/paths/dwc/eml/search.ts
```diff
@@ -0,0 +1,137 @@
...
+import { ESService } from '../../../services/es-service';
+
+const defaultLog = getLogger('paths/dwc/eml/search');
+
+export const GET: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          validSystemRoles: [PROJECT_ROLE.PROJECT_LEAD, PROJECT_ROLE.PROJECT_EDITOR],
```
NickPhura: Can delete this line, as the `discriminator: 'SystemUser'` is all we need right now

api/src/paths/dwc/eml/search.ts
```diff
@@ -0,0 +1,107 @@
...
+  ],
+  responses: {
+    200: {
+      description: 'Submission search response object.',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'array',
+            items: {
+              type: 'object'
```
NickPhura: We'll need to define this further, though for now we can probably wait since we don't really know what all of the inputs/outputs will be for searching EML

anissa-agahchen: Yeah for sure .. for now, it's just set up for project names.

api/src/services/dwc-service.test.ts
```diff
@@ -4,6 +4,8 @@ import chai, { expect } from 'chai';
 import { describe } from 'mocha';
 import sinon from 'sinon';
 import sinonChai from 'sinon-chai';
+import { ESService } from '../services/es-service';
+// import { getRequestHandlerMocks } from '../__mocks__/db';
```
NickPhura: Remove test import

api/src/services/dwc-service.ts
```diff
@@ -130,6 +132,78 @@ export class DarwinCoreService extends DBService {
     return { dataPackageId, submissionId };
   }
 
+  /**
+   *
```
NickPhura: Dont forget to add a description here

NickPhura: Same for the convert eml function below

NickPhura (approved):

PR #39, opened 2022-05-18
BHBC-1770: Make list submissions endpoint

curtisupshall:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1770


api/src/paths/dwc/dataset/list.test
```diff
@@ -0,0 +1,308 @@
+
```
NickPhura: This file is missing its `.ts` extension

curtisupshall: I did this intentionally, as the tests aren't finished and I wanted the rest of the test suite to run. Once the tests are done I'll fix this

NickPhura: gotcha no worries

curtisupshall:
Code is finished, just adding tests; Reviews welcome.

api/src/services/submission-service.ts
```diff
@@ -100,4 +101,28 @@ export class SubmissionService extends DBService {
...
+  }
+
+  /**
+   * Get a submission record's S3 key by its id.
+   *
+   * @param {number} submissionId
+   * @return {*}  {Promise<string | null>}
+   * @memberof SubmissionService
+   */
+   async getSubmissionRecordS3Key(submissionId: number): Promise<string | null> {
```
KjartanE: Would this function ever return null? (submission?.input_key || null ) has the potential to return null but the submissionRepo will only ever return an actual value. It will throw an error if no row is found.

api/src/repositories/submission-repository.ts
```diff
@@ -333,4 +337,39 @@ export class SubmissionRepository extends BaseRepository {
...
+  /**
+   * Fetch a submission record by primary id.
+   *
+   * @param {number} submissionId
+   * @return {*}  {Promise<ISubmissionModel>}
+   * @memberof SubmissionRepository
+   */
+   async listSubmissionRecords(): Promise<ISubmissionModelWithStatus[]> {
+    const sqlStatement = SQL`
+      SELECT
```
KjartanE, line 399: If this SQL works then all good! But maybe there is an interest in using knex for more complicated sql so its more human readable. I trust that this works but I cant follow in my head

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
```diff
@@ -0,0 +1,77 @@
...
+];
+
+GET.apiDoc = {
+  description: 'Retrieve the signed URL of a submission.',
+  tags: ['dwc'],
+  security: [
+    {
+      Bearer: []
+    }
+  ],
```
KjartanE, line 33: If you are requiring submissionId in the params, don't forget to write that in the schema. Checkout `parameters: [
    {
      in: 'path',
      name: 'submissionId',
      schema: {
        type: 'integer',
        minimum: 0
      },
      required: true
    }
  ],`

api/src/repositories/submission-repository.ts
```diff
@@ -333,4 +337,39 @@ export class SubmissionRepository extends BaseRepository {
...
+        ON
+          ss.submission_status_type_id = sst.submission_status_type_id 
+        ORDER BY
+          ss.submission_id, ss.submission_status_id DESC) t1
+      ON
+        t1.submission_id = s.submission_id;
+    `;
+
+    const response = await this.connection.sql<ISubmissionModelWithStatus>(sqlStatement);
+
```
KjartanE, line 421: We have been following same function guide on most repo functions. Check other functions in file. Should check the response if any rows are present and throw error if empty.

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
```diff
@@ -0,0 +1,77 @@
...
+  return async (req, res) => {
+    const connection = getDBConnection(req['keycloak_token']);
+    const submissionId = Number(req.params.submissionId);
+
+    try {
+      await connection.open();
+      
+      const submissionService = new SubmissionService(connection);
+      
+      const s3Key: string | null = await submissionService.getSubmissionRecordS3Key(submissionId);
```
KjartanE: No error check required here if empty return is handled in the repo file.

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (!s3Key) {
+        throw new HTTP500('Failed to find submission S3 key.');
+      }
+
+      const signedS3Url: string | null = await getS3SignedURL(s3Key)
```
KjartanE: Should not be null

api/src/paths/dwc/dataset/list.ts
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  responses: {
+    200: {
+      description: 'Darwin Core data packages.',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'array',
+            items: {
+              //
```
KjartanE: Schema update?

NickPhura: Dont forget to fill this in

api/src/services/submission-service.ts
curtisupshall: Method deleted.

api/src/paths/dwc/dataset/list.ts
curtisupshall: Added.

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
curtisupshall: Added!

api/src/repositories/submission-repository.ts
curtisupshall: Since this method is just listing submissions, it's quite possible that it will return an empty array if there aren't any submissions in the database. Arguably, returning `[]` and not throwing an error is expected behaviour in this case.

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
curtisupshall: Method removed.

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
curtisupshall: `AWS.S3` returns `string | null`, so the helper function `getS3SignedURL` returns `string | null`. Refactored to make an explicit check.

api/src/repositories/submission-repository.ts
KjartanE: What about the case where rows doesn't exist on return from sql?

api/src/paths/dwc/dataset/{submissionId}/getSignedUrl.ts
KjartanE: looks good added testing for new possible error

api/src/repositories/submission-repository.ts
curtisupshall: In the event that the `submission` table is empty, the expected value of `response.rows` is `[]`. Since this is congruent with what we can expect the method to return in when said table is empty, it's safe to simply return `response.rows`. Are you suggesting that the value of `repsonse.rows` could be something other than `[]` such as `undefined`?

NickPhura: Just catching up on this thread.  

I don't believe `rows` can ever be null or undefined, so checking for an array isn't needed. 

`rowCount` could be 0 if there is no status records, but since that is okay, checking for `!response.rowCount` isn't needed.  

So I think this whole if statement can be removed.

NickPhura: If something catastrophic occurs it will throw an exception

api/src/repositories/submission-repository.test.ts
```diff
@@ -313,4 +313,43 @@ describe('SubmissionRepository', () => {
...
+        update_user: 15,
+        revision_count: 1
+      };
+      const mockQueryResponse = ({ rowCount: 1, rows: [mockResponse] } as any) as Promise<QueryResult<any>>;
+
+      const mockDBConnection = getMockDBConnection({
+        sql: async () => {
+          return mockQueryResponse;
+        }
+      });
```
NickPhura: (Not actually a change thats needed) This could be simplified to 
```typescript
const mockDBConnection = getMockDBConnection({ sql: async () => mockQueryResponse });
```

NickPhura: Maybe something we can do as a separate PR.  I want to introduce the import formatter package once all of the current PR's are merged, so maybe as part of that one.

api/src/repositories/submission-repository.ts
```diff
@@ -46,6 +46,10 @@ export interface ISubmissionModel {
   revision_count: number;
 }
 
+export interface ISubmissionModelWithStatus extends ISubmissionModel {
```
anissa-agahchen, line 45: Is the reason for extending ISubmission model that the status comes from a different table?

anissa-agahchen (dismissed):

NickPhura (approved):

api/src/repositories/submission-repository.ts
NickPhura: Yeah, I think that function is returning all of the submission rows + the 1 status row from the other table

curtisupshall: That is correct

anissa-agahchen (approved):

PR #44, opened 2022-05-30
BHBC-1641: Show dataset in a list view in BioHub App (POC)

curtisupshall:
# Overview

Show dataset in a list view in BioHub App (POC)

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1641

- GIVEN that datasets exists in the backbone
- WHEN a user accesses BioHub App
- THEN they should see the datasets in a list view


NickPhura:
Everything looks good. I do think we should add a bit more to this ticket though to cover the bit where we finds results based on a query to elastic search (which the ticket didnt originally describe very well).  

Lets chat in Discord about it.  

(None of the work done in this ticket so far will be scrapped, as it will all be useful eventually. There is just a step we need to add before raw file downloads are available that needs to be filled in)

app/src/hooks/api/useSearchApi.ts
```diff
@@ -35,9 +35,20 @@ const useSearchApi = (axios: AxiosInstance) => {
     return data;
   };
 
+  const listAllDatasets = async (): Promise<ISearchResponse<{ datasetTitle: string[] }>> => {
```
NickPhura, line 38: Could you add a test for this function

NickPhura (changes requested):
Everything looks good in general.

Just noticed 1 issue: The call to get eml search results on the DatasetsPage is running repeatedly in a loop. Will need to add a check to the useEffect to prevent it from running more than once.

NickPhura:
Come bug me before you work on that one issue, I have some data loading hooks I made that we could try (hadn't gotten around to pulling them in yet).

app/src/hooks/useDataLoader.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  callback: (() => any) | null | undefined;
+  
+}
+
+const TestComponent: React.FC<ITestComponentProps> = (props) => {
+
+  return <></>;
+};
+
+describe('useDataLoader', () => {
```
NickPhura: We can also tackle these tests in a separate branch, and then we can get this PR merged. Up to you either way, as we have lots of time left in the sprint still.

NickPhura (approved):

PR #46, opened 2022-05-31
BHBC-1788: Implement xml->json transform for EML

anissa-agahchen:
# Overview

## Links to Jira tickets

- https://quartech.atlassian.net/browse/BHBC-1788

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


app/src/components/fields/AutocompleteField.tsx
```diff
@@ -30,7 +30,7 @@ const AutocompleteField: React.FC<IAutocompleteField<string | number>> = <T exte
     const result = props.options.find((option) => existingValue === option.value);
 
     if (!result) {
-      return null as unknown as IAutocompleteFieldOption<T>;
+      return (null as unknown) as IAutocompleteFieldOption<T>;
```
NickPhura: (not this line specifically) It looks like some formatting is changing here and in some other files in the app. I believe the most recent version of the prettier formatter actually removes these brackets, so it might be that you need to run `make install` to ensure you have the latest version and then re-run the formatter, as I think there should be no formatting related changes here.

api/src/services/submission-service.ts
```diff
@@ -9,8 +11,12 @@ import {
   SUBMISSION_MESSAGE_TYPE,
   SUBMISSION_STATUS_TYPE
 } from '../repositories/submission-repository';
+import { getFileFromS3 } from '../utils/file-utils';
+import { getLogger } from '../utils/logger';
 import { DBService } from './db-service';
 
+const defaultLog = getLogger('paths/dwc/validate');
```
NickPhura: nit: update the string here.

api/src/services/submission-service.ts
```diff
@@ -131,4 +142,84 @@ export class SubmissionService extends DBService {
   async listSubmissionRecords(): Promise<ISubmissionModelWithStatus[]> {
     return this.submissionRepository.listSubmissionRecords();
   }
+
+  /**
+   *
```
NickPhura: nit: add a description here

api/src/services/submission-service.ts
```diff
@@ -131,4 +142,84 @@ export class SubmissionService extends DBService {
...
+   * @param {number} submissionId
+   * @return {*}  {Promise<ISourceTransformModel['metadata_transform_precompile']>}
+   * @memberof SubmissionService
+   */
+  async getEMLStyleSheetKey(submissionId: number): Promise<ISourceTransformModel['transform_precompile_key']> {
+    return (await this.submissionRepository.getSourceTransformIdBySubmissionId(submissionId)).transform_precompile_key;
+  }
+
+  /**
+   *
```
NickPhura: nit: add a description here

api/src/services/submission-service.ts
```diff
@@ -131,4 +142,84 @@ export class SubmissionService extends DBService {
...
+   *
+   * @param {number} submissionId
+   * @return {*}  {Promise<any>}
+   * @memberof SubmissionService
+   */
+  async getStylesheetFromS3(submissionId: number): Promise<any> {
```
NickPhura: Lets specify the return type here, rather than saying `any`.

api/src/services/submission-service.ts
```diff
@@ -131,4 +142,84 @@ export class SubmissionService extends DBService {
...
+      }
+
+      const s3File = await getFileFromS3(stylesheet_key);
+
+      if (!s3File) {
+        throw new HTTP500('Failed to get file from S3');
+      }
+      return s3File;
+    } catch (error) {
+      defaultLog.error({ label: 'getS3File', message: 'error', error });
```
NickPhura: nit: can you change `getS3File` to the name of this function

NickPhura: Also, now that I look at it, may as well just remove the try catch altogether since the error just gets re thrown. Could maybe argue the log message is valuable, but we'll probably end up logging something higher up as a result of this error anyways

app/src/hooks/api/useAxios.test.ts
```diff
@@ -13,7 +13,7 @@ describe('APIError', () => {
       }
     } as Partial<AxiosError>;
 
-    const apiError = new APIError(error as unknown as AxiosError);
+    const apiError = new APIError((error as unknown) as AxiosError);
```
NickPhura: Looks like the same formatting thing I mentioned here: https://github.com/bcgov/biohubbc-platform/pull/46/files?file-filters%5B%5D=.tsx&show-deleted-files=true&show-viewed-files=true#r890750038

database/src/migrations/20220602152015_add_record_sims_eml_transformation.ts
```diff
@@ -0,0 +1,55 @@
...
+    SET SEARCH_PATH = ${DB_SCHEMA};
+
+    ALTER TABLE survey DROP COLUMN surveyed_all_areas;
+
+    -- Create 'survey' view
+
+    SET SEARCH_PATH = ${DB_SCHEMA_DAPI_V1};
+
+    CREATE OR REPLACE VIEW survey AS SELECT * FROM ${DB_SCHEMA}.survey;
+  `);
```
NickPhura: This down looks like its from some unrelated migration file

app/src/components/fields/AutocompleteField.tsx
anissa-agahchen: yeah, that did it, thanks

api/src/services/submission-service.ts
anissa-agahchen: yes, true, we do log it higher up.

database/src/migrations/20220602152015_add_record_sims_eml_transformation.ts
anissa-agahchen: yup, removed now

NickPhura (approved):

PR #62, opened 2022-06-22
Tech debt hooks with tests

curtisupshall:
# Overview

Added helper classes and tests for newly added frontend hooks

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1641


app/src/hooks/useDataLoader.test.tsx
```diff
@@ -0,0 +1,189 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import { Deferred } from 'test-helpers/promises';
+import useDataLoader from './useDataLoader';
```
KjartanE, line 3: This file just needs a format-fix to adjust formatting.

curtisupshall: Woops 😬

app/src/test-helpers/promises.ts
```diff
@@ -0,0 +1,69 @@
...
+   * @constructor
+   * @returns {Deffered} The deferred promise.
+   */
+  constructor() {
+    this.promise = new Promise<R>((resolve, reject) => {
+      this.resolve = resolve;
+      this.reject = reject;
+    });
+
+    return this;
```
NickPhura: nit: I don't think you need to return here, as you already receive an instance of the class when you call `new`.

app/src/test-helpers/promises.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Resets the Deferred instance with a new unfulfilled promise.
+   * @returns {Deferred} The deferred promise.
+   */
+  reset = () => {
```
NickPhura: I'd recommend making this function a traditional function (rather than an arrow function). 

As it is currently, reset is just another class variable like promise, resolve, and reject, and doesn't really exist on the class until you construct an instance. 

It might not matter much in this particular case, because this class is already a test helper and isn't going to need to be mocked/stubbed.  

But for example, in our API tests where we mock class functions via their prototype.  We stub out the method on the class prototype so that any new class instances created later by the code under test gets created with the stubs already in place. In this case though, the reset method isn't actually part of the class prototype and so wouldn't be mockable.  You would still be able to mock it after creating a new instance, but that assumes your test is able to pass that mocked instance into the real code, which isn't always possible.

NickPhura (dismissed):
Nice one!

NickPhura (approved):

PR #67, opened 2022-06-27
BHBC-1818: Run All Spatial Transforms and Persist Results

KjartanE:
# Overview

## Links to Jira tickets

- {List all applicable Jira tickets}

## PR Checklist

### Code

- [x] New files/classes/functions have appropriately descriptive names and comment blocks to describe their use/behaviour
- [x] I have avoided duplicating code when possible, moving re-usable pieces into functions
- [x] I have avoided hard-coding values where possible and moved any re-usable constants to a constants file
- [x] My code is as flat as possible (avoids deeply nested if/else blocks, promise chains, etc)
- [x] My code changes account for null/undefined values and handle errors appropriately
- [x] My code uses types/interfaces to help describe values/parameters/etc, help ensure type safety, and improve readability

### Style

- [x] My code follows the established style conventions

### Documentation

- [x] I have commented my code sufficiently, such that an unfamiliar developer could understand my code

### Tests

- [x] I have added/updated unit tests for any code I've added/updated

### Linting/Formatting

- [x] I have run the linter and fixed any issues, as needed  
- [x] I have run the formatter and fixed any issues, as needed  

### SonarCloud

### DIFF
--- a/api/src/paths/dwc/submission/{submissionId}/normalize.ts
+++ b/api/src/paths/dwc/submission/{submissionId}/normalize.ts
@@ -46,22 +46,7 @@ POST.apiDoc = {
   ],
   responses: {
     200: {
-      description: 'Successfully normalized and saved data of submission file',
-      content: {
-        'application/json': {
-          schema: {
-            type: 'object',
-            required: ['submission_status_id'],
-            properties: {
-              submission_status_id: {
-                type: 'integer',
-                minimum: 1
-              }
-            },
-            nullable: true
-          }
-        }
-      }
+      description: 'Successfully normalized and saved data of submission file'
     },
     ...defaultErrorResponses
   }
@@ -88,11 +73,11 @@ export function normalizeSubmission(): RequestHandler {
 
       const dwcArchiveFile = await darwinCoreService.getSubmissionRecordAndConvertToDWCArchive(submissionId);
 
-      const response = await darwinCoreService.normalizeSubmissionDWCA(submissionId, dwcArchiveFile);
+      await darwinCoreService.normalizeSubmissionDWCA(submissionId, dwcArchiveFile);
 
       await connection.commit();
 
-      res.status(200).json(response);
+      res.status(200).send();
     } catch (error) {
       defaultLog.error({ label: 'secureSubmission', message: 'error', error });
       await connection.rollback();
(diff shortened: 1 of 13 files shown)


api/src/paths/dwc/submission/{submissionId}/normalize.ts
```diff
@@ -57,7 +57,8 @@ POST.apiDoc = {
                 type: 'integer',
                 minimum: 1
               }
-            }
+            },
+            nullable: true
```
NickPhura: Why nullable? Since this is the 200 response, we should be able to control the response and always return consistently. We could decide that returning the submission_status_id isn't helpful, but in that case we should just remove the entire content section and just return `200`.

NickPhura: Technically there is nothing wrong with having different possible return values, just feels odd to me for w/e reason

api/src/repositories/spatial-repository.ts
```diff
@@ -0,0 +1,199 @@
...
+   * Insert given transformed data into Spatial Component Table
+   *
+   * @param {number} submissionId
+   * @param {Feature[]} transformedData
+   * @return {*}  {Promise<{ submission_spatial_component_id: number }>}
+   * @memberof SpatialRepository
+   */
+  async insertSubmissionSpatialComponent(
+    submissionId: number,
+    transformedData: Feature[]
```
NickPhura: I think we'll need to change this to `FeatureCollection` and account for it the geography code (possibly make a new function that works with FeatureCollections specifically).

Can chat about it as needed of course.

NickPhura: Actually, a FeatureCollection just has an array of Feature inside it, so probably dont need to change much at all to use the existing geometryCollection util function.

NickPhura (dismissed):

api/src/repositories/spatial-repository.test.ts
```diff
@@ -0,0 +1,278 @@
...
+      });
+
+      const generateGeometryCollectionSQLStub = sinon
+        .stub(spatialUtils, 'generateGeometryCollectionSQL')
+        .returns(SQL`valid`);
+
+      const spatialRepository = new SpatialRepository(mockDBConnection);
+
+      const response = await spatialRepository.insertSubmissionSpatialComponent(1, {
+        features: [
```
NickPhura, line 263: Can you add the `type: 'FeatureCollection'` property? Its the only thing missing that would make this a proper valid feature collection.

NickPhura: You may be able to remove the `as FeatureCollection` as a result.

NickPhura (approved):

PR #79, opened 2022-07-15
BHBC-1846: To-do Auditing

curtisupshall:
# Overview

General clean-up of "to-do" tasks given by code comments.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1846

 - Deleted submission occurence repository, service, endpoints and related code.
 - Added code to delete spatial data upon intake resubmission.

## PR Checklist

### Code

- [x] New files/classes/functions have appropriately descriptive names and comment blocks to describe their use/behaviour
- [x] I have avoided duplicating code when possible, moving re-usable pieces into functions
- [x] I have avoided hard-coding values where possible and moved any re-usable constants to a constants file
- [x] My code is as flat as possible (avoids deeply nested if/else blocks, promise chains, etc)
- [x] My code changes account for null/undefined values and handle errors appropriately
- [x] My code uses types/interfaces to help describe values/parameters/etc, help ensure type safety, and improve readability

### Style

- [x] My code follows the established style conventions

### Documentation

- [x] I have commented my code sufficiently, such that an unfamiliar developer could understand my code

### Tests

- [x] I have added/updated unit tests for any code I've added/updated

### Linting/Formatting

- [x] I have run the linter and fixed any issues, as needed  
- [x] I have run the formatter and fixed any issues, as needed  

### SonarCloud

- [x] I have addressed all SonarCloud Bugs, Vulnerabilities, Security Hotspots, and Code Smells


api/src/repositories/spatial-repository.ts
```diff
@@ -297,6 +297,18 @@ export class SpatialRepository extends BaseRepository {
...
+  async deleteSpatialComponentsBySubmissionId(submission_id: number): Promise<ISubmissionSpatialComponent[]> {
+    const queryBuilder = getKnexQueryBuilder()
+      .del()
+      .from('submission_spatial_component')
+      .where({ submission_id })
+      .returning('*');
+
+    const response = await this.connection.knex<ISubmissionSpatialComponent>(queryBuilder);
+
+    return response?.rows || [];
```
NickPhura: super nit: I believe rows is guaranteed to exist, and to be an array, so it should be safe to just return `response.rows`

api/src/services/dwc-service.test.ts
```diff
@@ -503,6 +507,8 @@ describe('DarwinCoreService', () => {
 
       await darwinCoreService.intake(multerFile, 'dataPackageId');
       expect(getSubmissionStub).to.be.calledWith('dataPackageId');
+      expect(deleteSpatialComponentsStub).to.be.calledWith(1);
+      // expect(deleteSpatialComponentsStub).to.
```
NickPhura: nit

NickPhura (commented):
Can you add some tests for the new delete function you added (in the spatial service and repository)? 

I think other than that it looks good

api/src/repositories/spatial-repository.ts
```diff
@@ -297,6 +297,21 @@ export class SpatialRepository extends BaseRepository {
     return response.rows;
   }
 
+  async deleteSpatialComponentsBySubmissionId(submission_id: number): Promise<{ submission_id: number }[]> {
+    const sqlStatement = SQL`
```
curtisupshall, line 301: @NickPhura Ended up modifying my approach from `getKnexQueryBuilder` to simply using raw SQL, since the query builder response was returning `undefined`.

NickPhura (approved):

PR #86, opened 2022-07-28
BHBC-1872: API Endpoint to Handle Spatial Query to Return Dataset Centroid

cgarrettjones:
# Overview

## Links to Jira tickets

- [BHBC-1872](https://quartech.atlassian.net/browse/BHBC-1872)

- addition of dataset centroid spatial transform
- adjustments to spatial transform implementation to implement application of all transforms found in spatial_transform table
- implemented deletion of spatiall_transform_submission records on re-ingest
- addition of datasetID parameter for search API
- removed hard coded spatial types

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/paths/dwc/spatial/search.ts
```diff
@@ -78,8 +90,6 @@ export function searchSpatialComponents(): RequestHandler {
 
       const response = await spatialService.findSpatialComponentsByCriteria(criteria);
 
-      await connection.commit();
-
```
NickPhura: I think we still want this. At the very least just for consistency with other cases.

The connection.open starts a "BEGIN" transaction block, and the connection.commit does the closing "COMMIT".

I'm not entirely sure what it means to open a block, do only reads, and then not close it.  Maybe its fine? My worry would be if it left a dangling transaction or something that ate up resources until it eventually timed out.

api/src/services/spatial-service.ts
```diff
@@ -99,6 +111,38 @@ export class SpatialService extends DBService {
...
+          dataPoint.result_data
+        );
+
+        await this.insertSpatialTransformSubmissionRecord(
+          transformRecord.spatial_transform_id,
+          submissionSpatialComponentId.submission_spatial_component_id
+        );
+      });
+    });
+  }
```
NickPhura: We'll need to tweak this slightly.  

A `forEach` loop doesn't honour async await between loops.  It will just run as fast as possible, kicking off N number of calls, which internally will honour the async await, but externally will run in parallel and may finish out of order, etc.

So the `transformed` variable may actually get written/read in an unexpected order, the way it is set up currently.
Easy fix is just to initialize `transformed` inside the loop, rather than outside, so its self contained.
`let transformed = await this.spatia...`

Also, if it matters that these transformations run in order, then we'll need to use a different kind of loop.

NickPhura: Fixed

NickPhura (approved):

PR #89, opened 2022-08-02
Fix: Display only data from intersection of DB and ElasticSearch

anissa-agahchen:
# Overview

## Links to Jira tickets

- {List all applicable Jira tickets}

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/api/src/repositories/submission-repository.ts
+++ b/api/src/repositories/submission-repository.ts
@@ -418,7 +418,7 @@ export class SubmissionRepository extends BaseRepository {
    * @return {*}  {Promise<number>}
    * @memberof SubmissionRepository
    */
-  async getObservationCountByDatasetId(datasetId: string): Promise<number> {
+  async getSpatialComponentCountByDatasetId(datasetId: string): Promise<number> {
     const sqlStatement = SQL`
       SELECT
         count(distinct submission_spatial_component_id)::integer
@@ -429,7 +429,7 @@ export class SubmissionRepository extends BaseRepository {
       ON
         s.submission_id = ssc.submission_id
       WHERE
-        s.uuid = ${datasetId}
+        s.uuid = ${datasetId};
     `;
 
     const response = await this.connection.sql<{ count: number }>(sqlStatement);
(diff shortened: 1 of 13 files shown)


api/src/repositories/submission-repository.ts
```diff
@@ -376,9 +380,37 @@ export class SubmissionRepository extends BaseRepository {
...
+    const sqlStatement = SQL`
+      SELECT
+        uuid
+      FROM
+        submission
+      WHERE
+        record_end_date IS NULL;
+    `;
+
+    const response = await this.connection.sql<{ uuid: string }>(sqlStatement);
```
NickPhura: You can also use your IDatasetId interface here: `.sql<IDatasetId>`

api/src/services/submission-service.ts
```diff
@@ -162,10 +163,20 @@ export class SubmissionService extends DBService {
...
     return this.submissionRepository.getSubmissionRecordJSONByDatasetId(datasetId);
   }
 
+  /**
+   *
+   *
+   * @return {*}  {Promise<string>}
+   * @memberof SubmissionService
+   */
+  async getDatasetIdsFromDB(): Promise<IDatasetId[]> {
```
NickPhura: If we are no longer using this, lets get rid of it (and whatever other related code)

api/src/repositories/submission-repository.ts
```diff
@@ -48,6 +48,10 @@ export interface ISubmissionModel {
 export interface ISubmissionModelWithStatus extends ISubmissionModel {
   submission_status: string;
 }
+
+export interface IDatasetId {
```
curtisupshall: Should we consider a different name for this interface?

curtisupshall (dismissed):
Looks good to me. My single comment is a nit-pick

api/src/repositories/submission-repository.ts
curtisupshall: Namely I wouldn't expect an object of type "dataset ID" to have a property "uuid" which actually contains said ID.

NickPhura: Good point, though this is going to be removed, so easy fix haha

edit: removed in recent commit

NickPhura (approved):

PR #106, opened 2022-08-15
Tech-debt: refactor dwc-service code

anissa-agahchen:
# Overview

The create function in DWC had 10 steps.  One step was redundant, and the remaining nine where broken out into their individual functions, and tests were rewritten.


AlfredRosenthal (approved):
looks good!

PR #109, opened 2022-08-16
BHBC-1844 Map Occurrence download 

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-1844](https://quartech.atlassian.net/browse/BHBC-1844?atlOrigin=eyJpIjoiMjVhNDZhYzM2NDc4NDIxM2IyZjkyZDViODQ0NWM0NzYiLCJwIjoiaiJ9)

- added new download button to dataset page
- new endpoint created to facilitate occurrence data download (currently raw search results as a json)
- new tests added for new endpoint

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


app/src/hooks/api/useSearchApi.ts
```diff
@@ -40,6 +40,20 @@ const useSearchApi = (axios: AxiosInstance) => {
 
   const getSpatialMetadata = async (submissionSpatialComponentId: number): Promise<ISpatialMetadata> => {
     const { data } = await axios.get(`/api/dwc/spatial/${submissionSpatialComponentId}/metadata`);
+
+    return data;
+  };
+
+  const getSpatialDataFile = async (criteria: {
```
NickPhura, line 47: Can you add a test for this function.  We just have a really basic test for each of the api functions. We haven't gone super exhaustive with those tests, mostly they just help ensure the URL doesnt get changed unintentionally, and verify the function params

app/src/features/datasets/DatasetPage.tsx
```diff
@@ -77,6 +87,14 @@ const DatasetPage: React.FC = () => {
     })
   );
 
+  useDataLoaderError(fileDataLoader, () => {
+    return {
+      dialogTitle: 'Error generating file',
```
NickPhura: super nit: capitalize the words in the title

app/src/features/datasets/DatasetPage.tsx
```diff
@@ -1,4 +1,4 @@
-import { Theme } from '@material-ui/core';
+import { Button, Grid, Theme } from '@material-ui/core';
```
NickPhura: Can you split these out into 3 separate imports (like `Box` below, etc).

Forever ago we moved to specific imports in an attempt to reduce build times. In theory grabbing specific mui components will be faster than grabbing the entire thing, and just using a few items.

NickPhura: There are some mixed posts online about it.  Some say it will reduce bundle size and/or improve startup times, others say webpack will handle it for you regardless (though they also say it depends on versions and how your app is configured, etc, so hard to say).

We did some testing locally at the time, and it did improve build times a bit (like 20% or something), though its still the slowest build overall vs the other builds.

This talks a bit about it: https://mui.com/material-ui/guides/minimizing-bundle-size/

(This is all just extra detail for those curious)

app/src/features/datasets/DatasetPage.tsx
AlfredRosenthal: done

app/src/hooks/api/useSearchApi.ts
AlfredRosenthal: I've added a basic test for the new endpoint.

app/src/features/datasets/DatasetPage.tsx
AlfredRosenthal: that makes sense, I've split the imports up.

api/src/paths/dwc/spatial/download.ts
```diff
@@ -0,0 +1,105 @@
...
+      boundary: JSON.parse(req.query.boundary as string) as Feature
+    };
+
+    const connection = req['keycloak_token'] ? getDBConnection(req['keycloak_token']) : getAPIUserDBConnection();
+
+    try {
+      await connection.open();
+      const spatialService = new SpatialService(connection);
+      const response = await spatialService.findSpatialComponentsByCriteria(criteria);
+      await connection.commit();
```
curtisupshall: Shouldn't `connection.commit()` occur at the end of the function, just before `res.status(200).send(zipToSend.toString('hex'))`? I'm not entirely certain, although that is the convention for most of our other endpoints.

AlfredRosenthal: moved `connection.commit()` to be consistent with the other endpoints

NickPhura: We can commit as soon as we are done making any database calls, though that being said, it realistically probably doesn't matter too much as nothing heavy is happening after we finish calling the service in this case.

NickPhura (approved):
👍

PR #111, opened 2022-08-17
TechDebt: Pipeline Cleanup/Improvements

NickPhura:
# Overview

## Links to Jira tickets

N/A

- Misc formatting

- Tightened the times on the startup/readiness/liveness probes
  - Ideally the startup time is improved as they check more often now

- Re-wrote some of the utils inherited from another team
  - Previously they weren't actually waiting correctly, and were harder to decipher, etc.

- Added an extra pod check in the db setup deployment.  It now checks that the database pod is ready before running the migrations, will hopefully eliminate the odd random failure we got when the setup pod would report it couldn't connect to the database.

- Added a makefile command for installing the pipeline packages (just to make it easier)

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


Makefile
```diff
@@ -317,6 +317,23 @@ format-fix: ## Runs `npm run format-fix` for all projects
...
 	@cd database && npm run format-fix && cd ..
 
+## ------------------------------------------------------------------------------
+## Run `npm` commands for all projects ./.pipeline
+## ------------------------------------------------------------------------------
+pipeline-install: ## Runs `npm install` for all projects
+	@echo "==============================================="
+	@echo "Running /api/.pipeline install"
+	@echo "==============================================="
+	@cd api/.pipeline && npm install && cd ../..
```
curtisupshall, line 327: Thoughts on using `pushd` and `popd` in lieu of of `cd`?

app/.pipeline/config.js
```diff
@@ -71,12 +73,9 @@ const phases = {
     instance: `${name}-dev-${deployChangeId}`,
     version: `${deployChangeId}-${changeId}`,
     tag: `dev-${version}-${deployChangeId}`,
-    host:
-      (isStaticDeployment && (staticUrls.dev)) ||
-      `${name}-${changeId}-a0ec71-dev.apps.silver.devops.gov.bc.ca`,
+    host: (isStaticDeployment && staticUrls.dev) || `${name}-${changeId}-a0ec71-dev.apps.silver.devops.gov.bc.ca`,
```
curtisupshall, line 76: Would a ternary be more readible here? I'm pretty sure it is functionally the same but I'm not certain

NickPhura: Should be the same in the end

That being said, there is definitely more to clean up in all of these pipeline files, especially around all of these variables, and how they get passed around (its a bit hard to trace them through the steps)

But maybe in another ticket

curtisupshall (approved):
Looks great! 🧹

Makefile
NickPhura: No strong thoughts, wasn't aware of pushd or popd tbh.

Did a quick search, seems like both do more or less the same thing.

I'd probably just leave them as is, unless we have a reason to change them

PR #114, opened 2022-08-19
Tech Debt: Metadata endpoint

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-1896](https://quartech.atlassian.net/browse/BHBC-1896?atlOrigin=eyJpIjoiNzc0YmNhNmViNzJkNDhiMTk3MzhhYTE0YzM0MWFkYmEiLCJwIjoiaiJ9)

- added security logic to meta data endpoint
- pulled some duplicated code into helper functions

Verifying will require secure data changes

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


curtisupshall (dismissed):
Looks good! 🔮

api/src/repositories/spatial-repository.ts
```diff
@@ -488,34 +488,11 @@ export class SpatialRepository extends BaseRepository {
...
-                    wuste.user_security_transform_exceptions @> wfscwst.spatial_component_security_transforms
-                  then
-                    wfscwst.spatial_component
-                  else
-                    coalesce(wfscwst.secured_spatial_component, wfscwst.spatial_component)
-                end
-            ) spatial_component
-          `
-        )
+        knex.fromRaw(this._buildSelectForSecureNonSecureSpatialComponents(knex))
```
NickPhura: I wonder if we need to pass `knex` through here, or if we can just get a new instance inside the function and it will still play nice?

We aren't leveraging any of the Knex database connection stuff, its only ever being used for its query building capabilities, so I would hope that using separate instances of knex would be fine. Maybe? Not sure how knex manages things internally though, maybe it would screw with the array of bindings, or something like that.

curtisupshall (approved):
👌

PR #115, opened 2022-08-22
BHBC-1905, 1906: Show Observation Metadata in a Single Tooltip

curtisupshall:
# Overview
Enhanced map search by grouping spatial data by coordinates, removing frontend clustering and "spiraling."

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1905
https://quartech.atlassian.net/browse/BHBC-1906

- Enhanced the spatial search endpoint to perform an additional distinct check on the coordinates part of the matching components, so that we don’t return spatial components that have the exact same coordinates. Check restricted to spatial components having `…properties.type = 'Occurrence'`.

 - Updated the spatial metadata endpoint to accept an array of spatial component ids, and return an array of metadata objects.

 - Updated the frontend to show paginated metadata.
 - Added a frontend util to download object arrays as a CSV file.
 - Removed marker clustering. Uninstalled `react-leaflet-cluster`.
## Known issues
 - Console warns: `listener not found`. See: https://bytemeta.vip/repo/geoman-io/leaflet-geoman/issues/1160 (Supposedly to be patched with release of Leaflet `1.8.1`).


api/src/paths/dwc/spatial/metadata.ts
```diff
@@ -50,9 +55,9 @@ GET.apiDoc = {
  *
  * @returns {RequestHandler}
  */
-export function getSpatialMetadataById(): RequestHandler {
+export function getSpatialMetadataBySubmissionSpatialComponentIds(): RequestHandler {
   return async (req, res) => {
-    const submissionSpatialComponentId = Number(req.params.submissionSpatialComponentId);
+    const submissionSpatialComponentIds = ((req.query.submissionSpatialComponentIds || []) as string[]).map(Number);
```
anissa-agahchen, line 60: style consideration:

I think  
```
const submissionSpatialComponentIds = ((req.query.submissionSpatialComponentIds || []) as string[]).map((item) =>
      Number(item)
    );
```
is more readable. @NickPhura, any preference?

anissa-agahchen (dismissed):

api/src/paths/dwc/spatial/metadata.ts
NickPhura: No preference really, either is fine. Your way Anissa is maybe a bit easier to read, for someone not familiar with the shorthand, but I think its fine its fine either way.

curtisupshall: Per discussion with Nick, `Number` constructor is acceptable to use in lieu of arrow function. See https://stackoverflow.com/questions/48343478/what-does-mapnumber-do-here

api/src/services/spatial-service.ts
```diff
@@ -221,24 +221,17 @@ export class SpatialService extends DBService {
...
-      submissionSpatialComponentId
-    );
+    const response = (await userService.isSystemUserAdmin())
+      ? this.spatialRepository.findSpatialMetadataBySubmissionSpatialComponentIdsAsAdmin(submissionSpatialComponentIds)
+      : this.spatialRepository.findSpatialMetadataBySubmissionSpatialComponentIds(submissionSpatialComponentIds);
 
-    return (response?.spatial_component?.spatial_data?.features?.[0]?.properties as Record<string, string>) || {};
+    return (await response)
+      .map((row) => row.spatial_component_properties)
+      .filter((row): row is Record<string, any> => Boolean(row));
```
NickPhura: I don't think we need this filter call?

Rather than cast this to a Record, we should just change the response of this function to an array of  `GeoJsonProperties`, to match what `spatial_component_properties` is already set to.

app/src/components/map/DatasetPopup.tsx
```diff
@@ -46,21 +46,21 @@ const useStyles = makeStyles(() => ({
...
   const dataLoader = useDataLoader(() => {
-    return api.search.getSpatialMetadata(submissionSpatialComponentId);
+    return api.search.getSpatialMetadata(submissionSpatialComponentIds);
   });
 
   dataLoader.load();
 
   const { isLoading, isReady } = dataLoader;
-  const data = dataLoader.data as BoundaryCentroidFeatureProperties;
+  const data = (dataLoader.data || []) as BoundaryCentroidFeatureProperties[];
```
NickPhura, line 63: Lets chat about a potential way to improve this, so we don't need the cast.

api/src/services/spatial-service.ts
curtisupshall: My reasoning was that `GeoJsonProperties` is basically just `Record<string, any> | null`, so I cast it as a `Record` so it would never be `null`. Although we have a check for `spatial_component != {}` in the SQL query, I don't think this necessitates that the `spatial_component_properties` value can't be `null` (but I might be wrong?) - hence I added the `filter` call. If we go for changing the service response to `GeoJsonProperties[]`, we should probably do the same for the API hook in the frontend. What do you think?

app/src/components/map/FeaturePopup.tsx
```diff
@@ -58,47 +102,66 @@ const useStyles = makeStyles(() => ({
...
-   */
-  const formatDate = (dateString: string) => {
-    const d = new Date(dateString);
-    return d instanceof Date && !isNaN(d as unknown as number) ? d.toDateString() : dateString;
-  };
-
   dataLoader.load();
 
-  const { isLoading, data, isReady } = dataLoader;
+  const { isLoading, isReady } = dataLoader;
```
NickPhura: (Not necessarily something that needs changing, but just my thinking): I'm sure I haven't been 100% consistent on this, but I tend to lean more towards not redefining data like this, as it makes it harder further down to understand where `isLoading`, etc, came from. If you see `dataLoader.isLoading` you know exactly what isLoading is based on. It can also make it easier to modify `dataLoader` later on, because its easier to track down all of the spots it impacts, since you can just search `dataLoader` and find all of its uses.

app/src/components/map/FeaturePopup.tsx
```diff
@@ -58,47 +102,66 @@ const useStyles = makeStyles(() => ({
...
-const FeaturePopup: React.FC<React.PropsWithChildren<{ submissionSpatialComponentId: number }>> = (props) => {
-  const { submissionSpatialComponentId } = props;
+const FeaturePopup: React.FC<React.PropsWithChildren<{ submissionSpatialComponentIds: number[] }>> = (props) => {
+  const { submissionSpatialComponentIds } = props;
 
   const classes = useStyles();
   const api = useApi();
+  const [currentIndex, setCurrentIndex] = useState(0);
 
   const dataLoader = useDataLoader(() => {
```
NickPhura: For future PRs: we should use a descriptive name for this variable, for example something like `spatialDataLoader`.  Just so it is a bit more self-documenting when its used further down the code.

app/src/components/map/components/MarkerCluster.tsx
```diff
@@ -37,54 +27,126 @@ export interface IMarkerLayersProps {
...
+      if (props.draggable === true) {
+        marker.dragging.enable();
+      } else {
+        marker.dragging.disable();
+      }
+    }
+  }
+);
+
+const MarkerClusterGroup = Fragment;
```
NickPhura: nit: remove

app/src/components/map/components/MarkerCluster.tsx
```diff
@@ -37,54 +27,126 @@ export interface IMarkerLayersProps {
...
-                    {item.popup}
-                  </Popup>
-                )}
-              </Marker>
-            );
-          })}
-        </ReactLeafletMarkerClusterGroup>
+      <LayersControl.Overlay checked={layer.visible} name={layer.layerName} key={`marker-layer-${layer.layerName}}`}>
+        <FeatureGroup>
+          <MarkerClusterGroup>
```
NickPhura: We shouldn't need this fragment wrapper

app/src/components/map/components/MarkerClusterGroup.tsx
```diff
@@ -0,0 +1,67 @@
+import { createPathComponent, LeafletContextInterface } from '@react-leaflet/core';
```
NickPhura: (file) It doesn't look like anything from here is used anywhere, so probably can remove this whole file?

app/src/utils/spatial-utils.tsx
```diff
@@ -178,23 +84,78 @@ export const parseSpatialDataByType = (
   return { markerLayers: [occurrencesMarkerLayer], staticLayers: [boundaryStaticLayer] };
 };
 
-// checks which key should be used to identify the layer
-export const getFeatureLayerKey = (spatialRecord: ISpatialData, feature: Feature): string => {
-  let key = '';
-
-  if (isOccurrenceFeature(feature)) {
-    key = `${spatialRecord.associated_taxa}`;
+const occurrenceMarkerSetup = (
```
NickPhura, line 114: Can you add some comments in/around this function.  

Its got some relatively complex chained calls, would be good to add some helper comments to guide a future dev through its purpose. 

Same goes for the two functions below as well.

In general for any functions like these, don't worry about over explaining things, even using examples when appropriate, comments are free and I'd rather we over do it than under do it.

NickPhura: Especially if there is any business logic involved, that a future dev might not be aware of, and which can't be fully interpreted from the code.

app/src/utils/spatial-utils.tsx
```diff
@@ -178,23 +84,78 @@ export const parseSpatialDataByType = (
...
 
-  if (isBoundaryCentroidFeature(feature)) {
-    key = `${spatialRecord.submission_spatial_component_id}`;
-  }
+export const parseProjectResults = (data: ISpatialData[], datasetVisibility: IDatasetVisibility): ISearchResult[] => {
+  const results: ISearchResult[] = [];
+  data.forEach((item) => {
+    if (item.spatial_data.features[0]) {
+      if (item.spatial_data.features[0].properties) {
+        if (item.spatial_data.features[0].properties.type === SPATIAL_COMPONENT_TYPE.BOUNDARY_CENTROID) {
```
NickPhura: Could probably squash these down into 1 if statement, with a few question marks in there to cover the null checks

app/src/utils/spatial-utils.tsx
```diff
@@ -178,23 +84,78 @@ export const parseSpatialDataByType = (
...
-  if (isBoundaryFeature(feature)) {
-    key = `${spatialRecord.submission_spatial_component_id}`;
-  }
+  return null;
+};
 
-  if (isBoundaryCentroidFeature(feature)) {
-    key = `${spatialRecord.submission_spatial_component_id}`;
-  }
+export const parseProjectResults = (data: ISpatialData[], datasetVisibility: IDatasetVisibility): ISearchResult[] => {
```
NickPhura: Maybe rename this to `parseDatasetResults` or `parseBoundaryCentroidResults`, whichever better follows the existing naming conventions

app/src/utils/spatial-utils.tsx
NickPhura: Consider using the existing `isBoundaryCentroidFeature` function at the bottom of this file, as it already does a similar check.

app/src/utils/spatial-utils.tsx
```diff
@@ -178,23 +84,78 @@ export const parseSpatialDataByType = (
...
+
+export const parseOccurrenceResults = (
+  data: ISpatialData[],
+  datasetVisibility: IDatasetVisibility
+): ISearchResult[] => {
+  const taxaMap = {};
+  data.forEach((spatialData) => {
+    spatialData.taxa_data.forEach((item) => {
+      // need to check if it is an occurnece or not
+      if (item.associated_taxa) {
```
NickPhura: We should use the `type` field from the spatial component properties to determine if it is an occurrence. And ideally leverage the `isOccurrenceFeature` function too.

NickPhura: Only because the associated_taxa is not the 'official' property that defines an occurrence, and this might impact us in future

api/src/services/spatial-service.ts
NickPhura: Yeah I was thinking about the `null` aspect too.  It is technically correct that a geojson feature can have null properties, even if we don't think we will ever set it to null in our database.

But on the other hand, if we think that null is unrealistic since we control the data in those tables, then the other option is to just define the properties as `Record<string, any>` and not bother using `GeoJsonProperties` at all.

I think it would be better to either use `GeoJsonProperties` and then be forced to handle a few extra null checks at the app level, or use `Record<string, any>` and not worry about null checks, and have it blow up if null gets returned unexpectedly.  But what we don't want to do is filter out nulls silently, which we don't really think should exist anyways, and then be left unaware of unexpected nulls in the database.

I guess I would lean towards using `GeoJsonProperties` if I had to pick, only because it is technically the most correct and forces us to be the most safe.

And yeah, whichever option we pick, we should propagate it everywhere so that the types at all levels are accurate.

curtisupshall: > But what we don't want to do is filter out nulls silently, which we don't really think should exist anyways, and then be left unaware of unexpected nulls in the database.

Well put.

> I think it would be better to either use GeoJsonProperties and then be forced to handle a few extra null checks at the app level...

I think I will opt for this.

app/src/components/map/DatasetPopup.tsx
curtisupshall: Added generic type to API hook, as per out conversation

app/src/components/map/FeaturePopup.tsx
curtisupshall: That's a very good point.

app/src/utils/spatial-utils.tsx
curtisupshall: Renamed to `parseBoundaryCentroidResults`.

NickPhura (dismissed):
Nice work, map is starting to look really good

edit: there is just 1 code smell, maybe you can clean that up in this PR, or just grab it in your next one.

NickPhura (approved):

PR #117, opened 2022-08-23
BHBC 1903

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC 1903](https://quartech.atlassian.net/browse/BHBC-1903?atlOrigin=eyJpIjoiMzBjNDQ0Y2FmMWQwNDhmYmE0YmJlZTFjMGRhOTI0YTQiLCJwIjoiaiJ9)

Updated data set search endpoint to return number of occurrences based on users security.
To Test: 
- navigate and try Find Dataset search as a public user (not logged in), track the number of occurrences
- login and try the Find Dataset search and track number of occurrences (should be more)

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/repositories/base-repository.ts
```diff
@@ -12,4 +14,12 @@ export class BaseRepository {
   constructor(connection: IDBConnection) {
     this.connection = connection;
   }
+
+  async isSystemUserAdmin(): Promise<boolean> {
+    const userService = new UserService(this.connection);
+    const userObject = await userService.getUserById(this.connection.systemUserId());
+    return [SYSTEM_ROLE.SYSTEM_ADMIN, SYSTEM_ROLE.DATA_ADMINISTRATOR].some((systemRole) =>
+      userObject.role_names.includes(systemRole)
+    );
```
NickPhura: Moving this to a reusable function is a good call.

It does spark some thoughts though about my original setup with the search repository functions. I don't think I should have done the admin check at the repository level, I should have done that at the service level. I'm thinking the repositories should probably never call the services, as they should be kept simple and focused only on building/running SQL queries.

If you're cool with refactoring this a bit, I think it would be best to move this `isSystemUserAdmin` function into the user service, and then move the admin checks we are currently doing in the spatial repository up to the spatial service. And then there will be no reference to any services in the spatial-repository, and the functions that do the admin checks currently can be removed.

api/src/repositories/submission-repository.ts
```diff
@@ -398,35 +398,108 @@ export class SubmissionRepository extends BaseRepository {
...
+      .with('with_user_security_transform_exceptions', (qb) => {
+        qb.select(knex.raw('array_agg(suse.security_transform_id) as user_security_transform_exceptions'))
+          .from('system_user_security_exception as suse')
+          .where('suse.system_user_id', this.connection.systemUserId());
+      })
+      // filter spatial components for data set
+      .with(
+        'with_filtered_spatial_component_with_security_transforms',
+        knex.raw(`
+      SELECT
```
NickPhura: nit: can you indent the body of this knex.raw (and the 2 other instances below) a bit.

NickPhura: having to manually format these template strings is a bit annoying, but helps so much in legibility

NickPhura:
Also, when this is ready to merge, just make sure to clean up the commit messages.  Reduce it down to a few key points, rather than the 15-20 raw commit messages.

api/src/repositories/submission-repository.ts
AlfredRosenthal: done

api/src/repositories/base-repository.ts
AlfredRosenthal: Good call moving it out of the repos. That logic lives in the services now. Tests are also updated to reflect those changes

api/src/repositories/spatial-repository.ts
```diff
@@ -506,9 +476,10 @@ export class SpatialRepository extends BaseRepository {
           'with_filtered_spatial_component_with_security_transforms as wfscwst, with_user_security_transform_exceptions as wuste'
         )
       );
-
+    console.log('');
+    console.log('___________________');
     const response = await this.connection.knex<ISubmissionSpatialSearchResponseRow>(queryBuilder);
-
+    console.log(response);
```
NickPhura: nit: couple of missed logs

api/src/repositories/base-repository.ts
NickPhura: Just looked through it all, looks really good!

NickPhura (approved):

PR #120, opened 2022-08-31
Bug fix: Fix search value for Moose

anissa-agahchen:
Reason for this:

When we search for a species, we pass in a species code from app, and the ocde for moose needed to be adjusted.


AlfredRosenthal (approved):
🎸

PR #133, opened 2022-10-13
BHBC-1953: Map Search Species Not Retained On Refine Search

curtisupshall:
# Overview
After successfully searching for records on the map, filtered by species, if you try to add more species to the existing species filter list, the original species get overwritten, rather than appended to.

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-1953

 - Fixes a bug in which species selection is not retained when completing and then refining a map search.
 - Add type definitions for taxonomy API hook


app/src/features/datasets/components/DatasetSearchForm.tsx
```diff
@@ -69,6 +69,8 @@ const DatasetSearchForm: React.FC<IDatasetSearchFormProps> = (props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [formikProps.values.area]);
 
+  const curatedSpeciesList = handleSortSelectedOption(formikProps?.values?.species_list || [], speciesList);
```
NickPhura: nit: I would maybe just move this call to handleSort directly into the options prop, since its never used anywhere else.

app/src/components/fields/MultiAutocompleteField.tsx
```diff
@@ -23,14 +22,26 @@ export interface IMultiAutocompleteField {
...
+  const selectedOptionsValue = selected.map((item) => item.value);
+  const remainingOptions = optionsLeft.filter((item) => !selectedOptionsValue.includes(item.value));
+
+  return [...selected, ...remainingOptions];
+};
+
 const MultiAutocompleteField: React.FC<IMultiAutocompleteField> = (props) => {
-  const { values, touched, errors, setFieldValue } = useFormikContext<IMultiAutocompleteFieldOption>();
+  const { getFieldMeta, setFieldValue } = useFormikContext();
+  const { value, touched, error } = getFieldMeta<IMultiAutocompleteFieldOption[]>(props.id);
```
NickPhura, line 47: Nice!

I originally had been using lodash `get` because it solved the issue of getting values (`values[name]`) when the name was something like `my.nested.name`.  But the `getFieldMeta` function also handles that apparently, which is a bit cleaner. Good catch.

app/src/components/fields/MultiAutocompleteField.tsx
```diff
@@ -23,14 +22,26 @@ export interface IMultiAutocompleteField {
   handleSearchResults?: (input: string) => Promise<void>;
 }
 
+export const handleSortSelectedOption = (
+  selected: IMultiAutocompleteFieldOption[],
+  optionsLeft: IMultiAutocompleteFieldOption[]
```
NickPhura: nit: I would maybe name these 2 variables `selectedOptions` and `remainingOptions`. Subjective of course, but I think its a bit easier to follow. Also lines up with the `selectedOptions` use state in the component below.

app/src/components/fields/MultiAutocompleteField.tsx
```diff
@@ -23,14 +22,26 @@ export interface IMultiAutocompleteField {
   handleSearchResults?: (input: string) => Promise<void>;
 }
 
+export const handleSortSelectedOption = (
```
NickPhura: Can you add a JSDOC description for this function? Just because its a bit of a unique business case to sort the list by appending the selected options to the top.

It might be good to include `autocomplete` in the name of this function somewhere, just because its being exported for use in other components. Maybe something like `sortAutocompleteOptions`? Or `prepareAutocompleteOptions`?

NickPhura (commented):
Just a couple tiny nits.  Looks great in general

NickPhura (approved):

NickPhura (dismissed):

NickPhura (dismissed):
🐐

NickPhura (approved):
🐑

PR #144, opened 2022-11-18
BHBC-2052

AlfredRosenthal:
# Overview

## Links to Jira tickets

[BHBC-2052](https://quartech.atlassian.net/browse/BHBC-2052?atlOrigin=eyJpIjoiZDIxNDcwNmE5MDA2NDA2Yjg4YjU5Zjc2YWM0NjQ4NTQiLCJwIjoiaiJ9)

- Recent changes for intake means the system no longer silently fails
- Issue found with intake endpoint failing with only project metadata (no occurrence information)
- Logic added to circumvent occurrence specific steps

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/api/src/services/dwc-service.ts
+++ b/api/src/services/dwc-service.ts
@@ -85,7 +85,8 @@ export class DarwinCoreService extends DBService {
 
       const isOnlyMetadata = await this.isSubmissionMetadataOnly(submissionId);
 
-      // skip these if the submission only contains metadata
+      // the following steps are for processing occurrence data
+      // skip these if the submission only contains metadata (no occurrence data)
       if (!isOnlyMetadata) {
         await this.create_step7_normalizeSubmissionDWCA(submissionId);
 
@@ -633,8 +634,8 @@ export class DarwinCoreService extends DBService {
 
   /**
    * Gets DwCArchive from submission ID and returns true/false if the submission is only metadata
-   * @param submissionId
-   * @returns
+   * @param {number} submissionId
+   * @returns {*} {Promise<boolean>}
    */
   async isSubmissionMetadataOnly(submissionId: number): Promise<boolean> {
     const dwcArchive = await this.getSubmissionRecordAndConvertToDWCArchive(submissionId);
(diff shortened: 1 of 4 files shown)


api/src/services/dwc-service.ts
```diff
@@ -83,11 +83,16 @@ export class DarwinCoreService extends DBService {
 
       await this.create_step6_transformAndUploadMetaData(submissionId, dataPackageId);
 
-      await this.create_step7_normalizeSubmissionDWCA(submissionId);
+      const isOnlyMetadata = await this.isSubmissionMetadataOnly(submissionId);
 
-      await this.create_step8_runSpatialTransforms(submissionId);
+      // skip these if the submission only contains metadata
```
NickPhura: nit: can you elaborate on this comment. Include something about how these steps all involve processing the darwin core observation data, and so can be skipped if the incoming dataset doesn't include it.

api/src/utils/media/dwc/dwc-archive-file.ts
```diff
@@ -85,6 +85,18 @@ export class DWCArchive {
     }
   }
 
+  // checks if this DwCArchive only has meta data
```
NickPhura: nit: mind turning this into a full jsdoc comment

AlfredRosenthal: done

api/src/services/dwc-service.ts
AlfredRosenthal: added a bit more to clarify

api/src/utils/media/dwc/dwc-archive-file.ts
```diff
@@ -85,6 +85,22 @@ export class DWCArchive {
...
 
+  /**
+   * This function checks worksheet data if the DwCArchive only contains metadata
+   *
+   * @returns {boolean} True if no worksheet data is present
+   */
+  isMetaDataOnly(): boolean {
+    return (
+      this.eml != null &&
+      this.worksheets.event == null &&
```
curtisupshall: Should we be using triple-equals instead, to check if these are strictly equal to null? If we're just checking that these properties are truthy and not if they're strictly null, maybe you could just use `!this.worksheets.event && ...`.

AlfredRosenthal: good point. I've refined the checks

NickPhura (approved):

PR #155, opened 2023-02-03
BHBC-2154: Implement Artifact Submission API

curtisupshall:
# Overview

This is the endpoint that will receive artifacts (documents, files) from collector systems. It will create database records for each artifact, and upload them to S3. This will allow artifacts to be reviewed and displayed in BioHub.

## Links to Jira tickets

 - https://quartech.atlassian.net/browse/BHBC-2154

 - Adds a new endpoint for uploading submitted artifacts
 - Adds a `key` column to the `artifact` table
 - Adds a new Artifact service and repository class
 - Adds a `getOrCreateSubmission` service method


api/src/repositories/submission-repository.ts
```diff
@@ -213,6 +197,39 @@ export class SubmissionRepository extends BaseRepository {
...
+  async getOrInsertSubmissionRecord(submissionData: IInsertSubmissionRecord): Promise<{ submission_id: number }> {
+    const sqlStatement = SQL`
+      INSERT INTO submission (
+        source_transform_id,
+        uuid
+      ) VALUES (
+        ${submissionData.source_transform_id},
+        ${submissionData.uuid}
+      )
+      ON CONFLICT (uuid) DO UPDATE SET uuid = ${submissionData.uuid}
```
curtisupshall, line 220: I'll add a comment explaining this, but basically if you put `ON CONFLICT ... DO NOTHING`, then `RETURNING` will fail to return anything

api/src/paths/artifact/intake.ts
```diff
@@ -0,0 +1,188 @@
...
+      
+      // Retrieve the ID that will be assigned to the artifact once it's inserted
+      const insertArtifactId = await artifactService.getNextArtifactId();
+
+      const artifactPersistPromises = files.map(async (file: Express.Multer.File, fileIndex: number) => {
+        const artifactMetadata: IArtifactMetadata = JSON.parse(metadataRecords[fileIndex]);
+        const artifactId = insertArtifactId + fileIndex;
+
+        // Generate the S3 key for the artifact, using the preemptive artifact ID + the package UUID
+        const s3Key = generateS3FileKey({
```
curtisupshall: Still undecided on what the final S3 path will be, but right now it's `platform/<UUID>/artifacts/<artifactID>/<file>`

api/src/repositories/submission-repository.ts
```diff
@@ -223,7 +240,7 @@ export class SubmissionRepository extends BaseRepository {
    */
   async updateSubmissionRecordInputKey(
     submissionId: number,
-    inputKey: IInsertSubmissionRecord['input_key']
+    inputKey: string
```
curtisupshall: `IInsertSubmissionRecord` will lose most of its params due to DB restructure; We will need a new `IInsertSubmissionMetadataRecord` interface and corresponding repo method

api/src/paths/artifact/intake.ts
```diff
@@ -0,0 +1,188 @@
...
+        source_transform_id: sourceTransformRecord.source_transform_id,
+        uuid: dataPackageId
+      });
+      
+      // Retrieve the ID that will be assigned to the artifact once it's inserted
+      const insertArtifactId = await artifactService.getNextArtifactId();
+
+      const artifactPersistPromises = files.map(async (file: Express.Multer.File, fileIndex: number) => {
+        const artifactMetadata: IArtifactMetadata = JSON.parse(metadataRecords[fileIndex]);
+        const artifactId = insertArtifactId + fileIndex;
```
curtisupshall: Theoretically this should be safe against collisions. Can't do an `await artifactService.getNextArtifactId()` because until the transaction finishes, this resolves the same artifact ID.

api/src/paths/artifact/intake.ts
```diff
@@ -0,0 +1,188 @@
...
+      const { submission_id } = await submissionService.getOrInsertSubmissionRecord({
+        source_transform_id: sourceTransformRecord.source_transform_id,
+        uuid: dataPackageId
+      });
+      
+      // Retrieve the ID that will be assigned to the artifact once it's inserted
+      const insertArtifactId = await artifactService.getNextArtifactId();
+
+      const artifactPersistPromises = files.map(async (file: Express.Multer.File, fileIndex: number) => {
+        const artifactMetadata: IArtifactMetadata = JSON.parse(metadataRecords[fileIndex]);
```
curtisupshall: Alternative approach would be to inline all the metadata at the request level...

```
const meta_title = req.body?.title[fileIndex];
const meta_desc = req.body?.description[fileIndex];
```

curtisupshall: Would allow us to write OpenAPI schema for the artifact metadata

api/src/repositories/artifact-repository.ts
```diff
@@ -0,0 +1,87 @@
...
+  async getNextArtifactId(): Promise<number> {
+    defaultLog.debug({ label: 'getNextArtifactId' });
+
+    const sqlStatement = SQL`
+      SELECT
+        COALESCE(MAX(artifact_id), 0) + 1
+      AS
+        artifact_id
+      FROM
+        artifact;
```
NickPhura: I believe there is a sequence you can call instead, and it will get the next available value for you:

```
select nextval('artifact_seq');
```

api/src/paths/artifact/intake.ts
curtisupshall: Actually, this won't work if we use `artifact_seq`, because it wouldn't increment the sequence

api/src/paths/artifact/intake.ts
NickPhura: I think we will need to change it so that the `artifactID` is the final portion of the path. 

Looking at the artifact table, I dont see a key column, so I think we will need the S3 path to be deterministic:
`platform/<UUID>/artifacts/<artifactID>`

NickPhura: Not sure of the top of my head what advantage this provides over just having a key column, but that seems to be how it is set up currently

api/src/paths/artifact/intake.ts
NickPhura: If we swap the SQL to use the sequence, then we won't be able to do this, as the sequence won't have incremented enough times.

I'm not sure if there are any downsides, but this seems to be a potential solution:
`select nextval('your_sequence') from generate_series(1,<number of values to return>);`

NickPhura: That came up online as a way of fetching multiple values from a sequence. I can't speak to any pros/cons about it though.

NickPhura:
@curtisupshall One thing we should do, is move all of the code between the connection.open and commit, into a single service function, similar to the dwc intake endpoint. Just so its not residing in the endpoint itself.

api/src/paths/artifact/intake.ts
NickPhura: We should be able to define an array of objects, with all of its properties, in the openapi schema.


```
metadata: {
  type: 'array',
  items: {
    type: 'object',
    properties: {
      file_name: {
        type: 'string'
      },
      other: {...}
    }
    description: 'JSON containing the metadata for each file'
  }
},
```

api/src/paths/artifact/intake.ts
curtisupshall: > I'm not sure if there are any downsides, but this seems to be a potential solution:
`select nextval('your_sequence') from generate_series(1,<number of values to return>);`

That seems promising - I'll give it a try

AlfredRosenthal (commented):
🐶

AlfredRosenthal (approved):
🐩

PR #159, opened 2023-02-13
BHBC-2168: Queue Scheduler Process

NickPhura:
# Overview

## Links to Jira tickets

https://quartech.atlassian.net/browse/BHBC-2168

New queue process that is designed to be deployed on its own (own dockerfile, etc).
Will poll the database for queue records, kicking off jobs, etc.

## Testing

To insert a queue record for testing purposes, you can run the following as many times as you want.

```
  with sub as (
      insert into submission (
          source_transform_id, 
          uuid
      )
      values (
          1, 
          public.gen_random_uuid()
      )
      returning submission_id
  )
  insert into submission_job_queue (
      submission_job_queue_id,
      submission_id,
      key
  )
  values (
      (select nextval('submission_job_queue_seq')),
      (select submission_id from sub),
      'a/fake/key'
  );
```

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/queue/queue-scheduler.ts
```diff
@@ -0,0 +1,201 @@
...
+import { SubmissionJobQueueService } from '../services/submission-job-queue-service';
+import { SystemConstantService } from '../services/system-constant';
+import { getLogger } from '../utils/logger';
+import { Queue } from './queue';
+
+const defaultLog = getLogger('queue/queue-scheduler');
+
+export const QUEUE_DEFAULT_ENABLED = false;
+export const QUEUE_DEFAULT_CONCURRENCY = 4;
+export const QUEUE_DEFAULT_PERIOD = 5000; // 5 seconds
```
curtisupshall, line 12: Minor nitpick: Thoughts on adding units to these two consts? E.g. `QUEUE_DEFAULT_PERIOD_MS`

api/src/services/system-constant.ts
```diff
@@ -0,0 +1,21 @@
+import { IDBConnection } from '../database/db';
+import { ISystemConstant, SystemConstantRepository } from '../repositories/system-constant';
+import { DBService } from './db-service';
+
+export class SystemConstantService extends DBService {
```
curtisupshall: +1 for system constant service 👍

api/src/repositories/system-constant.ts
```diff
@@ -0,0 +1,56 @@
...
+   * @return {*}  {Promise<ISystemConstant>}
+   * @memberof SystemConstantRepository
+   */
+  async getSystemConstant(constantName: string): Promise<ISystemConstant> {
+    const knex = getKnex();
+    const queryBuilder = knex.queryBuilder().select().from('system_constant').where('constant_name', constantName);
+
+    const response = await this.connection.knex<ISystemConstant>(queryBuilder);
+
+    return response.rows[0];
```
curtisupshall: Needs case when `rows === undefined`?

curtisupshall: Same for `getSystemConstants`

database/src/migrations/release.0.7.0/populate_system_constant.sql
```diff
@@ -6,4 +6,10 @@ insert into system_constant (constant_name, character_value, description) values
...
+-- job queue
+insert into system_constant (constant_name, character_value, description) values ('JOB_QUEUE_ENABLED', 'true', 'Controls whether or not the job queue scheduler is enabled. Set to "true" to enable the job queue schdulerF, "false" to disable it.');
+insert into system_constant (constant_name, numeric_value,   description) values ('JOB_QUEUE_CONCURRENCY', 4, 'The number of job queue processes that can run concurrently (integer > 0).');
+insert into system_constant (constant_name, numeric_value,   description) values ('JOB_QUEUE_PERIOD', 5000, 'The frequency with which the job queue scheduler will check for new job queue records (milliseconds).');
+insert into system_constant (constant_name, numeric_value,   description) values ('JOB_QUEUE_ATTEMPTS', 2, 'The total number of times a job will be attempted until it finishes successfully (integer >= 1).');
+insert into system_constant (constant_name, numeric_value,   description) values ('JOB_QUEUE_TIMEOUT', 60000, 'The maximum duration a running job can take before it is considered timed out.');
```
curtisupshall: Missing units in `description` column

api/src/repositories/submission-job-queue-repository.ts
```diff
@@ -112,4 +114,132 @@ export class SubmissionJobQueueRepository extends BaseRepository {
...
+
+    queryBuilder.orderBy('submission_job_queue_id', 'ASC');
+
+    if (concurrency) {
+      queryBuilder.limit(concurrency);
+    }
+
+    const response = await this.connection.knex<ISubmissionJobQueueRecord>(queryBuilder);
+
+    return response.rows;
```
curtisupshall, line 190: `rows === undefined`

api/src/repositories/system-constant.ts
NickPhura: `rows` will always be an array, so accessing the 0th index is safe, it will just return `undefined` if there are no elements.

That being said, this function is actually never used, so I am just going to remove it.

api/src/queue/queue-scheduler.ts
NickPhura: No strong feelings either way. I guess the only consideration is that we should probably do the same thing with all constants then, which might get a bit messy.

database/src/migrations/release.0.7.0/populate_system_constant.sql
NickPhura: Added, will wait to push as testing is happening in OpenShift atm

api/src/repositories/submission-job-queue-repository.ts
NickPhura: Same as above: rows will always be an array, so we shouldn't need to check it.

anissa-agahchen (dismissed):

anissa-agahchen (dismissed):

AlfredRosenthal:
Pulled our changes and did some more testing, should subsequent submissions work? It looks like the transforms are failing because of a duplicate key issue on `security_transform_submission`.

NickPhura:
> Pulled our changes and did some more testing, should subsequent submissions work? It looks like the transforms are failing because of a duplicate key issue on `security_transform_submission`.

Hmm, yeah that seems like a bug. We do want to support subsequent submissions (end-date old records, insert new records).

AlfredRosenthal (approved):
🥼

PR #164, opened 2023-03-01
BHBC-2230: API - Refactor API For Frontend Calls - Spatial 2

KjartanE:
# Overview

## Links to Jira tickets

- https://quartech.atlassian.net/browse/BHBC-2230

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/repositories/spatial-repository.ts
```diff
@@ -711,13 +716,13 @@ export class SpatialRepository extends BaseRepository {
...
             ),
             knex.raw(
               "jsonb_array_elements(ssc.spatial_component -> 'features') #> '{properties, dwc, vernacularName}' as vernacular_name"
             ),
             'ssc.submission_spatial_component_id',
-            'ssc.submission_id',
+            'ssc.submission_observation_id',
             'ssc.spatial_component',
             'ssc.secured_spatial_component'
           )
```
NickPhura, line 731: (It wont let me comment on the right line)

After Line 729, we need to do an additional join on the `submission_observation` table and add a where clause for `record_effective_timestamp is null`.

The submission_observation table will have multiple records for a single submission (if there were updates) and so we need to filter out end dated records, so we don't return old spatial components in our result.

The same join and where clause would need to be added to the `findSpatialMetadataBySubmissionSpatialComponentIdsAsAdmin` function as well.

```
  .join('submission_observation as so', 'so.submission_observation_id', 'ssc.submission_observation_id')
  .where('so.record_end_timestamp', null)
```

NickPhura: There is a small bug (fixed in Anissas branch) that will cause you issues if you try to re-send a dataset (as an update). An easy fix, but might be easier to discuss in discord tomorrow.

NickPhura (approved):

PR #184, opened 2023-04-27
SIMSBIOHUB-11: Misc UI fixes.

NickPhura:
# Overview

## Links to Jira tickets

- n/a

Some misc UI fixes that I noticed while testing the changes for (ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-11) (PR: https://github.com/bcgov/biohubbc/pull/1021)

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud

### DIFF
--- a/app/src/components/map/FeaturePopup.tsx
+++ b/app/src/components/map/FeaturePopup.tsx
@@ -35,6 +35,8 @@ export type BoundaryCentroidFeature = Feature & { properties: BoundaryCentroidFe
 
 export type BoundaryCentroidFeatureProperties = {
   type: SPATIAL_COMPONENT_TYPE.BOUNDARY_CENTROID;
+  datasetID: string;
+  datasetTitle: string;
 };
 
 export enum COMMON_METADATA_PROPERTIES {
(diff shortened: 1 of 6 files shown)


app/src/components/map/FeaturePopup.tsx
```diff
@@ -216,7 +218,10 @@ const FeaturePopup: React.FC<React.PropsWithChildren<{ submissionSpatialComponen
               size="small"
               variant="text"
               startIcon={<Icon path={mdiChevronLeft} size={1} />}
-              onClick={() => handlePrev()}
+              onClick={(event) => {
+                event.stopPropagation();
```
NickPhura, line 222: Clicking the Prev/Next buttons was causing the popup to close (clicking on the map closes the popup)

app/src/utils/spatial-utils.tsx
```diff
@@ -75,7 +75,7 @@ export const parseSpatialDataByType = (
         if (visible) {
           boundaryStaticLayer.features.push({
             geoJSON: feature,
-            key: feature.id || feature.properties.id,
+            key: feature.id || feature.properties.datasetID,
```
NickPhura, line 78: The previous value (`...properties.id`) didn't exist, and so the feature key was defaulting to a generic index (when it gets rendered later on by a different component), which isn't unique, and was causing funny behaviour when toggling the marker visibility. As markers were added/removed from the array of visible markers, their indexes were changing to whatever their position in the array was, and so you'd have the wrong markers toggling, etc.

curtisupshall (approved):
Nice :) Code looks good

anissa-agahchen (approved):

PR #187, opened 2023-06-01
SIMSBIOHUB-2: Secure or Unsecure a Single Document or File

KjartanE:
# Overview

## Links to Jira tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-2

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/paths/security/persecution-harm/apply.ts
```diff
@@ -0,0 +1,138 @@
...
+        'application/json': {
+          schema: {
+            type: 'array',
+            items: {
+              title: 'Artifact Persecution',
+              type: 'object',
+              required: ['artifact_persecution_id'],
+              properties: {
+                artifact_persecution_id: {
+                  type: 'number'
```
curtisupshall: Nit: `type: 'integer'`

api/src/repositories/security-repository.ts
```diff
@@ -0,0 +1,124 @@
+import SQL from 'sql-template-strings';
+import { z } from 'zod';
+import { ApiExecuteSQLError } from '../errors/api-error';
+import { getLogger } from '../utils/logger';
+import { BaseRepository } from './base-repository';
+
+const defaultLog = getLogger('repositories/security-repository');
+
+export const PersecutionAndHarmSecurity = z.object({
```
curtisupshall, line 9: Are we still following the pattern of prefixing type interfaces with `I`, e.g. `IPersecutionAndHarmSecurity`?

api/src/repositories/security-repository.ts
```diff
@@ -0,0 +1,124 @@
...
+   */
+  async removeAllSecurityRulesFromArtifact(artifactId: number): Promise<void> {
+    defaultLog.debug({ label: 'removeAllSecurityRulesFromArtifact' });
+
+    const sqlStatement = SQL`
+      DELETE FROM
+        artifact_persecution
+      WHERE
+        artifact_id = ${artifactId};
+    `;
```
curtisupshall, line 116: Nit: Could add `RETURNING artifact_id` and return this from the function

api/src/paths/security/persecution-harm/list.ts
```diff
@@ -0,0 +1,88 @@
...
+              required: [
+                'persecution_or_harm_id',
+                'persecution_or_harm_type_id',
+                'wldtaxonomic_units_id',
+                'name',
+                'description'
+              ],
+              properties: {
+                persecution_or_harm_id: {
+                  type: 'number'
```
curtisupshall: Nit: change these to integers

api/src/services/security-service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Apply security rules to an artifact.
+   *
+   * @param {SecurityReason[]} securityReasons
+   * @param {Artifact[]} selectedArtifacts
+   * @return {*}  {(Promise<({ artifact_persecution_id: number } | undefined)[]>)}
+   * @memberof SecurityService
+   */
+  async applySecurityRulesToArtifacts(
```
curtisupshall, line 43: Could simplify this function a bit if we just took `artifactIds` and `securityReasonIds` as the two parameters.

api/src/services/security-service.ts
```diff
@@ -0,0 +1,76 @@
...
+    securityReasons: SecurityReason[],
+    selectedArtifacts: Artifact[]
+  ): Promise<{ artifact_persecution_id: number }[][]> {
```
curtisupshall: If we intend on consuming these IDs as a flat array, then instead of returning a 2D array, we can flatten the promises using `Array.reduce`:

```typescript
const promises: Promise<{ artifact_persecution_id: number }>[] = artifactIds
  .reduce((acc: Promise<{ artifact_persecution_id: number }>[], artifactId: number) => {
    securityReasonIds.forEach((securityReasonId: number) => {
      acc.push(this.securityRepository.applySecurityRulesToArtifact(artifactId, securityReasonId));
    });

    return acc;
  }, []);
  
return Promise.all(promises);
```

api/src/services/security-service.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Apply security rules to an artifact.
+   *
+   * @param {number} artifactId
+   * @return {*}  {Promise<void>}
+   * @memberof SecurityService
+   */
+  async removeAllSecurityRulesFromArtifact(selectedArtifacts: Artifact[]): Promise<void> {
```
curtisupshall: Could also simplify this function by just passing in the IDs:

```typescript
async removeAllSecurityRulesFromArtifact(artifactIds: number[]): Promise<void> {
  defaultLog.debug({ label: 'removeAllSecurityRulesFromArtifact' });

  await Promise.all(artifactIds.map(this.securityRepository.removeAllSecurityRulesFromArtifact));
}
```

api/src/paths/security/persecution-harm/apply.ts
```diff
@@ -0,0 +1,138 @@
...
+
+      const response: { artifact_persecution_id: number }[] = [];
+
+      if (securityReasons.length > 0) {
+        const artifactPersecutionIds = await securityService.applySecurityRulesToArtifacts(
+          securityReasons,
+          selectedArtifacts
+        );
+
+        forEach(artifactPersecutionIds, (artifactPersecution) => {
```
curtisupshall: Re: my other comment on `applySecurityRulesToArtifacts`, you should be able to avoid the work of flattening the results in these two `forEach` loops

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+        }}>
+        <Formik
+          innerRef={formikRef}
+          initialValues={SecurityReasonsInitialValues}
+          validationSchema={SecurityReasonsYupSchema}
+          validateOnBlur={true}
+          validateOnChange={false}
+          onSubmit={(values: { securityReasons: ISecurityReason[] }) => {
+            handleSubmit(values.securityReasons);
+            setApplySecurityText('You successfully applied security reasons to the file (s).');
```
curtisupshall: Nit: Could programatically pluralize this message

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+            onClose();
+          }}>
+          {(formikProps) => (
+            <>
+              <YesNoDialog
+                open={yesNoDialogOpen}
+                onClose={() => setYesNoDialogOpen(false)}
+                onYes={async () => {
+                  await formikProps.submitForm();
+                  setApplySecurityText('You successfully unsecued the file (s).');
```
curtisupshall: Typo: unsecured

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+                  <Box sx={{ mb: 2, display: 'flex', flexDirection: 'column', alignContent: 'center' }}>
+                    <Box sx={{ display: 'flex', flexDirection: 'row', justifyContent: 'center' }}>
+                      <Icon path={mdiInformationOutline} size={5} color={'#a12622'} />
+                    </Box>
+                    <DialogTitle id="component-dialog-title" align={'center'}>
+                      Warning
+                    </DialogTitle>
+                    <DialogContentText id="alert-dialog-description" align={'center'} sx={{ py: 2 }}>
+                      You are going to make this document available to the public. This document can be downloaded.
+                      Also, if there is any security reasons, they will be removed.
```
curtisupshall: Nit, grammar: "Also, if there **are** any security reasons,"

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+                    </DialogContentText>
+                  </Box>
+                }
+              />
+
+              <DialogTitle id="component-dialog-title">Apply Security Reasons</DialogTitle>
+              <DialogContent sx={{ display: 'flex', flexDirection: 'column', overflow: 'hidden' }}>
+                <Box sx={{ mb: 2 }}>
+                  <DialogContentText id="alert-dialog-description">
+                    Search for the security reasons and apply them to the selected doument(s).
```
curtisupshall: Typo: documents

curtisupshall: Also, could maybe dynamically pluralize this?

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const [yesNoDialogOpen, setYesNoDialogOpen] = useState(false);
+
+  const [applySecurityCompleted, setApplySecurityCompleted] = useState(false);
+  const [applySecurityText, setApplySecurityText] = useState('');
+
+  const [formikRef] = useState(useRef<FormikProps<any>>(null));
+
+  const handleSubmit = async (securityReasons: ISecurityReason[]) => {
+    const response = await biohubApi.security.applySecurityReasonsToArtifacts(selectedArtifacts, securityReasons);
+    console.log('response', response);
```
curtisupshall: 👩‍💻

app/src/hooks/api/useSecurityApi.ts
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @return {*}  {Promise<IListPersecutionHarmResponse>}
+   */
+  const listPersecutionHarmRules = async (): Promise<IListPersecutionHarmResponse> => {
+    const { data } = await axios.get('/api/security/persecution-harm/list');
+
+    return data;
+  };
+
+  const applySecurityReasonsToArtifacts = async (
```
curtisupshall, line 22: Since the server only cares about the artifact ID and security reason ID, we could save some work and just pass these along instead.

app/src/features/datasets/components/security/SecurityReasonSelector.tsx
```diff
@@ -0,0 +1,113 @@
...
+        <Box sx={{ p: 2, height: '100%', overflow: 'hidden', overflowY: 'scroll' }}>
+          <FieldArray
+            name="securityReasons"
+            render={(arrayHelpers) => (
+              <>
+                {values.securityReasons.map((securityReason, index) => {
+                  return (
+                    <Box key={securityReason.name} py={1} px={2}>
+                      <SecurityReason
+                        key={index}
```
curtisupshall: Should use `securityReason.id` as the key

app/src/features/datasets/components/DatasetArtifacts.tsx
```diff
@@ -269,6 +190,20 @@ const DatasetAttachments: React.FC<IDatasetAttachmentsProps> = (props) => {
             slots={{
               noRowsOverlay: NoArtifactRowsOverlay
             }}
+            onRowSelectionModelChange={(params) => {
```
curtisupshall, line 193: If we end up just maintain the selected artifact IDs, we should be able to avoid this callback altogether:

```typescript
  onRowSelectionModelChange={setSelectedArtifactIds}
/>
```

api/src/repositories/security-repository.ts
KjartanE: I followed example in code, 'artifact-repository.ts' 
image
Updating these variable names causes slight issues, as the prefix `I` causes a issues by type.
image
Where Zod is maintaining our json object restriction, naming I on the prefix no longer allows zod to be an object.

api/src/repositories/security-repository.ts
KjartanE: Can do this, but my consideration was that on the endpoint or frontend, we are doing nothing with that return. I could return and run a check if something were to return. But would just create more testing cases, if there were to be an erorr that happens in the DB then it will still be caught by endpoint try/catch

curtisupshall: Yeah totally fair. Feel free to ignore

api/src/paths/security/persecution-harm/apply.ts
NickPhura: Could add minimum: 1 as well

api/src/repositories/security-repository.ts
NickPhura: The `I` prefix shouldn't make or break anything, its just a stylistic choice.

We originally used it for interfaces, just because it was a pattern others were using. But for types, it starts to break down a bit, since its not technically an interface (even though the difference is usually negligible).

But ya, I don't have a strong preference personally and find myself using both at times. As long as the type is named appropriately that is what matters more. If anyone has an opinion one way or the other though, then we can do that.

api/src/services/security-service.ts
NickPhura: There is also the array `flat` method

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

If you know the depth you can specify it, but you can also just say `<array>.flat(Infinity)`

You might have to combine it with a map operation to pull out just the ids from the flattened array of response objects.

database/src/migrations/20230531012345_remove_unique_id_artifact_persecution.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    set search_path=biohub_dapi_v1;
+
+    drop view if exists artifact_persecution;
+
+    set search_path=biohub;
+    DROP INDEX if exists artifact_persecution_uk1;
+    CREATE INDEX artifact_persecution_uk1 ON artifact_persecution(persecution_or_harm_id);
```
cgarrettjones:
```suggestion
    CREATE unique INDEX artifact_persecution_uk1 ON artifact_persecution(persecution_or_harm_id, artifact_id);
```

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+              open: true
+            });
+            onClose();
+          }}>
+          {(formikProps) => (
+            <>
+              <YesNoDialog
+                open={yesNoDialogOpen}
+                onClose={() => setYesNoDialogOpen(false)}
+                onYes={async () => {
```
curtisupshall, line 94: Could avoid some code duplication by moving all the `showSnackbar` and associated React code into a callback, e.g.:

```tsx
const handleSuccess = () => {
  showSnackBar({
    snackbarMessage: (
      <Typography variant="body2" component="div">
        You successfully unsecured the file{selectedArtifacts.length !== 1 && 's'}.
      </Typography>
    ),
    open: true
  });
}
```

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    </DialogContentText>
+                  </Box>
+                }
+              />
+
+              <DialogTitle id="component-dialog-title">Apply Security Reasons</DialogTitle>
+              <DialogContent sx={{ display: 'flex', flexDirection: 'column', overflow: 'hidden' }}>
+                <DialogContentText id="alert-dialog-description">
+                  Search for the security reasons and apply them to the selected document
+                  {selectedArtifacts.length > 1 ? 's' : ''}.
```
curtisupshall, line 134: This check will append an "s" when zero documents are selected. Instead the check should be `selectedArtifacts.length !== 1 ? 's' : ''`

curtisupshall (approved):
Code looks great! Two minor nitpicks

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
KjartanE: Can do that, the dialog is in accessible without selecting at least one artifact.

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
KjartanE: Will absorb into next ticket SIMSBIOHUB-108, Continuation of development for this feature

al-rosenthal (approved):
🐕

PR #191, opened 2023-06-08
SIMSBIOHUB-96: Request Access to a Secured Documents

curtisupshall:
# Overview

Allows users to request access to attachments and documents that are secured for a given dataset, so they can use the information in those documents to do their own analysis.

## Links to Jira tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-96

 - Adds the ability to request access to secured artifacts for a dataset
 - Adds the `useSecurityApi` hook
 - Adds a util method for pretty-printing phone numbers
 - Adds artifact service and repo methods


api/src/paths/artifact/security/requestAccess.ts
```diff
@@ -0,0 +1,157 @@
...
+const defaultLog = getLogger('/api/artifact/security/requestAccess');
+
+export const POST: Operation = [requestAccess()];
+
+POST.apiDoc = {
+  description: 'Request access to secure artifacts in Biohub.',
+  tags: ['documents', 'security', 'biohub'],
+  security: [
+    {
+      OptionalBearer: []
```
NickPhura, line 16: (lets double check) but I think we do want to enforce that the user must be logged in before they can request access? Currently we don't track anything, but we will eventually probably need to track the request against a system user with valid IDIR or BCeID.

api/src/services/gcnotify-service.ts
```diff
@@ -1,33 +1,93 @@
...
+    city: string;
+    postalCode: string;
+  };
+  professionalOrganization?: {
+    organizationName: string;
+    memberNumber: string;
+  };
+}
+
+interface IGcNotifyArtifactRequestAccess {
```
NickPhura, line 36: nit: Should this live in the gcnotify service? Looks like its only referenced there

NickPhura: somehow I read this as not the gcnotfiy service... disregard..

NickPhura:
@curtisupshall Was testing this locally, and noticed a few things:

- The artifact endpoint was throwing a response validation error (because `foi_reason` was missing). Changing `foi_reason_description` to `foi_reason` everywhere seemed to fix it.
- It looks like the checkboxes on the documents table don't translate into the checkboxes in the security request dialog (the ones that are selected when the dialog opens should show as already selected, I believe).
- The confidentiality agreement section when you select `no` in the dialog, shows the fields as required (asterisk) but when I click submit no validation messages appear (might just be missing the corresponding yup schemas?)

curtisupshall:
Remaining work:
 - Disable the form while it's uploading
 - TODO comments re: jsdoc
 - Test coverage for new `useApi` hooks
 - Show error dialog when request fails
 - Manual testing, sonarcloud errors, linting, formatting

api/src/paths/artifact/security/requestAccess.ts
curtisupshall: Per discussion with Jer, according to stakeholders, users do not need to be signed in to send a request.

anissa-agahchen:
Functionality works well now .. as of 5:38 pm

anissa-agahchen (dismissed):
Functionality works great .. just wondering about the TODO with for the success dialog

.envold
```diff
@@ -0,0 +1,148 @@
+# ------------------------------------------------------------------------------
```
al-rosenthal: this should probably get taken out

curtisupshall: Woops, good catch!

al-rosenthal (approved):
🍪

anissa-agahchen (approved):

PR #194, opened 2023-06-13
SIMSBIOHUB-108: Secure Documents and See Document Status When Security Rules Applied

KjartanE:
# Overview

## Links to Jira tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-108

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


api/src/paths/dwc/submission/{datasetId}/artifacts.ts
```diff
@@ -35,66 +36,71 @@ GET.apiDoc = {
...
           schema: {
-            type: 'object',
-            required: ['artifacts'],
-            properties: {
-              artifacts: {
-                type: 'array',
-                items: {
+            type: 'array',
+            items: {
+              type: 'object',
```
NickPhura, line 41: Any reason for removing the top level object?

It probably doesn't really matter at the moment, since we probably weren't sending more than the 1 artifacts field anyway, but it does mean we have no way to send additional top level data to this endpoint (if needed), without making breaking changes to add the object layer back.

KjartanE: My consideration was the endpoint is designed for list of artifacts. Adding supplementary data inside the object seemed like it made more sense. Where in SIMS we actually contained that supplementaryData in that higher order object. IE:
`{ artifact: [], supplementaryData: [] }`
With the supplementaryData inside the artifact object now, I didn't see much consideration in having the artifact object. As any extra information per artifact will be inside, and any more global information would probably be housed at a different endpoint serving a higher order standard.

api/src/paths/security/persecution-harm/apply.ts
```diff
@@ -94,15 +95,21 @@ export function applySecurityRulesToArtifacts(): RequestHandler {
       await connection.open();
 
       const securityService = new SecurityService(connection);
+      const artifactService = new ArtifactService(connection);
 
+      // Remove all security rules from artifacts.
```
NickPhura: Now or in future: We probably will want to be more fine-grain with this. For example: If an update to security rules is sent, it is better if we can selectively add the new rules, remove the missing rules, and leave untouched any rules that didn't change. Rather than blindly removing all existing rules and adding all incoming rules as net new.

Its more annoying to code it this way, but technically it makes the auditing more accurate

KjartanE: Sounds good, have time this sprint to include that.

NickPhura: We do this sort of thing in the SIMS survey-service updateSurveyPermitData() function (that code is older, so can be scrutinized obviously as needed)

PR #199, opened 2023-06-15
SIMSBIOHUB-112: document security based on roles and permission

anissa-agahchen:
# Overview

## Links to Jira tickets

- {https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-112}

This is an API only ticket that allows/prevents document downloads.
Two ways to test this PR manually:
1) use Postman
2) Temporarily turn off the front-end logic that hides the `Download Document`.  To do this:
In the file AttachmentItemMenuButton, on line 53 set disabled to false:
`disabled={props.hasAdministrativePermissions ? false : false}`


api/src/services/security-service.ts
```diff
@@ -76,4 +84,86 @@ export class SecurityService extends DBService {
...
+    ) {
+      throw new HTTP403('You do not have access to this document');
+    }
+
+    // access is granted because
+    // 1) admin
+    // 2) document is unsecured (not pending review, and has no security rules)
+    // 3) non-admin has exceptions all security rules
+    const response = await this.artifactService.getArtifactById(artifactId);
+    return await getS3SignedURL(response.key);
```
curtisupshall: Nit: can remove the `await`, since `getS3SignedUrl` and `await getS3SignedUrl` return the same thing

anissa-agahchen: @curtisupshall good catch, thank you

api/src/services/security-service.ts
```diff
@@ -76,4 +84,86 @@ export class SecurityService extends DBService {
...
+    const isSystemUserAdmin = await this.userService.isSystemUserAdmin();
+
+    const userId = this.connection.systemUserId();
+
+    const isDocumentPendingReview = (await this.artifactService.getArtifactById(artifactId)).security_review_timestamp
+      ? false
+      : true;
+
+    if (!isSystemUserAdmin && isDocumentPendingReview) {
+      throw new HTTP403('Documents that are pending review can only be downloaded by administrators.');
```
NickPhura: (observation): not really a technical thing, and probably more of a Jeremy question, but we may have to review the language on security related errors since we may or may now want to reveal why the secured item is unavailable, since that itself reveals something about its security.

api/src/services/security-service.ts
```diff
@@ -76,4 +84,86 @@ export class SecurityService extends DBService {
...
+   *
+   * Rules:
+   * non-admin user cannot access the document when:
+   ** - it is pending review, OR
+   ** - user hasn't been granted an exception to every security rule
+   *
+   * non-admin user can access the document when:
+   * - document is not secured
+   * - user has the correct exceptions
+   *
```
NickPhura: nit: I like that you added the rules here, that is exactly the kind of thing we should include in JSDoc, and inline comments (as appropriate), I think we should just be more explicit in exactly the conditions that allow someone access or not, and don't hesitate to state the obvious either. 

For example: `document is not secure` could be more like `A security review has been completed (security review is no longer pending) and the document is deemed not secure (no security rules were applied).` 

Similar for `user has the correct exceptions`. That is true IF the document has had a review completed and was found to be secure, and had security rules applied.

You could even define what a pending review is (explaining the security timestamp being null or not, etc).

So just stuff like that, being overly descriptive and exactly how these states come about.

NickPhura: I would even potentially mirror some of the JSDoc comments in the code below where they are applicable, just to make it foolproof for future developers who need to modify this code.

api/src/services/security-service.ts
```diff
@@ -76,4 +84,86 @@ export class SecurityService extends DBService {
...
+   * @param {number} userId
+   * @return {*}  {Promise<number[]>}
+   * @memberof SecurityService
+   */
+  async getPersecutionAndHarmExceptionsByUser(userId: number): Promise<number[]> {
+    defaultLog.debug({ label: 'getPersecutionAndHarmExceptionsByUser' });
+
+    return (await this.securityRepository.getPersecutionAndHarmRulesExceptionsByUserId(userId)).map(
+      (item) => item.persecution_or_harm_id
+    );
```
NickPhura, line 243: Not really critical: but I would potentially consider removing the map, and instead including it in the calling function. That way this function could satisfy anyone that wants the list of rules, but not necessarily just the IDs.  Same for the function below.

NickPhura: Even though the repo functions only return the IDs right now, technically its all backwards compatible, so if we exapanded the SQL to return more fields, this function could still return an array of objects of a defined type, and the parent could still map out just the IDs.

NickPhura: I guess there is a bit of a question around what the service level is really providing at that point, which is maybe a fair counter argument... Hhmm

NickPhura: After all that, lets leave these functions as they are. Maybe just add the word `Id` to the function name, to make it clear that the response is just the Ids of the rules.

api/src/services/security-service.ts
anissa-agahchen: Talked to Jeremy.  Changing the tex to 'You do not have access to this document'.

NickPhura (dismissed):

curtisupshall (approved):

al-rosenthal (approved):
🍕

PR #203, opened 2023-06-23
SIMSBIOHUB-117: Delete Artifact

al-rosenthal:
# Overview

## Links to Jira tickets

[SIMSBIOHUB-117](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-117)

- Added new endpoint to delete artifacts
- Wired up existing button to delete artifacts in BioHub
- Added confirmation dialog

## PR Checklist

### Code

### Style

### Documentation

### Tests

### Linting/Formatting

### SonarCloud


app/src/features/datasets/components/DatasetArtifacts.tsx
```diff
@@ -78,6 +83,30 @@ const DatasetAttachments: React.FC<IDatasetAttachmentsProps> = (props) => {
     await downloadFile(signedUrl);
   };
 
+  const handleDeleteArtifact = async (artifactUUIDs: string[]) => {
+    const data = await biohubApi.artifact.deleteArtifacts(artifactUUIDs);
+    artifactsDataLoader.refresh();
+
+    console.log(data);
```
KjartanE: nit

al-rosenthal: removed

KjartanE (dismissed):
🐈 Tested full functionality across applications. Functioning as intended

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,125 @@
...
+  responses: {
+    200: {
+      description: '',
+      content: {
+        'application/json': {
+          schema: {
+            type: 'object',
+            required: ['success'],
+            properties: {
+              success: {
```
curtisupshall: Could just make the entire response be `true` or `false`, to be consistent with other endpoints that report success/failure in this way.

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,125 @@
...
+      const service = new ArtifactService(connection);
+
+      await service.deleteArtifacts(req.body.artifactUUIDs);
+    } catch (error: any) {
+      defaultLog.error({ label: 'deleteArtifact', message: 'error', error });
+      success = false;
+      await connection.rollback();
+    } finally {
+      connection.release();
+      res.status(200).json({ success });
```
curtisupshall: If `deleteArtifact` throws, we should probably respond with a 500 status; If this method doesn't throw, but the query doesn't update any rows, then the endpoint should respond with a 200 status and a response body of `false`.

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Deletes artifacts for a given array of UUIDs.
+ * This will always respond with a JSON object {success: boolean} indicating
+ * if the artifacts have been successfully removed or not
+ *
+ * @returns {RequestHandler}
+ */
+export function deleteArtifact(): RequestHandler {
+  return async (req, res) => {
+    defaultLog.debug({ label: 'deleteArtifact', message: 'request body', req_body: req.query });
+    let connection: IDBConnection;
```
curtisupshall: Could use a ternary here

al-rosenthal: good call

api/src/paths/artifact/delete.ts
al-rosenthal: good call, I've added a throw and modified the response to just be a boolean

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,115 @@
...
+        }
+      ]
+    };
+  }),
+  deleteArtifact()
+];
+
+POST.apiDoc = {
+  description:
+    'Deletes artifacts. This will always return an object { success: boolean } to indicate if the deletion was successful or not.',
```
NickPhura: super nit: we don't need to specify the response here, since its defined in the responses section below.

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,115 @@
...
+      content: {
+        'application/json': {
+          schema: {
+            type: 'boolean',
+            description: 'A boolean indicating if the delete action was successful or not.'
+          }
+        }
+      }
+    },
+    400: {
```
NickPhura, line 72: We should technically include 403/401 here as well, to cover if the caller isnt authenticated or doesnt have authorization.

api/src/paths/artifact/delete.ts
```diff
@@ -0,0 +1,115 @@
...
+    const connection = sourceSystem
+      ? getServiceAccountDBConnection(sourceSystem)
+      : getDBConnection(req['keycloak_token']);
+
+    try {
+      await connection.open();
+      const service = new ArtifactService(connection);
+
+      await service.deleteArtifacts(req.body.artifactUUIDs);
+      res.status(200).json(true);
```
NickPhura, line 111: Don't forget the `await connection.commit()` call.

api/src/paths/artifact/intake.ts
```diff
@@ -154,35 +154,31 @@ export function intakeArtifacts(): RequestHandler {
...
-      res.status(200).json(response);
-
-      await connection.commit();
-    } catch (error) {
-      defaultLog.error({ label: 'intakeArtifacts', message: 'error', error });
-      await connection.rollback();
-      throw error;
-    } finally {
-      connection.release();
+    if (sourceSystem) {
```
NickPhura: I think this breaks the openapi 200 definition.

If the sourceSystem is null, then we return nothing (presumably it defaults to a generic 200?) which wouldn't match the openapi as it currently exists.

api/src/paths/dwc/submission/queue.ts
```diff
@@ -160,28 +160,24 @@ export function queueForProcess(): RequestHandler {
...
-
-      await connection.commit();
-      res.status(200).json(queueRecord);
-    } catch (error) {
-      defaultLog.error({ label: 'queueForProcess', message: 'error', error });
-      await connection.rollback();
-      throw error;
-    } finally {
-      connection.release();
+    if (sourceSystem) {
```
NickPhura: Possibly same comment: https://github.com/bcgov/biohubbc-platform/pull/203/files#r1246913287

api/src/repositories/security-repository.ts
```diff
@@ -160,6 +160,26 @@ export class SecurityRepository extends BaseRepository {
     await this.connection.sql(sqlStatement);
   }
 
+  /**
+   * Delete persecution rules for a given artifact uuid
```
NickPhura: Debated if this was a valid comment or not, certainly subjective, so feel free to ignore: my instinct would be to add "if any exist" here, to help explain why we don't have any error handling, or just to be more verbose in case its helpful.

Something like: "Deletes all persecution and harm rules, if any exist, for a given artifact uuid"

api/src/services/artifact-service.ts
```diff
@@ -144,4 +146,45 @@ export class ArtifactService extends DBService {
...
+    defaultLog.debug({ label: 'deleteArtifact' });
+
+    const artifact = await this.artifactRepository.getArtifactByUUID(uuid);
+
+    if (artifact) {
+      try {
+        const service = new SecurityRepository(this.connection);
+        await service.deleteSecurityRulesForArtifactUUID(uuid);
+
+        await this.artifactRepository.deleteArtifactByUUID(uuid);
```
NickPhura, line 182: You could also just call the `deleteArtifacts([uuid])` function above. But really not sure which is better, or if it really matters tbh.

al-rosenthal: there is only a single repo function for deleting an artifact, I didn't create a batch delete in the repo because of the S3 object delete

api/src/repositories/security-repository.ts
al-rosenthal: updated

api/src/paths/artifact/intake.ts
al-rosenthal: reverted changes

api/src/paths/dwc/submission/queue.ts
al-rosenthal: reverted changes

api/src/paths/artifact/delete.ts
al-rosenthal: good catch

api/src/paths/artifact/delete.ts
al-rosenthal: added

api/src/paths/artifact/delete.ts
al-rosenthal: updated

api/src/repositories/security-repository.ts
```diff
@@ -161,7 +161,7 @@ export class SecurityRepository extends BaseRepository {
   }
 
   /**
-   * Delete persecution rules for a given artifact uuid
+   * Deletes all persecution and hard rules for a given UUID if they exists
```
NickPhura: hmm... 😜

NickPhura: Also, I guess this should really say, 'deletes all security rules' (we only have persecution rules atm, but I suppose that will eventually change when custom rules, etc, get added. Didn't think of that before.

api/src/services/artifact-service.ts
NickPhura: I was thinking the service function above, that iterates over an array of uuids. But in hindsight, theres really no point in calling it over the repo function directly. So ignore this!

NickPhura (dismissed):
👍

KjartanE (dismissed):
🐈

KjartanE (approved):
😹

NickPhura (approved):

PR #204, opened 2023-06-26
SIMSBIOHUB-110: Manage Previously Applied Security Rules for Documents

curtisupshall:
# Overview
Allows Data Administrator to be able to see which security reasons were previously applied to a given document.

## Links to Jira tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-110

 - Adds the `pluralize` Util method
 - Provides an array of the applied security rules in an artifact record's _supplemental data_ section;
 - Adds the `loading` prop to YesNo dialog button props
 - Shows which security rules are applied to a document when applying security
 - Refactors the Apply Security dialog


app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -79,31 +143,52 @@ const ApplySecurityDialog: React.FC<IApplySecurityDialog> = (props) => {
...
+                handleShowSnackBar(
+                  `You successfully applied security reasons to ${selectedArtifacts.length} ${p(
+                    selectedArtifacts.length,
+                    'file'
+                  )}.`
+                );
+                onClose();
+              })
+              .catch(() => {
+                // @TODO show an error dialog
```
KjartanE: nit

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -79,31 +143,52 @@ const ApplySecurityDialog: React.FC<IApplySecurityDialog> = (props) => {
...
+                onYes={() => {
+                  handleSubmit({ securityReasons: [] })
+                    .then(() => {
+                      handleShowSnackBar(
+                        `You successfully unsecured ${selectedArtifacts.length} ${p(selectedArtifacts.length, 'file')}.`
+                      );
+                      onClose();
+                    })
+                    .catch(() => {
+                      // @TODO show an error dialog
```
KjartanE: nit

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
```diff
@@ -13,43 +14,106 @@ import YesNoDialog from 'components/dialog/YesNoDialog';
 import { DialogContext, ISnackbarProps } from 'contexts/dialogContext';
 import { Formik, FormikProps } from 'formik';
 import { useApi } from 'hooks/useApi';
-import { IArtifact } from 'interfaces/useDatasetApi.interface';
+import useDataLoader from 'hooks/useDataLoader';
+import { IArtifact, IPersecutionAndHarmRule } from 'interfaces/useDatasetApi.interface';
 import React, { useContext, useRef, useState } from 'react';
+import { pluralize as p } from 'utils/Utils';
```
KjartanE, line 21: great idea, for utils function! Is there any reasons for casting as `p`? Only a nit, but sometimes nice to see function name to understand intention without scrolling up. Either way nice work!

KjartanE (dismissed):
🐈 Looks great! Tested functionality :D

app/src/features/datasets/components/security/ApplySecurityDialog.tsx
curtisupshall: Cheers! I shortened it to `p` as to increase the readibility of the template strings; Throwing in `pluralize(1, 'word')` breaks up the sentence a bit, compared to `p(1, 'word')`

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
```diff
@@ -134,12 +152,27 @@ export function getArtifactsByDatasetId(): RequestHandler {
       const artifacts = await artifactService.getArtifactsByDatasetId(datasetId);
 
       const artifactsWithRules = await Promise.all(
-        artifacts.map(async (artifact) => {
-          const appliedSecurity = await securityService.getSecurityAppliedStatus(artifact.artifact_id);
-          return {
-            ...artifact,
-            supplementaryData: { persecutionAndHarm: appliedSecurity }
-          };
+        artifacts.map((artifact) => {
```
al-rosenthal: this seems like a lot of logic for an endpoint to have, might be better wrapped in a new service function

app/src/components/dialog/YesNoDialog.tsx
```diff
@@ -1,4 +1,4 @@
-import Button, { ButtonProps } from '@mui/material/Button';
+import LoadingButton, { LoadingButtonProps } from '@mui/lab/LoadingButton';
```
al-rosenthal, line 1: would it make sense to at this point to remove the old `LoadingButton` component and replace it with the mui one?

app/src/features/datasets/components/AttachmentItemMenuButton.tsx
```diff
@@ -48,66 +49,65 @@ const AttachmentItemMenuButton: React.FC<IAttachmentItemMenuButtonProps> = (prop
...
+                    data-testid="attachment-action-menu-download">
+                    <ListItemIcon>
+                      <Icon path={mdiLockPlus} size={0.875} />
+                    </ListItemIcon>
+                    Apply Security to Document
+                  </MenuItem>
+
+                  <MenuItem
+                    onClick={() => {
+                      console.log('Delete artifact not implemented yet.');
```
al-rosenthal, line 99: nit

curtisupshall: This gets addressed in https://github.com/bcgov/biohubbc-platform/pull/203/files#diff-12fa4d97ebcdda177e6b9872e35f0e2fc2b2b0e8e57955e64920b86dcb302e75

app/src/components/dialog/YesNoDialog.tsx
curtisupshall: The `LoadingButton` component you're referring to is actually in SIMS, not in this repo. We can't remove the SIMS `LoadingButton` component until we update MUI to version 5.

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
curtisupshall: I decided to tackle this code in the endpoint as to pollute the Artifacts Service with one-off methods that are only used by one endpoint. To that end, I think that is especially relevant in this case because 1) endpoints will cobble together supplementary data in different ways, 2) the code doesn't have any side effects (it doesn't modify any records in the database), and 3) its job is to simply enrich the records that the endpoint is returning, which is a common task for endpoints.

app/src/components/dialog/YesNoDialog.tsx
al-rosenthal: 🤡

al-rosenthal (dismissed):
💻

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
NickPhura: Good points.

A couple of notes:
- It would be my preference to use async/await here ideally, just because its a bit less complex to read/understand (less nesting).
- More optional, but we could move the inner chunk of code into its own function within this file, that constructs and returns the supplementaryData object. It could still assemble the response here in the map in the same way. Mainly just as a way to make the endpoint a bit smaller/easier to read (more modular), and the supplementary data code documentable and testable in isolation.

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
```diff
@@ -134,12 +152,27 @@ export function getArtifactsByDatasetId(): RequestHandler {
...
+            .getPersecutionAndHarmRulesByArtifactId(artifact.artifact_id)
+            .then((persecutionAndHarmRules) => {
+              let persecutionAndHarmStatus: SECURITY_APPLIED_STATUS = SECURITY_APPLIED_STATUS.PENDING;
+
+              if (artifact.security_review_timestamp) {
+                persecutionAndHarmStatus =
+                  persecutionAndHarmRules.length > 0
+                    ? SECURITY_APPLIED_STATUS.SECURED
+                    : SECURITY_APPLIED_STATUS.UNSECURED;
+              }
```
NickPhura: Regarding `persecutionAndHarmStatus`, this comment is probably relevant here as well for the same reasons: https://github.com/bcgov/biohubbc-platform/pull/205/files#r1246871748

app/src/utils/Utils.ts
```diff
@@ -319,3 +319,21 @@ export const buildUrl = (...urlParts: (string | undefined)[]): string => {
...
+ * @example p(1, 'banana'); // => 'banana'
+ * @example p(10, 'berr', 'y', 'ies'); // => 'berries'
+ *
+ * @param quantity The quantity used to infer plural or singular
+ * @param word The word to pluralize
+ * @param {[string]} singularSuffix The suffix used for a singular item
+ * @param {[string]} pluralSuffix The suffix used for plural items
+ * @returns
+ */
+export const pluralize = (quantity: number, word: string, singularSuffix = '', pluralSuffix = 's') => {
```
NickPhura, line 337: Nice one

app/src/utils/Utils.test.ts
```diff
@@ -412,3 +413,30 @@ describe('downloadFile', () => {
...
+describe('pluralize', () => {
+  it('pluralizes a word', () => {
+    const response = pluralize(2, 'apple');
+    expect(response).toEqual('apples');
+  });
+
+  it('pluralizes a word with undefined quantity', () => {
+    const response = pluralize(null as unknown as number, 'orange');
+    expect(response).toEqual('oranges');
+  });
```
NickPhura, line 426: This is hardly critical, and entirely subjective, so feel free to ignore: my instinct with a null/undefined value would be that it just returns the original word unchanged (when in doubt, do nothing). null/undefined is maybe analogous to a count of 0, in this context.

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
KjartanE: If this PR gets merged, I was going to handle on SIMSBIOHUB-10

app/src/utils/Utils.test.ts
curtisupshall: In my opinion, the word should always be plural even if the quantity is undefined, since without this function, the word would typically be written in plural anyway. And since `null` or `undefined` quantities are synonymous with having zero of something, to me it makes sense to pluralize the word the same way you would for a quantity of `0`.

api/src/paths/dwc/submission/{datasetId}/artifacts.ts
curtisupshall: Refactored to use `async`

KjartanE (approved):
🐈

al-rosenthal (approved):
🔒

PR #205, opened 2023-06-28
SIMSBIOHUB-10: pending review status for related datasets

anissa-agahchen:
# Overview

## Links to Jira tickets

[ **SIMSBIOHUB-10**](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-10)

- Added supplementary data that reflects if the related dataset has documents pending review
- Front end changes to show pending review status when applicable


api/src/services/security-service.ts
```diff
@@ -272,4 +272,23 @@ export class SecurityService extends DBService {
...
+   * @memberof SecurityService
+   */
+  async isDatasetPendingReview(datasetId: string): Promise<boolean> {
+    const artifactIds = (await this.artifactService.getArtifactsByDatasetId(datasetId)).map((item) => item.artifact_id);
+
+    const artifactSecurityRules = await Promise.all(
+      artifactIds.map(async (artifactId) => await this.isArtifactPendingReview(artifactId))
+    );
+
+    const isPendingReview = artifactSecurityRules.find((item) => item === true) ? true : false;
```
curtisupshall: I would change this to
```javascript
const isPendingReview = artifactSecurityRules.includes(true)
```

api/src/services/submission-service.ts
```diff
@@ -419,13 +425,15 @@ export class SubmissionService extends DBService {
       return [];
     }
 
-    return result[0].value.map((relatedProject: any) => {
+    const relatedDatasets = result[0].value.map((relatedProject: any) => {
```
curtisupshall: I would revert this change - guessing this is a testing artifact

curtisupshall (dismissed):
Code looks good! Two minor nitpicks

api/src/paths/dwc/submission/{datasetId}/related.ts
```diff
@@ -80,12 +90,24 @@ export function getRelatedDatasetsByDatasetId(): RequestHandler {
...
+        datasets.map(async (dataset) => {
+          const isDatasetPendingReview = await securityService.isDatasetPendingReview(dataset.datasetId);
+
+          return {
+            ...dataset,
+            supplementaryData: { isPendingReview: isDatasetPendingReview }
+          };
+        })
+      );
+
```
NickPhura: Discussed with Anissa, but never wrote down. So adding here just in case:

Its not clear, but I think we can assume based on the frontend logic, that we don't want the API to reveal any information about the artifact if it is in a pending state, to non-admin users.

So this supplementary data section, should really only return `isPendingReview` if the calling user is an admin, and return an empty object if they are a non admin. The openapi and frontend types/interfaces would need to be updated accordingly of course, to make `isPendignReview` optional.

This same logic should also apply to our other Artifact endpoint that returns the `SECURE`, `UNSECURE`, `PENDING` statuses. If the user is an admin, no change. But if they are a non admin, we should really only return `SECURE` or `UNSECURE` (if the status is PENDING, we would return SECURE instead). This exact logic is handled this way on the frontend, but really should be a backend task, and the frontend should just display the status without having to check what role the user has. So the backend will get slightly more complicated, but the frontend will get simpler, and overall the API response data is more secure.

al-rosenthal (approved):
ran locally, looks good 🌞

curtisupshall (approved):
Code looks good 🌳

PR #209, opened 2023-11-15
SIMSBIOHUB-374: BioHub Feature Submission Schema

NickPhura:
# Overview

Note: Includes breaking migration changes, so we will need to drop dev before merging this. Also, the api/app is likely going to be broken due to having removed lots of old database tables.

## Links to Jira tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-374

New migrations to add new feature/search/security/message tables. These tables are all net new. The one `feature_submission` table does hang off the existing `submission` table.

New see to populate the new lookup tables (with some realistic, but not final, values).

Removed all old biohub tables that we either know we won't need, or are likely to not need. I just updated the existing migrations, so we will need to do a clean wipe of dev before merging this PR.

Removed all code related to the database views, including the biohub_dapi_v1 user.
- Why? It seemed like they weren't providing any value. I believe the original theory behind them was that the biohub_api user would have reduced permissions to mess with the database, but in testing some of the database roles, etc, it seemed like the biohub_api user had more-or-less the same permissions as the regular postgres user. So in an effort to simplify things, and maybe improve performance (maybe negligibly), I just removed them, and gave the biohub_api user full permissions. We aren't operating in an environment where users have their own database accounts - everything goes through the API and it may as well have full powers.

I also removed lots of references to `DB_SCHEMA` as the `biohub` schema may as well be hardcoded, given the earliest migrations all reference it directly anyways, so its not really configurable.

Added a seed to insert mock test data (disabled by default).


Makefile
```diff
@@ -56,6 +56,14 @@ clean: ## Closes and cleans (removes) all project containers
 	@echo "==============================================="
 	@docker-compose -f docker-compose.yml down -v --rmi all --remove-orphans
 
+prune: ## Deletes ALL docker artifacts (even those not associated to this project)
```
curtisupshall, line 59: Awesome!

curtisupshall (commented):
Code looks good, but when I build I get the following error on the database container:
```
2023-11-15 23:52:30.133 PST [1] LOG:  starting PostgreSQL 12.5 (Debian 12.5-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
2023-11-15 23:52:30.133 PST [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2023-11-15 23:52:30.133 PST [1] LOG:  listening on IPv6 address "::", port 5432
2023-11-15 23:52:30.142 PST [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2023-11-15 23:52:30.165 PST [85] LOG:  database system was shut down at 2023-11-15 23:52:30 PST
2023-11-15 23:52:30.171 PST [1] LOG:  database system is ready to accept connections
2023-11-15 23:52:55.132 PST [137] ERROR:  permission denied for table user_identity_source
2023-11-15 23:52:55.132 PST [137] CONTEXT:  SQL statement "select user_identity_source_id from user_identity_source where name = p_user_identity_source_name and record_end_date is null"
        PL/pgSQL function api_set_context(character varying,character varying) line 21 at SQL statement
2023-11-15 23:52:55.132 PST [137] STATEMENT:  select api_set_context($1, $2);
```

Makefile
NickPhura: Also exists in sims, just porting it over

curtisupshall: I had forgotten 😅

NickPhura:
> Code looks good, but when I build I get the following error on the database container:
> 
> ```
> 2023-11-15 23:52:30.133 PST [1] LOG:  starting PostgreSQL 12.5 (Debian 12.5-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
> 2023-11-15 23:52:30.133 PST [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
> 2023-11-15 23:52:30.133 PST [1] LOG:  listening on IPv6 address "::", port 5432
> 2023-11-15 23:52:30.142 PST [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
> 2023-11-15 23:52:30.165 PST [85] LOG:  database system was shut down at 2023-11-15 23:52:30 PST
> 2023-11-15 23:52:30.171 PST [1] LOG:  database system is ready to accept connections
> 2023-11-15 23:52:55.132 PST [137] ERROR:  permission denied for table user_identity_source
> 2023-11-15 23:52:55.132 PST [137] CONTEXT:  SQL statement "select user_identity_source_id from user_identity_source where name = p_user_identity_source_name and record_end_date is null"
>         PL/pgSQL function api_set_context(character varying,character varying) line 21 at SQL statement
> 2023-11-15 23:52:55.132 PST [137] STATEMENT:  select api_set_context($1, $2);
> ```

Good catch.

I've pushed a change to fix this. The issue was that I assumed "GRANT ALL ON ..." would do the trick, but apparently that only works on currently existing items and doesn't work for items created afterwards. It now changes the default permissions for the biohub_api user, so it should have proper permissions for all future items as well.

database/src/migrations/20231109000002_search_tables.ts
```diff
@@ -0,0 +1,226 @@
...
+ * @return {*}  {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    ----------------------------------------------------------------------------------------
+    -- Create tables
+    ----------------------------------------------------------------------------------------
+    set search_path=biohub,public;
+
+    CREATE TABLE search_string(
```
al-rosenthal, line 23: nit: it would be great to see a short explanation for these new tables. Without knowing what each table is supposed to be doing it's hard to understand the architecture by just looking at them

NickPhura:
@AlfredRosenthal  Re Architectural Overview

I should probably move this somewhere more formal, and maybe add some pictures, or something, but hopefully this helps a bit for now.

The schema is basically divided into 4 chunks (not including users and messages, and other random tables): Data, Types/Properties, Search, Security.

## Data
**- submission** (An entire submission record. Ie: a Survey).
**- submission_feature** (A single typed element of a submission. Each row represents 1 item (such as a single observation, or single animal. We also expect there to always be exactly 1 'dataset' element, representing the general metadata of the dataset))

## Types/Properties
These lookup tables will be used to provide context to the search_* fields, and also to provide our rules for validating incoming data.
**- feature_type** (Types of submission_features. Ex: dataset, observation, animal, sample site, etc)
**- feature_type_property** (Join table. What properties can a given feature type have. Ex: datasets can have 'name' and 'method', while observations can have 'count', 'sex', and 'lifestage')
**- feature_property** (Known/recognized properties. Ex: name, method, start_date, count, sex, lifestage, description, lat, lng, etc)
**- feature_property_type** (Types of properties. Ex: string, number, boolean, datetime, spatial, etc)

## Search
The search tables contain all of the key/value pairs parsed out of the submission_feature properties, based on the known/recognized properties for that feature type. The search tables are split into multiple tables so that the `value` column in each one can be strongly typed. Each row in one of these tables will point to one of the feature_property rows, giving context to what it is. Ex. A search_number record might have a value of 50, and a feature_property id that points to the 'count' feature_property. Since this search_number record also points to a submission_feature (presumably of type 'observation') we can now run a sql query to "find all observation records where 'count' is >= 50".
**- search_string**
**- search_number**
**- search_datetime**
**- search_spatial**

## Security
**- security_rule** (Contains a list of named rules as defined by the admins. Ex: 'Female Moose in Skeena')

Similar to the search tables, the security condition tables are broken out into multiple tables so that each one can be strongly typed, and only worry about executing the rule against a specific type.
**- security_string** (Is a child of 1 security rule. Defines security conditions on string values from the search_string table.)
**- security_number** (Is a child of 1 security rule. Defines security conditions on number values from the search_number table.)
**- security_datetime** (Is a child of 1 security rule. Defines security conditions on datetime values from the search_datetime table.)
**- security_spatial** (Is a child of 1 security rule. Defines security conditions on spatial values from the search_spatial table.)

### Security Functions
**- evaluate_security_rule** (Given the id of a submission_feature, runs all security rules against the feature, and returns those rules where ALL conditions returned true. This function runs all 4 of the below functions.)

Each security condition function takes the name of a security condition rule, and the id of a submission_feature, and returns true if it applies. The functions themselves do a bit of dynamic sql generation allowing for a bit of query building for the rules (they read the comparator and value columns of the security_* row to generate sql).
**- evaluate_security_string_condition** 
**- evaluate_security_number_condition** 
**- evaluate_security_datetime_condition**
**- evaluate_security_spatial_condition**

KjartanE (approved):
Ran locally, works as intended

MacDeluca (approved):
Looks good! Built successfully from fresh install

PR #211, opened 2023-12-06
SIMSBIOHUB-379: Parse intake key-value pairs

curtisupshall:
# Overview

## Links to Jira tickets

 - [SIMSBIOHUB-379](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-379)

- Adds service and repo methods to parse search fields from submission features


api/src/services/search-index-service.ts
```diff
@@ -0,0 +1,62 @@
...
+  searchIndexRepository: SearchIndexRepository;
+
+  constructor(connection: IDBConnection) {
+    super(connection);
+
+    this.searchIndexRepository = new SearchIndexRepository(connection);
+  }
+
+  async indexFeaturesBySubmissionId(submissionId: number): Promise<void> {
+    const submissionService = new SubmissionRepository(this.connection);
```
NickPhura: nit: `submissionRepository`

api/src/services/search-index-service.ts
```diff
@@ -0,0 +1,62 @@
...
+export class SearchIndexService extends DBService {
+  searchIndexRepository: SearchIndexRepository;
+
+  constructor(connection: IDBConnection) {
+    super(connection);
+
+    this.searchIndexRepository = new SearchIndexRepository(connection);
+  }
+
+  async indexFeaturesBySubmissionId(submissionId: number): Promise<void> {
```
NickPhura, line 34: Looks good! 

We will probably have to consider nested properties eventually, where the property feature type is 'array' or 'object. But I think we can just assume the properties are flat for now, in order to get the first pass out the door.

NickPhura:
@curtisupshall I pushed some changes into this PR, mostly just around updating the intake endpoint to work with the latest structure being sent from SIMS (SIMS changes are in this open PR https://github.com/bcgov/biohubbc/pull/1185). Once the SIMS PR goes through, we should be able to run the sims publish and see the BioHub intake get hit, including the validation and a call to your parsing code.

api/src/services/search-index-service.ts
```diff
@@ -0,0 +1,97 @@
...
+
+          case 'string':
+            stringRecords.push({ submission_feature_id, feature_property_id, value: value as string });
+            break;
+        }
+      });
+    });
+
+    if (datetimeRecords.length) {
+      this.searchIndexRepository.insertSearchableDatetimeRecords(datetimeRecords);
```
NickPhura: These need `awaits`

api/src/services/search-index-service.ts
```diff
@@ -0,0 +1,97 @@
...
+    const featurePropertyTypeMap: Record<string, FeaturePropertyRecordWithPropertyTypeName> = Object.fromEntries(
+      featurePropertyTypeNames.map((propertyType) => {
+        const { name } = propertyType;
+        return [name, propertyType];
+      })
+    );
+
+    features.forEach((feature) => {
+      const { submission_feature_id } = feature;
+      Object.entries(feature.data.properties).forEach(([feature_property_name, value]) => {
```
NickPhura: Should just be `feature.data` rather than `feature.data.properties`

api/src/repositories/search-index-respository.ts
```diff
@@ -0,0 +1,222 @@
...
+   * @memberof SearchIndexRepository
+   */
+  async insertSearchableSpatialRecords(
+    spatialRecords: InsertSpatialSearchableRecord[]
+  ): Promise<SpatialSearchableRecord[]> {
+    defaultLog.debug({ label: 'insertSearchableSpatialRecords' });
+
+    const queryBuilder = getKnex().queryBuilder().insert(spatialRecords).into('search_spatial').returning('*');
+
+    const response = await this.connection.knex<SpatialSearchableRecord>(queryBuilder);
```
NickPhura: The generic type should be the second parameter in the function call: `this.connection.knex(queryBuilder, SpatialSearchableRecord);`

The way it is now, it just forces the result type without actually running the zod check.

NickPhura: Got an error when the code reached here: the issue is that the incoming geojson data is being inserted into the `geometry` column, which is the postgis geometry type column. 

The raw geojson would need to go in the `geojson` column, and then we also need to populate the `geometry` column with the encoded value. Can see similar examples of converting raw geojson into postgis geometry if you search for `ST_GeomFromGeoJSON`.

NickPhura: Just realized/starting to remember that we actually can't convert GeoJSON Feature or FeatureCollection types to postgis geometry. It only accepts regular type (polygon, point, etc) and GeometryCollection. We actually have a function in SIMS `generateGeometryCollectionSQL` for this exact reason. So we either need to steal it or create a better version and leverage that here.

api/src/services/search-index-service.ts
curtisupshall: Good catch

api/src/services/search-index-service.ts
curtisupshall: Fixed

api/src/repositories/search-index-respository.ts
curtisupshall: Good catch, fixed.

curtisupshall: Solved.

al-rosenthal (commented):
Running this locally I've run into an issue publishing a survey without an end date. The UI doesn't force a survey end date to be specified but the intake in BioHub fails when trying to insert into the `search_datetime` table. Should end date be required in SIMS? will that be applicable for other submitters?

curtisupshall:
> Running this locally I've run into an issue publishing a survey without an end date. The UI doesn't force a survey end date to be specified but the intake in BioHub fails when trying to insert into the `search_datetime` table. Should end date be required in SIMS? will that be applicable for other submitters?

Good catch! Committed a change which ignores date values that are `null`.

@NickPhura thoughts on this approach?

api/src/paths/submission/search-idx.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+      required: true
+    }
+  ],
+  responses: {
+    200: {
+      description: 'TODO', // TODO
+      content: {
+        'application/json': {
+          schema: {
```
KjartanE, line 51: api doc still required todo

app/src/hooks/api/useDatasetApi.ts
```diff
@@ -32,6 +32,19 @@ const useDatasetApi = (axios: AxiosInstance) => {
...
 
+  /**
+   * Fetch dataset data by datasetUUID.
+   *
+   * @param {string} datasetUUID
+   * @return {*}  {Promise<any>}
+   */
+  const getDataset = async (datasetUUID: string): Promise<any> => {
+    const { data } = await axios.get(`api/dataset/${datasetUUID}`);
+    console.log('data', data);
```
KjartanE: nit: remove

api/src/services/artifact-service.ts
```diff
@@ -84,7 +84,11 @@ export class ArtifactService extends DBService {
     });
 
     // Create a new submission for the artifact collection
-    const { submission_id } = await this.submissionService.insertSubmissionRecordWithPotentialConflict(dataPackageId);
+    const { submission_id } = await this.submissionService.insertSubmissionRecordWithPotentialConflict(
+      dataPackageId,
+      'TODO_Temp',
```
KjartanE, line 89: Is this function being used? could we delete the entire function?

NickPhura: There is a lot of old code we can remove. We were just holding off so as no to make actual feature PRs bigger/more complex.

NickPhura: That being said, removing small bits as needed, if it is hindering progress on current work, is totally fine.

api/src/paths/submission/intake.ts
```diff
@@ -40,26 +42,28 @@ POST.apiDoc = {
...
+              title: 'A description of the submission. Should not include sensitive information.',
+              type: 'string',
+              maxLength: 3000
             },
             features: {
               type: 'array',
               items: {
                 $ref: '#/components/schemas/SubmissionFeature'
               },
+              maxItems: 1,
```
KjartanE, line 66: why maxItems? This endpoint should accept all features connected to submission

app/src/hooks/api/useDatasetApi.ts
curtisupshall: Removed

al-rosenthal (approved):
running this locally and everything looks good 🕶️

PR #214, opened 2023-12-19
Header Improvements

curtisupshall:
# Overview
Added header improvements from SIMS.


al-rosenthal (commented):
I think @jeznorth updated the available headers with his last UI update. I think the header should just have Home, Submissions, Manage Users. Header changes look good otherwise

al-rosenthal (approved):
looks good 🎄

app/src/components/layout/header/Header.tsx
```diff
@@ -1,27 +1,25 @@
-import { mdiHelpCircle } from '@mdi/js';
+import { mdiMenu } from '@mdi/js';
 import Icon from '@mdi/react';
-import { Theme } from '@mui/material';
+import { Menu, MenuItem, /*Theme*/ } from '@mui/material';
```
KjartanE: nit: remove

KjartanE (approved):
🐈 Ran locally, functioning properly

app/src/components/layout/header/Header.tsx
curtisupshall: Good catch, removed

PR #219, opened 2024-01-02
TechDebt: Port Auth Changes From SIMS + Refactor Service Client Auth

NickPhura:
# Overview

## Links to Jira tickets

n/a

Port Auth changes from SIMS (newer keycloak libraries, etc)

Refactor service client authentication
- No longer relies on any hard-coded keycloak service client values, everything is read in from the database.
- Added new `db-constants.ts` file, which is used to load constants from the database during API initialization (currently only loads service client information, but can be updated to support other constants as needed).

Populate submission source_system column from live submission data.
- Was hard coded before.

Update seed to mark a pseudo-random number of submissions as reviewed, and a subset of those as published.

UI change for Jer: Change default switch case for security review status to UNSECURED (rather than PENDING), so that the user only ever sees either UNSECURED, SECURED, or PARTIAL status.

## Testing

Everything related to authentication works as before, including service clients.
You will need to run a clean build to get the latest database changes.

Have tested the submission of SIMS data to Biohub, and it seems to be working as expected as far as I can tell.


curtisupshall (approved):
Looks good to me 🛺

al-rosenthal (approved):
ran this locally, looks good!

KjartanE (approved):
🐈 ran locally, functioning properly

PR #227, opened 2024-01-05
SIMSBIOHUB-428: Apply Security to Different Features

curtisupshall:
# Overview
Apply security rules to selected features.
## Links to Jira tickets

 - https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-428

## Testing
You should be able to apply and remove multiple security rules to any number of selected features.

 - Updated endpoints and service and repo methods to support applying security to selected features.
 - Refactored the frontend to support selecting features, applying security rules, and to see already applied security rules.

### Removed endpoints
| **Endpoint**                     | **Purpose**                                                                                                                                                                                         |
|----------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `.../security/apply.ts`          | Was used to apply and override security rules to features. This has been replaced.                                                                                                                  |
| `.../security/category/fetch.ts` | Was used to fetch all security rule rules *with* their associated. This has been replaced.                                                                                                           |
| `.../security/remove.ts`         | Was used to remove security rules from a set of features. Has been superseded by `.../security/rules`.                                                                                              |
| `.../security/index`             | Was originally used to get all security rules. Because we don't currently have a use for getting rules *without* their categories, this endpoint has been removed in favor of `.../security/rules`. |

### Moved endpoints
| **Endpoint**                                                | **Purpose**                                 |
|-------------------------------------------------------------|---------------------------------------------|
| `...security/category/index` → `.../security/categories.ts` | Used to list all active security categories |

### New endpoints

| **Endpoint**                                | **Purpose**                                                                                                                            |
|---------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------|
| `.../security/rules`                        | Used to fetch all security rule rules *with* their associated                                                                          |
| `.../security/submission/{submissionId}.ts` | Used to list already-applied security rules, and to apply/remove security for any number of features belonging to the given submission |


NickPhura:
@curtisupshall 
#### Running Locally

Just tried adding/removing security to all/some records randomly, and it all seems to function correctly as far as I can tell. 
The UI for adding/removing security rules to all or individual records looks pretty slick :)
Combined with the review/publish statuses, and how they toggle, the whole thing in general is looking really nice.

---

I did noticed a few things though (no idea if they were there before, or introduced in this PR):

#### Add/edit security dialog
If I select a security rule (staged), and then remove it (unstage it), if I go to select it again it doesn't get added to the list.

#### Add/edit security dialog
This console warning is thrown when selecting a security rule:
```
MUI: The value provided to Autocomplete is invalid.
None of the options match with `{"security_rule_id":1,"name":"Caribou location","description":"Caribou locations are secured","record_effective_date":"2023-06-13","record_end_date":null,"security_category_id":1,"category_name":"Government Interest","category_description":"Government Interest","category_record_effective_date":"2023-01-14","category_record_end_date":null}`.
You can use the `isOptionEqualToValue` prop to customize the equality test. 
Autocomplete@http://localhost:7207/static/js/bundle.js:37578:83
div
...
...
```

--- 

#### Openapi Spec
On the new security endpoints, it looks like the openapi definition isn't quite correct, but because none of the fields are `required` it is silently working. We should mark all fields as `required` (if appropriate). And then you'll probably have to update the spec a bit to match the response. I think maybe some of the descriptions are copy/pasted as well and could be tweaked. A few super minor nits as well: could make a few numbers into integers and add the `minimum: 1`.

#### Duplicate Request
I notice that when I add/edit security and click save. The patch request is kicked off, and a few other requests are run to refresh the content, and I noticed that one of those requests runs twice every time (the others all run just once): `api/administrative/security/submission/{submissionId}`.

app/src/components/security/SecurityRuleForm.tsx
```diff
@@ -5,50 +5,141 @@ import Autocomplete, { createFilterOptions } from '@mui/material/Autocomplete';
...
+    }
   };
 
+  const applyRulesAvailableSortedOptions = useMemo(() => {
+    return alphabetizeObjects(allSecurityRules, 'name');
+  }, [allSecurityRules]);
+
+  const hasNoSecuritySelected = !initialAppliedSecurityRules.length;
+
+  console.log(formikProps.values);
```
al-rosenthal: nit: remove pls

api/src/repositories/security-repository.ts
```diff
@@ -379,43 +384,89 @@ export class SecurityRepository extends BaseRepository {
...
+
+    const response = await this.connection.knex(queryBuilder, SubmissionFeatureSecurityRecord);
+
+    return response.rows;
+  }
+
+  /**
+   * Removes the given security rules for a given set of given submission features.
+   *
+   * @param {number[]} submissionFeatureIds
```
al-rosenthal, line 409: nit: missing the security rule param

api/src/services/security-service.ts
```diff
@@ -330,53 +330,95 @@ export class SecurityService extends DBService {
...
+   * Patches security rules that are applied or removed to the given set of submission features. If a
+   * particular rule happens to belong to both `applyRuleIds` and `removeRuleIds`, it will always be
+   * added.
    *
-   * @param {number[]} features
-   * @param {number[]} rules
+   * @param {number[]} submissionFeatureIds IDs of the submission features whose security will be updated.
+   * @param {number[]} applyRuleIds IDs of the rules which will be applied after the patch operation
+   * @param {number[]} removeRuleIds IDs of the rules which will be removed after the patch operation
    * @param {boolean}
```
al-rosenthal: nit

api/src/services/security-service.ts
```diff
@@ -330,53 +330,95 @@ export class SecurityService extends DBService {
   }
 
   /**
-   * Applies all given security rules to the given set of submission feature ids.
-   * This process is additive unless the override flag is set to `true`.
-   * If the override is set to `true` all security rules for the given set of features will be removed.
-   * Then the new security rules will be applied as normal.
+   * Patches security rules that are applied or removed to the given set of submission features. If a
```
al-rosenthal, line 333: 👍

app/src/constants/i18n.ts
```diff
@@ -26,15 +26,16 @@ export const DeleteSystemUserI18N = {
 };
 
 export const ApplySecurityRulesI18N = {
-  applySecuritySuccess: (ruleCount: number, featureCount: number) => {
-    if (ruleCount === 0) {
-      return `Successfully removed all security rules for ${featureCount} ${p(featureCount, 'feature')}.`;
-    }
+  // TODO
```
al-rosenthal: nit

al-rosenthal (commented):
Ran this locally and it works great. Couple of nit picks but overall solid

al-rosenthal (approved):
😎

api/src/services/security-service.ts
```diff
@@ -330,53 +330,94 @@ export class SecurityService extends DBService {
...
    * @return {*}  {Promise<SubmissionFeatureSecurityRecord[]>}
    * @memberof SecurityService
    */
-  async removeSecurityRulesFromSubmissionFeatures(features: number[]): Promise<SubmissionFeatureSecurityRecord[]> {
-    return this.securityRepository.removeSecurityRulesFromSubmissionFeatures(features);
+  async removeSecurityRulesFromSubmissionFeatures(
+    submissionFeatureIds: number[],
+    removeRuleIds?: number[]
+  ): Promise<SubmissionFeatureSecurityRecord[]> {
+    if (!submissionFeatureIds.length) {
```
KjartanE, line 378: Is this check required when `(removeRuleIds.length > 0)` is done before on line 354 to even call this function?

KjartanE, line 378: Is this check required when a similar check is already called in the parent function? redundancy is good just wondering if this function will be called from anywhere else.

api/src/services/security-service.ts
```diff
@@ -330,53 +330,94 @@ export class SecurityService extends DBService {
...
-      // we want to override any security rules present and can achieve this by remove everything first
-      await this.securityRepository.removeSecurityRulesFromSubmissionFeatures(features);
+  async patchSecurityRulesOnSubmissionFeatures(
+    submissionFeatureIds: number[],
+    applyRuleIds: number[],
+    removeRuleIds: number[]
+  ): Promise<void> {
+    defaultLog.debug({ label: 'patchSecurityRulesOnSubmissionFeatures', applyRuleIds, removeRuleIds });
+
+    if (!submissionFeatureIds.length) {
```
KjartanE, line 350: Is this check required when our api doc declares that min 1 item is required by the PATCH request.

api/src/services/security-service.ts
```diff
@@ -330,53 +330,94 @@ export class SecurityService extends DBService {
...
    *
-   * @param {number[]} features
+   * @param {number[]} submissionFeatureIds
+   * @return {*}  {Promise<SecurityRuleRecord[]>}
+   * @memberof SecurityService
+   */
+  async getSecurityRulesForSubmissionFeatures(
+    submissionFeatureIds: number[]
+  ): Promise<SubmissionFeatureSecurityRecord[]> {
+    if (!submissionFeatureIds.length) {
```
KjartanE, line 399: Same questions as above. Is this check required when the GET request API doc forces a minimum 1 id.

KjartanE: Is this function even used in the current state?

KjartanE (approved):
Tested locally, works great. Just couple questions about service functions. 🐈

api/src/services/security-service.ts
curtisupshall: To add to my previous comment, I think of the API schema as moreso a way to catch malformed queries made by the user, whereas this check is added to simply forgo making an empty request on the database.

api/src/services/security-service.ts
curtisupshall: (Hoping this reply sufficiently addresses the other comments) I think the ethos behind service-level classes suggests that they may be called from both various endpoints and auxiliary services. For that reason, I added this check to all the new service-level methods, and removed it from the underlying repository methods.

To that end, I think the checks are worth it because they add virtually no overhead, and because they help to potentially avoid an unhelpful and costly operation (making a round trip to the database and returning an empty response).

api/src/services/security-service.ts
curtisupshall: This function is not currently being used, because a new function is providing similar enough functionality. But because this function is adjacent to a problem that is still evolving in scope (applying and removing security) I think it makes sense to keep it around (at least for now).

PR #230, opened 2024-01-12
SIMSBIOHUB-424: Support Artifact Intake + Misc Enhancements

NickPhura:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-424

## Description of Changes

Added a call to clamav to app.ts, alongside the multer middleware, so that we shouldn't have to call it in the endpoint anymore. Endpoint code should just be able to assume any files getting that far have already been checked and are safe.

Update submission/intake endpoint/service/repo
- Updates to better traverse the submission json structure, inserting each submission feature.
  - Now also correctly assigns `parent_submission_feature_id` when inserting submission features

Update submission search indexing
- Previously, checked all feature properties against all incoming features.
- Now, only checks feature properties registered to a given feature type, against incoming features of the same type.
- Some room for optimizing the code here a bit, as right now it calls `.find` repeatedly, when it probably should just pre-parse all of the codes ahead of time, and then do direct object de-referencing. It was doing this before, but the codes response object got more complicated and I just skipped optimizing it for now in order to get this PR out quicker.

Update codes endpoint/service/repo
- Updated api types, and modified endpoint response
- Updated app types accordingly

Update artifact/intake endpoint/service/repo

Add `required_value` column to feature_type_property table to indicate if properties are required or optional when running validation.

Added `additionalProperties: false` to most endpoints, and fixed response validation issues.

Update get submission features endpoint to replace the original `data` object with one populated from `search_*` rows.
- Added related service/repo code

Update submission patch function to also unpublish all other submissions for the same `uuid` on `publish=true`

## Testing Notes

If running both SIMS and BioHub branches, should be able to publish a survey that has 0 or more attachments. Attachments should have their corresponding submission feature records created, and matching search_string s3_key records added, and the file should be uploaded the S3.

Related PR for the SIMS changes: https://github.com/bcgov/biohubbc/pull/1193


NickPhura:
Working on unit tests:

artifact-service.ts (Nick - done)
~artifact-repository.ts~

code-repository.ts (Curtis - done)

submission-service.ts (Nick - done)
submission-repository.ts (Nick - done)

search-index-repository.ts (Curtis - done)

artifact/intake.ts  (Nick - done)

api/src/repositories/submission-repository.ts
```diff
@@ -1454,6 +1465,88 @@ export class SubmissionRepository extends BaseRepository {
...
+    if (criteria?.featureTypeNames?.length) {
+      const featureTypeNames = criteria?.featureTypeNames;
+      // Filter by feature type names
+      queryBuilder.whereIn('feature_type_id', (qb) => {
+        qb.select('feature_type_id')
+          .from('feature_type')
+          .where(
+            knex.raw('LOWER(name)'),
+            'IN',
+            featureTypeNames.map((featureTypeName) => featureTypeName.toLowerCase())
```
curtisupshall, line 1639: Is this syntactically correct? I thought you had to wrap the `WHERE IN` values in brackets, e.g. `WHERE LOWER(name) IN (?,?,?)`

api/src/services/artifact-service.ts
```diff
@@ -49,59 +52,47 @@ export class ArtifactService extends DBService {
...
-    return this.insertArtifactRecord({
-      ...metadata,
-      artifact_id,
-      submission_id,
-      key: s3Key,
-      uuid: fileUuid
-    });
+    defaultLog.debug({ label: 'uploadSubmissionFeatureArtifact', message: 'S3 key', artifactS3Key });
+
+    // TODO add api codes cache: so lookups like this are fast (especially since codes dont change often)
```
curtisupshall, line 76: Curious to hear your thoughts on how we might want to achieve query caching 🙂

api/src/utils/file-utils.ts
```diff
@@ -272,15 +272,15 @@ export async function getObjectMeta(key: string): Promise<HeadObjectOutput> {
...
  */
-export function generateArtifactS3FileKey(options: IArtifactS3FileKey) {
+export function generateSubmissionFeatureS3FileKey(options: IArtifactS3FileKey) {
   const keyParts: (string | number)[] = [];
 
   keyParts.push(_getS3KeyPrefix());
-  keyParts.push('datasets');
-  keyParts.push(options.datasetUUID);
-  keyParts.push('artifacts');
+  keyParts.push('submissions');
```
curtisupshall: Can we simplify this function a bit:

```javascript
return [
  'datasets',
  options.datasetUUID,
  // ...
].filter(Boolean).join('/');
```

al-rosenthal (commented):
2 things I noticed:
- an import in `useSubmissionDataGridColumns` is upset about a missing interface, not sure if it was intentionally removed
- s3_key needs to be added to `validateProperties` in the `validation-service`

On that topic, does it make more sense to name this key vs s3_key? I'm also wondering if the key needs it's own property type since it's just another string.

NickPhura:
> 2 things I noticed:
> 
>     * an import in `useSubmissionDataGridColumns` is upset about a missing interface, not sure if it was intentionally removed
> 
>     * s3_key needs to be added to `validateProperties` in the `validation-service`
> 
> 
> On that topic, does it make more sense to name this key vs s3_key? I'm also wondering if the key needs it's own property type since it's just another string.

Yeah a few issues have cropped up after Mac and I's code got combined. The primary resolution is to leverage that 'calculated' column in the properties table. The validation can use it to skip validating calculated fields (currently it treats all properties as required, which breaks when it gets to s3_keys as they aren't generated until after validation has run, and also won't be saved to the properties column at all but will be added to the search_strings table directly), and then the frontend can still use it to understand how to render 's3_key' fields which will be part of the search_* table results that the frontend uses to render the data tables.

Re: 'key' vs 's3_key' - Yeah we can change it. I think I originally just landed on s3_key for no other reason than it would be easy to change later, and all of the existing artifact code uses 's3' in its naming. But, we can certainly change it now to whatever we think is best. 'artifact_key' might even be better? Just so it describes itself a bit but is still generic?

api/src/utils/file-utils.ts
NickPhura: For sure, the main reason for it being separate pushes, historically, was that we had some if statements in there that would change the key depending on which fields were provided. Something we borrowed from SIMS. Though in BioHub it looks like we never had any optional fields, so there were no conditional key parts.

api/src/paths/administrative/security/submission/{submissionId}.ts
```diff
@@ -51,18 +51,59 @@ GET.apiDoc = {
...
+                create_user: {
+                  type: 'integer',
+                  minimum: 1
+                },
+                update_date: {
+                  type: 'string',
+                  nullable: true
+                },
+                update_user: {
+                  type: 'string',
```
curtisupshall: Wouldn't `update_user` have the same type as `create_user`?

api/src/repositories/submission-repository.ts
curtisupshall: On second thought, I'm guessing this is handled by Knex 🤔 Feel free to resolve this comment

api/src/paths/administrative/security/submission/{submissionId}.ts
NickPhura: Yup, will fix

api/src/repositories/submission-repository.ts
NickPhura: Sorry, yeah I believe knex is cool with it. It did take a few trys to get it right though. I tried using `whereIn` which you would think would work exactly like the current implementation, just minus the `IN` param. But the type checking kept throwing errors. Changing it to a regular `Where` + the `IN` param made it happy.

curtisupshall (approved):
Code looks good to me!

al-rosenthal (approved):
ran this locally and things are working as expected 🌯

api/src/services/artifact-service.ts
NickPhura: Yeah there are a few things we can do, depending on what is most helpful and/or the development cost:

Backend caching of just static data (like codes) by just fetching them on server start, so the endpoint/service can respond immediately. Could do the same on the frontend for static data as well (which we might already be achieving with the codes context?)

Frontend caching of all/some requests (using something like react-query and/or axios-cache-interceptor).

Backend caching of all/some requests (leveraging eTags and 304 statuses) (Definitely the most intense option).

PR #234, opened 2024-01-22
SIMSBIOHUB-455 - ITIS Taxon cache + Taxon Alias tables

MacDeluca:
## Links to Jira Tickets
[Ticket-455](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-455?)

## Description of Changes

- New Taxon ITIS cache table
- New Taxon Alias table
- New Language Lookup table
- New Taxon Alias Origin table

## Questions

1. Should we have soft delete unique indexes on scientific_name / bc_taxon_code and itis_tsn or just scientific_name?


database/src/migrations/20240119000000_taxonomy_tables.ts
```diff
@@ -0,0 +1,102 @@
...
+    -- Create tables
+    ----------------------------------------------------------------------------------------
+    set search_path=biohub,public;
+
+    CREATE TABLE taxon (
+      taxon_id                integer             GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+      itis_tsn                integer             NOT NULL,
+      bc_taxon_code           varchar(10),
+      scientific_name         varchar(50)         NOT NULL,
+      common_name             varchar(50),
```
NickPhura: Are we confident in these varchar lengths? I would probably go a bit higher just to be safe.

(The actual length specified has no impact on performance/storage, because it only actually stores the length of the string. Its just a constraint on the length.)

database/src/migrations/20240119000000_taxonomy_tables.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    COMMENT ON COLUMN taxon_alias.taxon_alias_id  IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN taxon_alias.taxon_id        IS 'A foreign key that points to a taxon.';
+    COMMENT ON COLUMN taxon_alias.alias           IS 'A taxon alias.';
+    COMMENT ON COLUMN taxon_alias.create_date     IS 'The datetime the record was created.';
+    COMMENT ON COLUMN taxon_alias.create_user     IS 'The id of the user who created the record as identified in the system user table.';
+    COMMENT ON COLUMN taxon_alias.update_date     IS 'The datetime the record was updated.';
+    COMMENT ON COLUMN taxon_alias.update_user     IS 'The id of the user who updated the record as identified in the system user table.';
+    COMMENT ON COLUMN taxon_alias.revision_count  IS 'Revision count used for concurrency control.';
+    COMMENT ON TABLE  taxon_alias                 IS 'Taxon alias table, for assigning additional alias names to taxons';
```
NickPhura: super nit: period at the end 😎

database/src/migrations/20240119000000_taxonomy_tables.ts
```diff
@@ -0,0 +1,102 @@
...
+      update_user             integer,
+      revision_count          integer             DEFAULT 0 NOT NULL,
+
+      CONSTRAINT              taxon_pk            PRIMARY KEY (taxon_id)
+    );
+
+    COMMENT ON COLUMN taxon.taxon_id              IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN taxon.itis_tsn              IS 'ITIS primary key identifier, populated from ITIS response. ITIS (Integrated Taxonomic Information System), TSN (Taxonomic Serial Number). https://itis.gov/pdf/faq_itis_tsn.pdf';
+    COMMENT ON COLUMN taxon.bc_taxon_code         IS 'British Columbia standard taxon identifier.';
+    COMMENT ON COLUMN taxon.scientific_name       IS 'Taxon scientific name, initially populated from ITIS response.';
```
NickPhura: Not 100% certain, but I feel like the `scientific_name` should be `itis_scientific_name` (just implying that it is always the itis value, and not just initially populated from it and then potentially overwritten by BC later). Possibly the same for the `common_name` column? 

Just thinking that we probably want the itis values to be the source of truth at all times, and the aliases/bc values are just a way of finding the correct record with the source-of-truth itis values.

Thoughts in general?

database/src/migrations/20240119000000_taxonomy_tables.ts
```diff
@@ -0,0 +1,102 @@
...
+    COMMENT ON COLUMN taxon.record_effective_date IS 'Record level effective date.';
+    COMMENT ON COLUMN taxon.record_end_date       IS 'Record level end date.';
+    COMMENT ON COLUMN taxon.create_date           IS 'The datetime the record was created.';
+    COMMENT ON COLUMN taxon.create_user           IS 'The id of the user who created the record as identified in the system user table.';
+    COMMENT ON COLUMN taxon.update_date           IS 'The datetime the record was updated.';
+    COMMENT ON COLUMN taxon.update_user           IS 'The id of the user who updated the record as identified in the system user table.';
+    COMMENT ON COLUMN taxon.revision_count        IS 'Revision count used for concurrency control.';
+    COMMENT ON TABLE  taxon                       IS 'Taxon cache table, extending ITIS webservice response.';
+
+    CREATE TABLE taxon_alias (
```
NickPhura, line 57: We may as well add the effective/end timestamp columns here too while we're at it

database/src/migrations/20240119000000_taxonomy_tables.ts
MacDeluca: I can bump them up. 100?

NickPhura: Yeah 100 seems more reasonable. 

Wikipedia apparently has a page for long taxonomic names, the longest beign 73 characters. No idea how accurate it is, but 100 seems like it would cover all cases then.

https://en.wikipedia.org/wiki/List_of_long_species_names

NickPhura: Actually, if we search ITIS by name for 'llanfair­pwll­gwyn­gyll­go­gery­chwyrn­drobwll­llan­tysilio­gogo­gochensis' I wonder if it comes up and how long the full scientific name is

database/src/migrations/20240119000000_taxonomy_tables.ts
MacDeluca: .

database/src/migrations/20240119000000_taxonomy_tables.ts
curtisupshall: Per discussion, ITIS uses `varchar(300)` for their `completename` column.

database/src/migrations/20240119000000_taxonomy_tables.ts
```diff
@@ -0,0 +1,166 @@
...
+      update_date             timestamptz(6),
+      update_user             integer,
+      revision_count          integer                   DEFAULT 0 NOT NULL,
+
+      CONSTRAINT              taxon_alias_pk            PRIMARY KEY (taxon_alias_id)
+    );
+
+    COMMENT ON COLUMN taxon_alias.taxon_alias_id        IS 'System generated surrogate primary key identifier.';
+    COMMENT ON COLUMN taxon_alias.taxon_id              IS 'A foreign key that points to a taxon.';
+    COMMENT ON COLUMN taxon_alias.alias                 IS 'A taxon alias.';
```
NickPhura: Just missing comments for the new language/origin fields. Otherwise looks great.

al-rosenthal (dismissed):
🎸

curtisupshall (dismissed):
LGTM

NickPhura (approved):

curtisupshall (approved):

al-rosenthal (approved):
🧈

PR #236, opened 2024-02-09
SIMSBIOHUB-459: Fixed save taxonomy repo method

curtisupshall:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-459

## Description of Changes

Fixed a race condition that occurs when two requests try to read/write from the `taxon` table.

### DIFF
--- a/api/src/repositories/taxonomy-repository.ts
+++ b/api/src/repositories/taxonomy-repository.ts
@@ -3,6 +3,7 @@ import { z } from 'zod';
 import { getKnex } from '../database/db';
 import { getLogger } from '../utils/logger';
 import { BaseRepository } from './base-repository';
+import { ApiExecuteSQLError } from '../errors/api-error';
 
 const defaultLog = getLogger('repositories/taxonomy-repository');
 
@@ -64,7 +65,7 @@ export class TaxonomyRepository extends BaseRepository {
     commonName: string | null,
     itisData: Record<string, unknown>,
     itisUpdateDate: string
-  ): Promise<void> {
+  ): Promise<TaxonRecord> {
     defaultLog.debug({ label: 'addItisTaxonRecord', itisTsn });
 
     const sqlStatement = SQL`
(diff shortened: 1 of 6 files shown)


api/src/repositories/taxonomy-repository.ts
```diff
@@ -60,40 +62,43 @@ export class TaxonomyRepository extends BaseRepository {
...
+        DO NOTHING
+        RETURNING *
       )
-      RETURNING
-        *;
+      SELECT * FROM inserted_row
+      UNION
+      SELECT * FROM taxon
+      WHERE 
+        taxon.itis_tsn = 93
```
NickPhura: 93 -> ${itisTsn}

curtisupshall: whoops!

curtisupshall: Fixed.

api/src/services/taxonomy-service.ts
```diff
@@ -29,23 +32,25 @@ export class TaxonomyService {
...
       const itisService = new ItisService();
       const itisResponse = await itisService.searchItisByTSN(missingTsnIds);
 
-      patchedTaxonRecords = await Promise.all(itisResponse.map(async (item) => this.addItisTaxonRecord(item)));
+      await Promise.all(itisResponse.map(this.addItisTaxonRecord));
     }
 
     // Missing ids patched, return taxon records for all requested ids
-    return this._sanitizeTaxonRecordsData(existingTaxonRecords.concat(patchedTaxonRecords));
+    return this._sanitizeTaxonRecordsData(await this.taxonRepository.getTaxonByTsnIds(tsnIds));
```
NickPhura: Is the addItisTaxonRecord query not working as expected? I thought the hope was that we could get the new record back after inserting (or conflicting and not inserting), so we wouldn't have to make another call afterwards (and can just combine the 2 results sets).

curtisupshall: From what I could gather, the query was not returning anything after running successfully 🤔

NickPhura: Hmm weird, It seemed to be working for me in DBeaver, but maybe I need to try running it directly in PSQL just in case

KjartanE (dismissed):
Tested locally, works as expected!

api/src/services/taxonomy-service.ts
curtisupshall: After some testing, addressing this comments appears to have worked successfully.

NickPhura (dismissed):

NickPhura (dismissed):
Also, I snuck a tiny change into your PR, just to reduce our logging output in OpenShift

NickPhura (dismissed):

NickPhura (approved):
Okay, that should be all of the changes to sneak in... :)

al-rosenthal (approved):
🛷

PR #237, opened 2024-03-05
Sort ITIS response & multiple common names

mauberti-bc:
## Links to Jira Tickets

N/A

## Description of Changes

- Sort the ITIS search response to prioritize exact matches
- Allow for multiple common names, rather than taking only the first common name from ITIS

## Testing Notes

- When searching for taxa, exact matches on either scientific name or common name should be among the first taxa in the response. Eg. Searching for "black bear" would place "American black bear" near the top of the list
- Requires SIMS changes: https://github.com/bcgov/biohubbc/pull/1230


api/src/repositories/taxonomy-repository.ts
```diff
@@ -53,7 +53,7 @@ export class TaxonomyRepository extends BaseRepository {
    *
    * @param {number} itisTsn
    * @param {string} itisScientificName
-   * @param {(string | null)} commonName
+   * @param {(string | null)} commonNames
```
curtisupshall: This JS doc change should also be reflected in the actual function

api/src/services/itis-service.ts
```diff
@@ -67,19 +76,22 @@ export class ItisService {
...
     return data.map((item: ItisSolrSearchResponse) => {
-      const commonName = item.commonNames ? item.commonNames[0].split('$')[1] : null;
+      const englishNames = item.commonNames?.filter((name) => name.split('$')[2] === 'English');
+      const commonNames = englishNames?.map((name) => name.split('$')[1]);
 
       return {
         tsn: Number(item.tsn),
-        commonName: commonName,
-        scientificName: item.scientificName
+        commonNames: commonNames ?? [],
```
curtisupshall: I would move the coalesce to the same line that `commonNames` is declared on:

```javascript
const commonNames = englishNames?.map((name) => name.split('$')[1]) ?? [];
```

api/src/services/taxonomy-service.ts
```diff
@@ -7,7 +7,7 @@ const defaultLog = getLogger('services/taxonomy-service');
 
 export type TaxonSearchResult = {
   tsn: number;
-  commonName: string | null;
+  commonNames: string[] | [];
```
curtisupshall: You should remove the `[]` type. This should just just read `commonNames: string[];`.

api/src/repositories/taxonomy-repository.ts
curtisupshall: I understand the database doesn't support multiple common names. I think a good compromise for now would be to pass an array of common names, then simply store the first item in the array, and assume that anyone calling this function should only expect the first array item to be stored.

api/src/services/taxonomy-service.ts
```diff
@@ -73,9 +72,10 @@ export class TaxonomyService {
    * @memberof TaxonomyService
    */
   async addItisTaxonRecord(itisSolrResponse: ItisSolrSearchResponse): Promise<TaxonRecord> {
-    let commonName = null;
+    let commonNames = null;
     if (itisSolrResponse.commonNames) {
-      commonName = itisSolrResponse.commonNames[0].split('$')[1];
+      const firstEnglishName = itisSolrResponse.commonNames.find((name) => name.split('$')[2] === 'English');
+      commonNames = firstEnglishName ? firstEnglishName.split('$')[1] : null;
```
curtisupshall: I would expect `commonNames` to be an array. I think leaving `commonNames` as type `string | null` may lead to some confusion. (See my other comment I left in taxonomy-repository)

api/src/services/taxonomy-service.ts
```diff
@@ -56,13 +56,12 @@ export class TaxonomyService {
...
-    return taxonRecords.map((item: TaxonRecord) => {
-      return {
-        tsn: item.itis_tsn,
-        commonName: item.common_name,
-        scientificName: item.itis_scientific_name
-      };
-    });
+    return taxonRecords.map((item: TaxonRecord) => ({
+      tsn: item.itis_tsn,
+      // placeholder: wrap commonNames in array until the database supports multiple common names
```
curtisupshall: This comment is useful but I would mark it with `TODO` so it can be easily found during future tech debt efforts.

To that end, if there is an existing Jira ticket that tracks this desired change (namely, **add database support for multiple common names**), it's a good idea to mention the ticket number or provide a link to the Jira ticket.

api/src/utils/itis-sort.test.ts
```diff
@@ -0,0 +1,3 @@
+// TODO
+export const placeholder = () => {}
```
curtisupshall: If you want to add a placeholder test file, I would write:
```javascript
describe('myFunctionName', () => {
 // TODO
});
```

curtisupshall: That being said, if you plan to add the tests in a future PR, it's often not worth it to include placeholder test files like this, since it's really easy to find missing test coverage using `npm run coverage`. Although I assume you will be adding them in this PR 👍

api/src/utils/itis-sort.ts
curtisupshall, line 1: I don't think I have anything specific to mention about these functions. I like that they're well documented and I can't see anything that resembles a peformance issue. I will say that it is a little difficult to follow what these methods are doing, mostly because of nested functions that share the same name and almost the same implemention. For example, `customSort` is defined three different times, and each time it is immediately referenced. I think in this case, you could inline the function, e.g. `return data.sort((a, b) => { // customSort logic goes here ...`

api/src/utils/itis-sort.ts
```diff
@@ -0,0 +1,141 @@
...
+      return 0; // Maintain the original order
+    } else {
+      return 0; // Maintain the original order
+    }
+  };
+
+  // Function to check if an item is a match with search terms
+  const checkForMatch = (item: TaxonSearchResult, searchTerms: string[]) => {
+    const commonNameWords = item.commonNames?.map((name: string) => name.toLowerCase());
+    const scientificNameWord = item.scientificName.toLowerCase();
```
curtisupshall: I might comb over these methods and consider renaming a few variables. For example, `scientificNameWord` does not describe what the variable contains any better than calling it `scientificName` would. However, if it was instead called something like `scientificNameSanitized`, I would understand that it contains the scientific name, but normalized in some way (in this instance, converted to lowercase).

curtisupshall: On the topic of converting the scientific name to lowercase, you could make the code [more dry](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by sanitizing the scientfic names and search terms in the parent method, allowing you to drop all the instances of `.toLowerCase()`. You could also sanitize these terms in other ways, such as trimming whitespace with `.trim`.

api/src/utils/itis-sort.ts
```diff
@@ -0,0 +1,141 @@
...
+      return -1; // Place items from searchTerms before other items
+    } else if (!aInReference && bInReference) {
+      return 1; // Place other items after items from searchTerms
+    } else {
+      return 0; // Maintain the original order if both are from searchTerms or both are not
+    }
+  };
+
+  const checkForMatch = (item: TaxonSearchResult, searchTerms: string[]) => {
+    const commonNameWords = item.commonNames?.flatMap((name: string) => name.toLowerCase().split(/\s+/));
```
curtisupshall: `.split(/\s+/)` will yield an array that could contain an emptystring, (for example, `" hello world".split(/\s+/)` will yield `["", "hello", "world"]`) so you may want to filter these out (you can just add `.filter(Boolean)`). To that end, if you're filtering out empty strings, you may as well just split on `' '` rather than a Regex expression

MacDeluca (commented):
I think the Solr query can be optimized to get around some of the sorting / filtering on exact matches. 
Note: vernacular terms (common names) are "$" delimited.

Heres a simplified query for only the common names portion

`q=vernacular:$black\%20bear$*+vernacular:$*black\%20bear$*`

1. This will search for a common name exact match first ie: "$Black bear$"
2. Then searches for any common name ending with "black bear" ie: "$American black bear$"


Still working on the scientific name portion of this query. But this is what I have so far.
You can create a postman variable for the {{SPECIES}} refs.

https://services.itis.gov/?wt=json&fl=tsn+scientificName:nameWOInd+kingdom+parentTSN+commonNames:vernacular+updateDate+usage+rank&q=((vernacular:${{SPECIES}}$*+vernacular:$*{{SPECIES}}$*)+AND+usage:/(valid|accepted)/)+((nameWOInd:{{SPECIES}}+nameWOInd:*{{SPECIES}})+AND+usage:/(valid|accepted)/)

I'll update this comment with the revised query tomorrow.

mauberti-bc:
> I think the Solr query can be optimized to get around some of the sorting / filtering on exact matches. Note: vernacular terms (common names) are "$" delimited.

@MacQSL True. I tried adjusting the query to remove the need for sorting but couldn't consistently get the same results as the sorting functions here, and I figured that the sorting functions in this PR work well enough for now. The query can be optimized down the road

api/src/services/taxonomy-service.ts
mauberti-bc: Thanks. commonNames is now always an array.

api/src/utils/itis-sort.ts
mauberti-bc: I removed the intermediate functions 👍

NickPhura, line 1: This file and its functions are really the ideal candidates for some unit tests: they have discrete inputs/outputs, and are complicated enough that tests become valuable in ensuring the intended output stays the same between updates/maintenance.

They should also be fairly straightforward to write as well, since there are no external calls being made or any need to mock/stub anything.

NickPhura: +1 for the jsdoc and comments though, thats great

mauberti-bc: Tests will be done before this goes in 👍

NickPhura (dismissed):

NickPhura (approved):

PR #246, opened 2024-06-25
BugFix: Common Name Parsing

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

Rename `itis-sort.ts` to `itis-utils.ts`
Add shared util function for parsing common names: correctly handles case where commonNames field is missing from itis response. Updated itis response interface to mark commonNames as optional.

## Testing Notes

BioHub taxon endpoints return taxon records as expected, with an emphasis on the commonNames portion containing the expected values.

In SIMS, when adding a taxon to a survey that has no common names (like a higher rank taxon), when you go to edit the survey, the taxon should correctly be returned and show up in the UI. Before this fix, taxons with no common names were not returned at all from one of the biohub taxon endpoints.

### DIFF
--- a/api/src/utils/itis-utils.ts
+++ b/api/src/utils/itis-utils.ts
@@ -87,3 +87,24 @@ export const sortTaxonSearchResults = (
   // Sort the data by the score
   return taxonSearchResults.sort((a, b) => calculateScore(b) - calculateScore(a));
 };
+
+/**
+ * Parse the raw common names string from an ITIS taxon record into an array of english common names.
+ *
+ * @example
+ * const commonNames = [
+ *   '$withered wooly milk-vetch$English$N$152846$2012-12-21 00:00:00$',
+ *   '$woolly locoweed$English$N$124501$2011-06-29 00:00:00$',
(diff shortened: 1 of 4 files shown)


api/src/utils/itis-utils.ts
```diff
@@ -87,3 +87,24 @@ export const sortTaxonSearchResults = (
...
+ *
+ * @example
+ * const commonNames = [
+ *   '$withered wooly milk-vetch$English$N$152846$2012-12-21 00:00:00$',
+ *   '$woolly locoweed$English$N$124501$2011-06-29 00:00:00$',
+ *   '$Davis Mountains locoweed$English$N$124502$2011-06-29 00:00:00$',
+ *   '$woolly milkvetch$English$N$72035$2012-12-21 00:00:00$'
+ * ]
+ *
+ * const result = _parseItisTaxonCommonNames(commonNames)
```
MacDeluca: Doc out of sync?

NickPhura: yup! Fixed

MacDeluca (approved):
🐅

PR #250, opened 2024-08-10
Endpoint for retrieving hierarchy information for multiple TSNs

mauberti-bc:
## Links to Jira Tickets

- Supports [SIMSBIOHUB-557](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-557)

## Description of Changes

- Adds endpoint for retrieving taxonomy hierarchy for multiple TSNs
- Supports https://github.com/bcgov/biohubbc/pull/1338


api/src/paths/taxonomy/taxon/tsn/hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const GET: Operation = [getTaxonHierarchyByTSN()];
+
+GET.apiDoc = {
+  description: 'Get taxon records by TSN ids.',
+  tags: ['taxon_id'],
+  security: [
+    {
+      Bearer: []
+    }
```
NickPhura: The other PR removes the security requirements for the previous 2 taxon endpoints. Should this also have no security?

api/src/paths/taxonomy/taxon/tsn/hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+            type: 'array',
+            items: {
+              title: 'Species',
+              type: 'object',
+              required: ['tsn', 'hierarchy'],
+              properties: {
+                tsn: {
+                  type: 'integer'
+                },
+                hierarchy: { type: 'array', items: { type: 'integer' } }
```
NickPhura: Small detail, but are the TSNs in the array in some order? kingdom -> target_tsn? Or something like that?

A description explaining that might be valuable

api/src/paths/taxonomy/taxon/tsn/hierarchy.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @returns {RequestHandler}
+ */
+export function getTaxonHierarchyByTSN(): RequestHandler {
+  return async (req, res) => {
+    defaultLog.debug({ label: 'getTaxonHierarchyByTSN', message: 'query params', query: req.query });
+
+    const connection = getAPIUserDBConnection();
+
+    const tsnIds: number[] = (req.query.tsn as (string | number)[]).map(Number);
```
NickPhura: super nit: 
Regarding `as (string | number)[]`:
The query params will always be a string array at this stage. All http params that aren't in the body get converted to strings.

api/src/services/itis-service.ts
```diff
@@ -75,10 +85,32 @@ export class ItisService {
...
+   * @param {number[]} tsnIds
+   * @return {*}  {Promise<TSNWithHierarchy[]>}
+   * @memberof ItisService
+   */
+  async getHierarchyForTSNs(tsnIds: number[]): Promise<TSNWithHierarchy[]> {
+    const url = await this.getItisSolrTsnHierarchyUrl(tsnIds);
+
+    defaultLog.debug({ label: 'getHierarchyForTSNs', message: 'url', url });
+
+    const response = await axios.get(url);
```
NickPhura: (This is something that could be added to all of the itis axios calls) 

It might be nice to define the response type, so that when we parse `response.data.response.docs` it doesn't just think the whole thing is `any`

```
const response = await axios.get<ItisSolrResponse>(url)
```

NickPhura: Assuming the base solr response is generic, could even do something with generics (to pass in the type of the `docs` field)

api/src/paths/taxonomy/taxon/tsn/hierarchy.ts
mauberti-bc: Yes, thanks for catching that

api/src/paths/taxonomy/taxon/tsn/hierarchy.ts
mauberti-bc: Fixed

api/src/services/itis-service.test.ts
```diff
@@ -182,6 +182,65 @@ describe('ItisService', () => {
...
+
+      try {
+        await itisService.getHierarchyForTSNs(mockTsns);
+
+        expect.fail();
+      } catch (error) {
+        expect((error as ApiGeneralError).message).to.equal('a test error');
+        expect(getItisSolrTsnHierarchyUrlStub).to.have.been.calledWith(mockTsns);
+      }
+    });
```
NickPhura, line 241: This test isn't really accurate. The function doesn't catch any errors, it would just let them bubble up (which is totally fine). I would just remove this test.
```suggestion
```

NickPhura: Cool feature if you haven't seen it yet: can suggest edits in the PR and they become commitable changes (only really good for small/obvious changes though).

NickPhura (approved):

PR #255, opened 2025-03-20
TechDebt: Version Updates, DevOps Updates

NickPhura:
## Links to Jira Tickets

n/a

## Description of Changes

In no particular order:

1. Update some package versions.
   - Update the AWS library and related utils functions (from SIMS)
   - Update winston logger and add file logging (from SIMS)
3. Update to use Deployments over deprecated DeploymentConfigs
   - Add file logging + volume for the api/queue (from SIMS)
5. Update to fix deprecated mui imports, and related.
   - Pulled some newer code from SIMS to make this easier (file upload components)
6. Updates to the git actions
7. Update the Postgres version, and related dockerfiles, etc.
   - Wrap all instances of `system_user` in double quotes to conform to postgres 16+ reserved word rules

## Testing Notes

Deploys and functions as it did before.

## Merge/Release Notes

This postgres version upgrade will require that the database be brought down in dev, test, prod (before its merged/released to each environment). Normally, we would need to backup and restore the prod data. But since there is no production data, we can just wipe the prod database like we would with dev and test. This PR can then just be merged/released as normal.


MacDeluca:
question: Is there any concern with losing the production taxonomy? Has any of to cached prod taxonomy been updated by the team?

NickPhura:
> question: Is there any concern with losing the production taxonomy? Has any of to cached prod taxonomy been updated by the team?

Good question! I wasn't thinking about the taxonomy cache. I don't believe any custom taxonomy data exists in prod (or any of the environments). I checked prod, and both of the tables involved in storing additional aliases, alongside the main taxon cache table, are empty.

MacDeluca (approved):
👍🏼

PR #270, opened 2025-09-18
SIMSBIOHUB-782: Let Contributor Systems Publish Codesets

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-782

## Description of Changes

- Adds ```contributor``` table for tracking contributing data systems, using Keycloak's ```client_id``` as the identifier for a system
- Adds ```contributor_system_user``` table for joining service accounts to the contributor system, letting those service accounts submit data on behalf of the system
- Adds ```POST api/contributor``` endpoint for service accounts to register new contributor systems
  - eg. If "System A" wants to publish data, System A's publisher service account will register by sending a POST request to ```api/contributor```
  - The API gets all required information from the bearer JWT (no request body needed)
  - The service account is automatically associated with the contributor system

## Testing Notes
- In Postman, you can authenticate as the SIMS service account using SIMS keycloak credentials (see keycloak-service.ts to know which values)
- After authenticating, the service account must send a POST request to ```api/contributor``` with an empty request body; the API gets all required information from the bearer JWT
- The endpoint will return a 201 status code on successful registration


api/src/openapi/schemas/codeset.ts
```diff
@@ -0,0 +1,51 @@
...
+              additionalProperties: false,
+              properties: {
+                label: {
+                  type: 'string',
+                  description: 'The displayed label of the code value in the contributing system (eg. "very hot").'
+                },
+                value: {
+                  type: 'number',
+                  description: 'The value for the code referenced in the data (eg. 1, which might encode "very hot").'
+                },
```
dylanrogowsky-oxd: Should this be a string, in the case of a contributor system passing GUIDs rather than IDs?

dylanrogowsky-oxd (approved):

PR #264, opened 2025-09-12
Upgrade dependencies & modernize frontend configs

mauberti-bc:
## Links to Jira Tickets

- [SIMSBIOHUB-769](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-769)

## Description of Changes

- Replace create-react-app with Vite
- Remove the configContext (replaced with config/config.ts, which exports variables directly)
- Switch frontend tests to Vitest
- Upgrade packages to latest versions
- Remove leaflet (plan to replace with Maplibre, but leaflet wasn't used anyway)
- Update eslint configuration
- Various fixes to resolve type errors after updating packages

## Testing Notes

- There's an unrelated error in the console caused by dataLoader.load() calls not wrapped in a useEffect: ```Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.```
- The app should behave the same as dev; no functionality changes

PR #275, opened 2025-09-28
Upgrade dependencies & modernize frontend configs

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-769](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-769)

## Description of Changes

- Replace create-react-app with Vite
- Remove the configContext (replaced with config/config.ts, which exports variables directly)
- Switch frontend tests to Vitest
- Upgrade packages to latest versions
- Remove leaflet (plan to replace with Maplibre, but leaflet wasn't used anyway)
- Update eslint configuration
- Various fixes to resolve type errors after updating packages

## Testing Notes

- There's an unrelated error in the console caused by dataLoader.load() calls not wrapped in a useEffect: ```Can't perform a React state update on a component that hasn't mounted yet. This indicates that you have a side-effect in your render function that asynchronously later calls tries to update the component. Move this work to useEffect instead.```
- The app should behave the same as dev; no functionality changes


dylanrogowsky-oxd:
[PR 264](https://github.com/bcgov/biohubbc-platform/pull/264) was reverted and this is a new PR that replaces it

PR #288, opened 2025-11-28
SIMSBIOHUB-813 - Managing Data Access Policies in BioHub

dwhogan:
## Links to Jira Tickets
* https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-813

## Description of Changes
As a BioHub System Administrator, I want to manage access policies to control users' access to information.

## View Access Policies
* Display all access policies in a table
* Paginate the policies in the table
* Server-side search with a single textfield matching policy name

## Create Access Policy
WHEN the button labelled "Create Policy" is clicked
THEN open a dialog with the following fields:
* Policy name
* Policy description
* Policy statements (json input like monaco-editor, structured like AWS policies)

## Edit Access Policy
WHEN clicking "Edit" on an access policy in the table
THEN open a dialog with the following fields pre-populated with the Policy's data
* Policy name
* Policy description
* Policy statements

## Acceptance Criteria
Create a new frontend route for policies at /admin/policies
The route should only be accessible to System Admins (protected using the <AuthGuard> component for SYSTEM_ADMIN, DATA_ADMINISTRATOR)

## Testing Notes
 * When **Logged Out** of the app, use the url http://localhost:7100/admin/policies to test routing to the logon page


app/src/contexts/urnEditorContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+      // Skip if already cached
+      if (submissionFeaturesCache.has(submissionId)) {
+        return;
+      }
+
+      try {
+        const featureGroups = await biohubApi.submissions.getSubmissionFeatureGroups(submissionId);
+        setSubmissionFeaturesCache((prev) => new Map(prev).set(submissionId, featureGroups));
+      } catch (error) {
+        console.error(`Failed to fetch features for submission ${submissionId}:`, error);
```
mauberti-bc: For error handling, we have the ```dialogContext``` which lets you add a snackbar or error dialog and avoid console logs. Snackbar probably make more sense here. You can find examples be searching ```dialogContext.setSnackbar```

app/src/features/admin/policies/ManagePoliciesPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+const ManagePoliciesPage: React.FC<React.PropsWithChildren> = () => {
+  const biohubApi = useApi();
+  const [searchTerm, setSearchTerm] = useState('');
+  const [debouncedSearchTerm, setDebouncedSearchTerm] = useState('');
+
+  const policiesDataLoader = useDataLoader((search?: string) =>
+    biohubApi.policies.getPolicies({ search: search || undefined })
+  );
+  policiesDataLoader.load(debouncedSearchTerm);
```
mauberti-bc, line 27: React complains if you call ```dataLoader.load()``` outside of a useEffect. You should see an error in the console originating from this line.

app/src/contexts/urnEditorContext.tsx
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Cache of submission features, keyed by submission ID.
+   */
+  submissionFeaturesCache: Map<number, IGetSubmissionGroupedFeatureResponse[]>;
+  /**
+   * Fetch features for a specific submission (with caching).
+   *
+   * @param submissionId - The submission ID to fetch features for
+   */
+  fetchFeaturesForSubmission: (submissionId: number) => Promise<void>;
```
mauberti-bc: The caching and autocomplete is awesome! I'd suggest making it more obvious that the data loaders inside the ```urnEditorContext``` should only ever be used for autocomplete - like rename to fetchFeaturesForAutocomplete and add extra comments.

We try to keep contexts very slim and focused on the feature they support, and it would be easy to start using this context's ```submissionsDataLoader``` for something beyond the original autocomplete functionality (which we want to avoid at this point).

app/src/features/admin/policies/utils/policyTransform.ts
```diff
@@ -0,0 +1,194 @@
...
+ * Transform full policy object to JSON document format for editing.
+ */
+export const transformPolicyToJson = (policy: IPolicy): string => {
+  return transformApiToPolicyJson(policy.statements);
+};
+
+const URN_PATTERN = /^urn:(\*|\d+):(\*|[a-z_]+):(\*|\d+)$/;
+
+/**
+ * Validate a single condition object.
```
mauberti-bc: Add ```params``` and ```returns``` to the jsdocs, and describe any ambiguous parameters, like stmtIndex.

```
/**
* <description of method> 
* 
* @param {any} cond
* @param {number} stmtIndex - <description of parameter if ambiguous>
* @returns {string | null}
*/

File: app/src/features/admin/policies/utils/policyTransform.ts [hunk 5]
```diff
@@ -0,0 +1,194 @@
...
+  if (cond.Value === undefined) {
+    return `Statement ${stmtIndex + 1}, Condition ${condIndex + 1}: Value is required`;
+  }
+  return null;
+};
+
+/**
+ * Validate conditions array for a statement.
+ */
+const validateConditions = (conditions: any, stmtIndex: number): string | null => {
```
mauberti-bc: We try to avoid ```any``` type. Unless it is truly anything, which isn't the case because the code rejects non-arrays, prefer doing the union of possible types.

```conditions: PolicyCondition[]``` or whatever the correct type is

File: app/src/components/dialog/EditDialog.tsx [hunk 6]
```diff
@@ -105,7 +105,7 @@ export const EditDialog = <T extends FormikValues>(props: React.PropsWithChildre
           aria-labelledby="edit-dialog-title"
           aria-describedby="edit-dialog-description">
           <DialogTitle id="edit-dialog-title">{props.dialogTitle}</DialogTitle>
-          <DialogContent>{props.component.element}</DialogContent>
+          <DialogContent sx={{ overflow: 'visible' }}>{props.component.element}</DialogContent>
```
mauberti-bc: This has pretty far-reaching effects on most dialogs and pushes the save/close buttons out of view. I think we should revert this change to the DialogContent.

We can improve styling in a different PR because the EditDialog component should accept more props to control sizing and styling.

app/src/features/admin/policies/ManagePoliciesPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  const debouncedRefresh = useMemo(
+    () =>
+      debounce((term: string) => {
+        setDebouncedSearchTerm(term);
+        policiesDataLoader.refresh(term);
+      }, 300),
+    [policiesDataLoader]
```
mauberti-bc: This currently sends a request on every keystroke, instead of sending a single request after 300ms of inactivity (it should wait 300ms since the last inputChange). You can see this by watching your browser's network tab while typing into the search input.

The reason is because of the policiesDataLoader dependency. The refresh method updates the policiesDataLoader, which refreshes the useMemo, which then repeats in a loop. 

If you search for other uses of ```debounce``` in the app, you can see that we use the //exhaustive-deps comment with an empty dependency array to fix the issue.

app/src/features/admin/policies/ManagePoliciesPage.tsx
dwhogan: I tried to to follow other code practise that use a hook for data loaders;

example;
const usersDataLoader = useDataLoader(() => biohubApi.user.getUsersList());
usersDataLoader.load();

There is an error in the browser at:
getUser | @ | useUserApi.ts:29

it might be caused by the timing of the fetch, that is before the use is authenticated.

do you want me to explore?

app/src/features/admin/policies/components/AddPolicyForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+              error={touched.policy_json ? (errors.policy_json as string) : undefined}
+            />
+          </UrnEditorContextProvider>
+        </Box>
+      </Box>
+    </form>
+  );
+};
+
+export default AddPolicyForm;
```
mauberti-bc: This is a nit but we've been using named exports (export keyword inline with the function definition) instead of default export to disallow components from being used under a different name - which has created confusion in the past.

eg. I could import AddPolicyForm as DeleteAllPoliciesForm. That's an extreme and unlikely example but I find that some of the AI models will happily import using different names.

app/src/features/admin/policies/ManagePoliciesPage.tsx
mauberti-bc: I think that 401 error is unrelated to your PR and you can ignore it here. It looks like there are 2 issues with the request made here: useUserApi.ts:29
1) Like you noticed, the request doesn't wait for authentication so there's no bearer token
2) The authorization check for the ```GET /user/self``` endpoint requires the authenticated user to exist in the system user table, but we don't yet have a mechanism for adding users, so it will always fail for non-seeded users.

We can fix those 2 issues outside of this PR (ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-826).

api/src/services/access-policy/policy-service.ts
```diff
@@ -1,15 +1,50 @@
...
 import { parseFeatureUrn } from '../../database/urn-utils';
 import { CreatePolicy, Policy, UpdatePolicy } from '../../models/policy';
+import { PolicyEffect, PolicyStatement } from '../../models/policy-statement';
+import { PolicyConditionOperator, PolicyStatementCondition } from '../../models/policy-statement-condition';
 import { PolicyRepository } from '../../repositories/authorization/policy-repository';
 import { DBService } from '../db-service';
+import { PolicyStatementConditionService } from './policy-statement-condition-service';
+import { PolicyStatementService } from './policy-statement-service';
+
+/**
```
mauberti-bc: We should be more consistent in putting interfaces and types in their own .interface.ts file. I'd suggest adding src/services/access-policy/policy-service.interface.ts, or feel free to implement a different pattern that avoids cluttering business logic code with interface definitions.

I suppose for types that are used across paths/, services/, and repositories/ the each location doesn't matter, as long as we're consistent.

app/src/features/admin/policies/utils/policyTransform.ts
```diff
@@ -0,0 +1,209 @@
...
+ * Validate a policy JSON string for structural correctness.
+ *
+ * Note: This performs basic structural validation only. Detailed validation
+ * (operator/key/value compatibility) is handled by the database trigger
+ * tr_validate_policy_condition_key().
+ *
+ * @param {string} policyJson - JSON string representing the policy document to validate
+ * @returns {string | null} Error message if validation fails, null if valid
+ */
+export const validatePolicyJson = (policyJson: string): string | null => {
```
mauberti-bc: It might be worth creating a return type for the validated json, instead of just using string, to help differentiate between validated/not-yet-validated data.

mauberti-bc (dismissed):
Looks great! A couple notes for future work that shouldn't block this from being merged (we can fix them after):
* Prefer using MUI's DataGrid over Table - we already use it elsewhere and it's just easier in my experience.
* There are some pagination helpers that will be merged in another PR, then we can update the policies table to use them. That way pagination is consistently handled across all paginated requests.
  * I couldn't actually get the Table to send a paginated request when reducing the number of rows per page and switching pages - but the incoming pagination helpers will handle that, so that doesn't block this PR.

AntBush (approved):

PR #292, opened 2025-12-05
Fix the failing GET /user/self endpoint

dwhogan:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-826

## Description of Changes

When refreshing any page in BioHub, a GET request is made to GET /user/self. The initial request fails with a 401 Unauthorized error

Acceptance Critieria
Fix the timing of the request to wait for the user to be authenticated
Remove the authorization middleware from the GET /user/self endpoint
The relevant middleware to remove is lines 12-20 in user/self.ts (the "SystemUser" discriminator, which checks that the user exists in the system_user table).
Keep the ```Bearer: []``` requirement (line 27 in user/self.ts) to require authentication

## Testing Notes

Open browser console, there should NOT be a reference to http://localhost:6100/api/user/self 401 (Unauthorized)

### DIFF
--- a/api/src/paths/user/self.ts
+++ b/api/src/paths/user/self.ts
@@ -2,24 +2,12 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { getDBConnection } from '../../database/db';
 import { HTTP400 } from '../../errors/http-error';
-import { authorizeRequestHandler } from '../../request-handlers/security/authorization';
 import { UserService } from '../../services/user-service';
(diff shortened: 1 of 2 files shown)


dylanrogowsky-oxd (approved):
Tested locally before and after the change. Issue is resolved

### Before (logged out)
before-logged-out

### After (logged out)
after-logged-out

### After (logged in)
after-logged-in

PR #293, opened 2025-12-08
Onboarding Users to BioHub

dwhogan:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-838

## Description of Changes

As a BioHub user, I want to be added as a system user when I first sign in, so that my account is properly created and recognized by the system.

Acceptance Criteria

With a PUT to /api/user/self

1. Check for Existing System User  
WHEN a user logs into BioHub,  
  THEN the system checks whether a matching record exists in the `system_user` table,  
  AND the match should be based on the user's `Keycloak GUID` where `record_end_date` is `NULL`.
See useBiohubUserWrapper.tsx and useAuthStateContext.tsx

2. Existing User Found  
WHEN a matching active system user record exists,  
  THEN the user is allowed to proceed with login without creating a new record. display_name and email are update if required
  AND return 200 with user 

3. No Existing User Found  
WHEN no active system user record is found for the user's Keycloak GUID,  
  THEN the system should allow login,  
  AND should create a new system user record
  AND return 201 with user

4. Invalid or Expired User Record  
WHEN a system user record exists but is invalid or expired (e.g., `record_end_date` is in the past),  
THEN return 401 unauthorized,

5. Required User Fields  
WHEN creating a new system user,  
  THEN the record must include:  
  • `user_identity_source`  
  • `user_identifier`  
  • `user_guid`  
AND a new record should be inserted into `system_user_role` with the 'Member' role
NOTE: Biohub added the following fields after the codebases diverged:
  ADD COLUMN display_name varchar(100)
  ADD COLUMN given_name varchar(100)
  ADD COLUMN family_name varchar(100)
  ADD COLUMN email varchar(100)
  ADD COLUMN agency varchar(100)
  ADD COLUMN notes varchar(250)

6. Auditing and Integrity  
There should be a unique constraint on keycloak guid in the `system_user` table

## Testing Notes
 
  • Remove you self from the sys user seed file

  • Open web page http://localhost:7100/search

  • Check if new user and role records are created

  • Change user's email in system user table

 • Open web page http://localhost:7100/search, check if email is updated in the system users table

 • Deactivated user in the system user table

 • Open web page http://localhost:7100/search check console for 401 status on self endpoint


app/src/hooks/api/useUserApi.ts
```diff
@@ -31,6 +32,48 @@ const useUserApi = (axios: AxiosInstance) => {
...
+   * Get the current user, registering them if they don't exist.
+   *
+   * Handles the get-or-create flow:
+   * 1. Try to get existing user
+   * 2. If 404 (not found), register as new user
+   * 3. If 409 on register (race condition), re-fetch user
+   *
+   * @return {*}  {Promise<ISystemUser>}
+   */
+  const getOrRegisterUser = async (): Promise<ISystemUser> => {
```
mauberti-bc, line 46: This logic can be moved to the backend, to the /self endpoint. 
* If the user exists and is valid -> return the user
* If the user exists but is invalid -> return 401 unauthorized
* If the user does not exist -> create the user

If a user makes a request to the /self endpoint, it should return 200 with the user record unless the user has an existing invalid record in the system_user table, in which case it returns 401 unauthorized.

api/src/paths/user/self.ts
```diff
@@ -163,3 +94,101 @@ export function getUser(): RequestHandler {
...
+
+/**
+ * Register the currently authenticated user as a new system user.
+ *
+ * Extracts user details from the Keycloak JWT token and creates a new system user
+ * with the 'Member' role. Returns 409 if the user already exists.
+ *
+ * @returns {RequestHandler}
+ */
+export function registerUser(): RequestHandler {
```
mauberti-bc: I agree with your comments about making this more like an upsert. See the comment below about handling the 3 scenarios (user exists and is valid, user exists and is invalid, and user doesn't exist).

mauberti-bc (changes requested):
I tested this by removing myself from the DB seeds, then logging in as a new user. I'm getting a Zod error from db-constants.ts -- you might have the ```DATABASE_RESPONSE_VALIDATION_ENABLED``` set to false? It should be true.

The issue is that the SystemUser zod object requires the new fields, but ```const selectServiceAccountsSqlStatement = ...``` on line 63 of db-constants.ts doesn't select them.

api/src/paths/user/self.ts
```diff
@@ -19,93 +30,21 @@ GET.apiDoc = {
   ],
   responses: {
     200: {
-      description: 'User details for the currently authenticated user.',
+      description: 'User profile updated successfully.',
       content: {
         'application/json': {
           schema: {
-            title: 'User Response Object',
```
mauberti-bc, line 26: I think the system user OpenAPI schema defined in ```root-api-doc.ts``` is just missing the new fields (family_name, given_name, etc.)

image

mauberti-bc (changes requested):
Looks great, other than the error!

mauberti-bc (approved):
👍

PR #295, opened 2025-12-09
Managing Teams of Users for Policies

dwhogan:
## Links to Jira Tickets
 
https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-828

## Description of Changes

As a BioHub system administrator, I want to create teams of users that can be assigned to access policies, so that I can minimize the workload of managing individual user access across multiple policies.

Acceptance Criteria
1. Manage Teams Container  
WHEN administrators access the policy management page,  
  THEN a "Manage Teams" container should be present on the same page where policies are created.

2. Team Creation  
WHEN a BioHub administrator wants to create a new team,
THEN the system should allow creating a team with:
a team name,
an optional description, and
a list of users selected from the system user table, excluding SYSTEM and DATABASE users.

3. Editing Teams  
WHEN a BioHub administrator wants to modify an existing team,  
  THEN they should be able to update the team name, description, and membership.

4. Deleting Teams  
WHEN a BioHub administrator decides to remove a team,  
  THEN the system should soft delete the team (add a record end date), with a confirmation dialog to prevent accidental removal

5. Managing Team Members  
WHEN a BioHub administrator wants to update a team’s membership,  
  THEN they should be able to add or remove users from the team.


AntBush (dismissed):
Looks good!

api/src/paths/administrative/users/index.ts
```diff
@@ -0,0 +1,90 @@
...
+export function getAvailableUsers(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const search = req.query.search as string | undefined;
+
+      const knex = getKnex();
```
mauberti-bc: We should keep all sql/DB interactions in the repository layer, like the other endpoints. This index.ts should use the ```userService```:

```
const userService = new UserService(connection)
const users = userService.getAvailableUsers(filters) 
```

And the ```userService.getUsers``` should use the user repository layer:

```
const users = await this.userRepository.getAvailableUsers(filters) 
```

api/src/services/access-policy/team-service.ts
```diff
@@ -65,4 +84,133 @@ export class TeamService extends DBService {
...
+  }
+
+  /**
+   * Get a single team with its members.
+   *
+   * @param {string} teamId - The ID of the team to fetch.
+   * @return {Promise<TeamWithMembers>}
+   * @memberof TeamService
+   */
+  async getTeamWithMembers(teamId: string): Promise<TeamWithMembers> {
```
mauberti-bc, line 137: Promise.all here?

api/src/services/access-policy/team-service.ts
```diff
@@ -65,4 +84,133 @@ export class TeamService extends DBService {
...
+   * Get team members with user details.
+   *
+   * @param {string} teamId - The ID of the team.
+   * @return {Promise<TeamMemberWithUser[]>}
+   * @memberof TeamService
+   */
+  async getTeamMembersWithUsers(teamId: string): Promise<TeamMemberWithUser[]> {
+    const knex = getKnex();
+    const query = knex
+      .table('team_member as tm')
```
mauberti-bc: Like the first comment: we should only put SQL in the repository layer, so this ```getTeamMembersWithUsers``` should call its counterpart repository method, which can also be called ```getTeamMembersWithUsers```.

mauberti-bc:
When the MUI table is set to show 10 rows, the GET request for teams doesn't include limit = 10 for the pagination to work properly. 

In the screenshot, switching to page 2 should make a new request for the next page, which should return a single record.
image

You can find an example of how we handle this in SIMS: https://github.com/bcgov/biohubbc/blob/dev/app/src/features/admin/alert/AlertContainer.tsx. We track paginationModel and sortModel locally, then pass the values to MUI's data grid (see AlertTable.tsx)

mauberti-bc (changes requested):

mauberti-bc (changes requested):
You might still have the ```API_RESPONSE_VALIDATION_ENABLED``` env var false? It should be true to catch these errors. 

image

mauberti-bc (approved):
👍

PR #297, opened 2025-12-18
Search for Features

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-830](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-830?)

## Description of Changes

As a BioHub user, I want to search for BioHub features and submissions using keywords and filters, so that I can quickly find relevant data.

Acceptance Criteria
1. Keyword Search  
WHEN a user enters a keyword into the search bar,  
  THEN the backend `/search` endpoint should split the keyword string into individual terms,  
  AND each term should be treated independently,  
  AND for each term, matches should be searched in the `search_string` table,  
  AND results should be aggregated across all terms,  
  AND the relevant features should be ranked and returned in order of relevancy.
2. Property Match Search  
WHEN a user provides a value for a specific property field (e.g., species, date, feature type),  
  THEN the backend `/search` endpoint should match the value against that property type,  
  AND multiple property fields can be provided,  
  AND results should be aggregated across all property filters,  
  AND results should be ranked by relevancy.
4. Relevancy Ranking  
WHEN the backend processes a search query,  
  THEN a relevancy should be computed (eg. with `tsquery` and `tsvector`),
  AND higher-scoring results should appear earlier in the returned list.
5. Feature & Submission Coverage  
WHEN a search is performed,  
  THEN the system should search across the search_* tables to find relevant features

Notes 

For the keyword search, it's just matching on the value field of the search_string table.
eg. If I search "Moose", all features with any property containing the word "moose" are returned.
For the property match search, it's matching on both the value field of the search_string table AND the property type.
eg. Instead of just passing "Moose", I say `focal_species = Moose`

## Testing Notes

- find a term in the search_string table to test the search with

### DIFF
--- a/app/src/features/submissions/list/SubmissionsListPage.tsx
+++ b/app/src/features/submissions/list/SubmissionsListPage.tsx
@@ -280,7 +280,7 @@ const SubmissionsListPage = () => {
             <Typography mb={1} variant="h1" sx={{ color: '#013366' }}>
               BioHub BC
             </Typography>
-            <Typography variant="body1" color="textSecondary">
+            <Typography color="textSecondary">
               Open access to British Columbia's terrestrial, aquatic species and habitat inventory data.
             </Typography>
           </Box>
--- a/app/src/features/submissions/list/SubmissionsListPage.tsx
+++ b/app/src/features/submissions/list/SubmissionsListPage.tsx
@@ -1,45 +1,205 @@
+import { mdiClose } from '@mdi/js';
+import { Icon } from '@mdi/react';
 import Box from '@mui/material/Box';
-import Container from '@mui/material/Container';
-import Paper from '@mui/material/Paper';
+import IconButton from '@mui/material/IconButton';
 import Typography from '@mui/material/Typography';
 import RecordsFoundSkeletonLoader from 'components/skeleton/submission-card/RecordsFoundSkeletonLoader';
 import SubmissionCardSkeletonLoader from 'components/skeleton/submission-card/SubmissionCardSkeletonLoader';
-import { SearchInput } from 'features/search/SearchComponent';
-import SubmissionsList from 'features/submissions/list/SubmissionsList';
-import SubmissionsListSortMenu from 'features/submissions/list/SubmissionsListSortMenu';
+import {
+  ActiveFilter,
+  FilterSidebar,
+  FilterGroupType,
+  SearchInput,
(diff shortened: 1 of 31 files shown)


app/src/features/submissions/list/SubmissionsListPage.tsx
```diff
@@ -51,61 +211,102 @@ const SubmissionsListPage = () => {
...
-          elevation={0}
-          sx={{
-            py: 4
-          }}>
-          <Container maxWidth="xl">
-            <Typography mb={1} variant="h1">
+      {/* Main layout: Sidebar | Content (full height from nav header) */}
+      <Box sx={{ display: 'flex', minHeight: 'calc(100vh - 70px)' }}>
+        {/* Filter Sidebar */}
+        <FilterSidebar
```
mauberti-bc, line 270: When typing in the main search bar, there's a delay between typing and seeing the new input value. It looks like it's caused by the ```FilterSidebar``` component re-rendering whenever the inputValue state changes -- the FilterSidebar is pretty heavy.

The simplest fix is probably to memoize the FilterSidebar component, which we should do in this PR. Our long term plan should be to separate the sidebar's logic from the search results to prevent re-rendering - but we can make that change in a separate PR since we're going to restyle the sidebar anyway.

app/src/features/submissions/list/SubmissionsListPage.tsx
```diff
@@ -51,61 +211,102 @@ const SubmissionsListPage = () => {
...
+
+        {/* Content area: Header + Search + Results */}
+        <Box sx={{ flex: 1, display: 'flex', flexDirection: 'column' }}>
+          {/* Page Header */}
+          <Box sx={{ py: 3, px: 4, borderBottom: '1px solid #D8D8D8' }}>
+            <Typography mb={1} variant="h1" sx={{ color: '#013366' }}>
               BioHub BC
             </Typography>
-            <Typography mb={3} variant="body1" color="textSecondary">
+            <Typography variant="body1" color="textSecondary">
```
mauberti-bc: nit: ```variant="body1"``` is the default so it can be omitted as a prop

mauberti-bc (dismissed):
Looks great overall. We'll need to change the frontend UX, but we can do that in a separate PR.

mauberti-bc (approved):
👍

PR #303, opened 2026-01-08
Enrich Submission Validation

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-845](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-845)

## Description of Changes
When a submission is being validated, each feature in the submission must be checked to ensure it conforms to the expected structural and allowed data combinations.

A feature has an ID, a feature_type, a properties object, and a references to other features. Validation ensures that only supported feature types are used, that properties are valid for the given feature type, and that all references resolve to existing features within the submission.

## Acceptance Criteria
1. Feature Structure Validation
WHEN a submission enters the validation process  
THEN each feature must include an ID, a feature_type, a properties object, and a references collection  
WHEN a feature is missing a required field  
THEN the submission must fail validation with an error identifying the missing field  

2. Feature Type Validation
WHEN a feature declares a feature_type  
THEN the feature_type must exist in the list of supported feature types (exists in feature_type table with null record_end_date)
WHEN a feature declares an unsupported feature_type  
THEN the submission must fail validation with an error identifying the invalid feature_type  

3. Feature Property Validation
WHEN a feature includes properties  
THEN each property must be explicitly allowed for the declared feature_type  
WHEN a feature includes a property that is not allowed for its feature_type  
THEN the submission must fail validation with an error identifying the invalid property  
WHEN a required property for a feature_type is missing  
THEN the submission must fail validation with an error identifying the missing property  

4. Feature Reference Validation
WHEN a feature includes references to other features  
THEN each referenced feature ID must exist within the submission  
WHEN a feature references a feature ID that does not exist  
THEN the submission must fail validation with an error identifying the invalid reference  

5. Validation Result
WHEN all features in the submission pass validation  
THEN the submission is considered valid  
AND the features should be inserted into the "submission_feature" table
WHEN one or more validation rules fail  
THEN the submission must be rejected and return structured validation errors describing all validation failures  

## Testing Notes

**Unit tests:** cover validation logic: `npm test` in `api/`

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -295,7 +295,6 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.958.0.tgz",
       "integrity": "sha512-ol8Sw37AToBWb6PjRuT/Wu40SrrZSA0N4F7U3yTkjUNX0lirfO1VFLZ0hZtZplVJv8GNPITbiczxQ8VjxESXxg==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@aws-crypto/sha1-browser": "5.2.0",
         "@aws-crypto/sha256-browser": "5.2.0",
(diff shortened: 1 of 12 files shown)


mauberti-bc (approved):
Looks great!

PR #304, opened 2026-01-12
SIMSBIOHUB-793: Fix for service account

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-793: Sending SIMS submissions to TAR file](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-793)

## Description of Changes

**Problem:**
Service client users (e.g., `service-account-sims-svc-****`) were failing to authenticate when calling `/api/submission/upload/archive`, resulting in a "Failed to set user context" error. The `api_set_context` database function couldn't find the user because it was being queried with the wrong identity source.

**Root Cause:**
Service client tokens don't include an `identity_provider` field, causing `getUserIdentitySource` to default to `DATABASE` identity source. However, service clients are stored in the database with `SYSTEM` identity source, creating a mismatch that prevented the user lookup from succeeding.

**Solution:**
Updated `_setUserContext` in `api/src/database/db.ts` to check if the token belongs to a known service client (using `getServiceClientSystemUser`, which matches by the token's `sub` field). If it's a known service client, the identity source is overridden to `SYSTEM` instead of defaulting to `DATABASE`.

**Changes:**
- Modified `api/src/database/db.ts` to check for known service clients before setting user context
- Added import for `getServiceClientSystemUser` from `keycloak-utils`

## Testing Notes

- Test from SIMS

### DIFF
--- a/api/src/database/db.ts
+++ b/api/src/database/db.ts
@@ -6,7 +6,7 @@ import { SYSTEM_IDENTITY_SOURCE } from '../constants/database';
 import { ApiExecuteSQLError, ApiGeneralError } from '../errors/api-error';
 import * as UserQueries from '../queries/database/user-context-queries';
 import { SystemUser } from '../repositories/user-repository';
-import { getUserGuid, getUserIdentitySource } from '../utils/keycloak-utils';
+import { getServiceClientSystemUser, getUserGuid, getUserIdentitySource } from '../utils/keycloak-utils';
 import { getLogger } from '../utils/logger';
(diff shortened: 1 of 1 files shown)


mauberti-bc (approved):
👍 

Future work: Might be worth making identity source required. The codes PR has a `contributor` concept which should help with this, by inserting a system_user record for every service account that sends a POST to `/contributor`. We can set the identity source during that insert.

PR #312, opened 2026-01-22
SIMSBIOHUB-852: Fix migration duplicate key error for feature types

dylanrogowsky-oxd:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-852

## Description of Changes

- Fixed migration `20251001000000_insert_feature_types.ts` to handle existing data gracefully
- Replaced all `INSERT` statements with `INSERT ... SELECT ... WHERE NOT EXISTS` pattern to prevent duplicate key constraint violations
- Updated four INSERT operations:
  1. `feature_type` table - checks for existing records by name
  2. `feature_property_type` table - checks for existing records by name
  3. `feature_property` table - checks for existing records by name
  4. `feature_type_property` table - checks for existing records by feature_type_id and feature_property_id
- Migration is now idempotent and can be safely re-run without failing on duplicate key errors
- This resolves the error: `duplicate key value violates unique constraint "feature_type_nuk1"` that was occurring during TEST deployment

## Testing Notes

- Migration should be tested in TEST environment to verify it completes successfully
- Verify that existing feature types, feature properties, and their relationships are not duplicated
- Migration should be idempotent - running it multiple times should not cause errors or create duplicate records
- Check that all feature types and properties are correctly inserted if they don't already exist
- Verify the migration can be run on a database that already has some of these records (partial data scenario)

### DIFF
--- a/database/src/migrations/20251001000000_insert_feature_types.ts
+++ b/database/src/migrations/20251001000000_insert_feature_types.ts
@@ -16,7 +16,7 @@ export async function up(knex: Knex): Promise<void> {
     -- Insert feature types
     ----------------------------------------------------------------------------------------
     INSERT INTO feature_type (name, display_name, description)
-    VALUES 
+    SELECT * FROM (VALUES 
         (
             'animal',
             'Animal',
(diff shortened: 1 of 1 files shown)


mauberti-bc (approved):
Nice solution

PR #316, opened 2026-01-29
SIMSBIOHUB-823 Creating download package

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-823](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-823)

## Description of Changes

As a BioHub client, I want to download the content I have selected from my shopping cart, so that I can retrieve the features I am authorized to access in a single package.

Acceptance Criteria
1. Initiating a Download Task
   - WHEN a client clicks the "Checkout" button from their shopping cart,
     THEN DownloadService.createDownloadRequest() should create a download record with status pending from the cart's submission_feature_ids and optional fragment_size_mb,
     AND `publishProcessDownloadJob()` should queue a background task to process the request.

2. Background Processing of Downloads
   - WHEN a worker picks up a pending download task from the queue,
     THEN it should collect all features associated with the download request,
     AND package them into a single zip file in object storage.
   - WHEN the package has been successfully created,
     THEN the `download` record's status should be updated to `ready`,
     AND a download link should be made available to the client.

3. Completing the Download
   - WHEN the client downloads the zip package,
     THEN the `download` record's status should be updated to `downloaded`.

4. Download Share/Ownership
   - WHEN an authenticated user initiates a download via cart checkout,
     THEN the download record's system_user_id should be set to the current user,
     AND no entry should be created in download_share for the creator.
   - WHEN an anonymous user initiates a download via cart checkout,
     THEN the download record's system_user_id should be NULL,
     AND the download_id UUID serves as the sole credential for accessing the download.
   - WHEN a download owner shares a download with another user,
     THEN a `download_share` record should be created for that user,
     AND that user should be able to access the download and see it in their download list.

5. Claiming an Anonymous Download
   - WHEN an authenticated user sends PUT /api/download/{downloadId} for an anonymous download (system_user_id IS NULL and team_id IS NULL),
     THEN system_user_id should be set to the current user,
     AND the download should appear in their download list.
   - WHEN a user attempts to claim a download that is already owned or tied to a data request,
     THEN the API should return 409 Conflict.

6. Download Access Control
   - WHEN a download has system_user_id set OR team_id set,
     THEN the detail and fragment URL endpoints should require authentication.
   - WHEN a download has both system_user_id and team_id NULL (anonymous),
     THEN the detail and fragment URL endpoints should be accessible without authentication.
   - WHEN an authenticated user requests GET /api/download (list),
     THEN the response should include downloads they own (system_user_id), downloads shared with them (download_share), and downloads via approved data requests (team_id → team_member).

## Notes
A dedicated download task queue should exist to process all download requests in order.
The process should ensure that only authorized features are included in the zip file.
Status tracking (`pending` → `ready` → `downloaded`) provides transparency for both clients and administrators.

## Testing Notes

There is no endpoint to start the download, instead a TBD endpoint (ie `POST /api/cart/{cartId}/checkout`) should call `DownloadService.createDownloadRequest()`  and `publishProcessDownloadJob()` to initiate the background download.

To test full download worker flow (publish job → worker picks up → zip → S3 → status ready) run the integration test below

* `make test-sys SPEC=download-system`


api/src/models/download.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const DownloadRecord = z.object({
+  download_id: z.number(),
+  system_user_id: z.number(),
+  job_id: z.string().uuid().nullable(),
+  process_status: ProcessStatusStatusZod,
+  s3_key: z.string().nullable(),
+  file_name: z.string().nullable(),
+  file_size_bytes: z.number().nullable(),
+  metadata: z.record(z.unknown()).nullable(),
```
mauberti-bc: We should use `z.object({})` instead of `z.record()` for jsonb fields. I would expect an error like `failed to read undefined while reading (values)` related to zod trying to serialize the value when using `z.record()`.

api/src/repositories/download-repository.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * Get a download record by ID.
+   *
+   * @param {number} downloadId - The download ID.
+   * @return {Promise<DownloadRecord | null>}
+   * @memberof DownloadRepository
+   */
+  async getDownloadById(downloadId: number): Promise<DownloadRecord | null> {
```
mauberti-bc: If a data-fetching method can potentially return null, it should be named `find*`. On the other hand, methods prefixed with `get* `should not return null and should instead throw an error if the data is not found.

api/src/models/download.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ProcessStatusStatusZod } from './process-status';
+
+export const DownloadRecord = z.object({
+  download_id: z.number(),
+  system_user_id: z.number(),
+  job_id: z.string().uuid().nullable(),
+  process_status: ProcessStatusStatusZod,
+  s3_key: z.string().nullable(),
+  file_name: z.string().nullable(),
+  file_size_bytes: z.number().nullable(),
```
mauberti-bc: `file_size_bytes` is BIGINT, so I expect postgres to return it as a string - so the zod object should be `file_size_bytes: z.string().nullable()`. We could update `api/src/database/db.ts` to include a custom handler (lines ~40-57) for converting to a JS BigInt, but we'll need to cast back to a string for the API response anyway.

api/src/repositories/download-repository.ts
```diff
@@ -0,0 +1,187 @@
...
+  ): Promise<void> {
+    const sql = SQL`
+      UPDATE download
+      SET
+        process_status = ${processStatus},
+        s3_key = COALESCE(${metadata?.s3_key ?? null}, s3_key),
+        file_name = COALESCE(${metadata?.file_name ?? null}, file_name),
+        file_size_bytes = COALESCE(${metadata?.file_size_bytes ?? null}, file_size_bytes),
+        metadata = ${JSON.stringify(metadata ?? null)}::jsonb,
+        started_at = CASE WHEN ${processStatus} = 'pending' THEN now() ELSE started_at END,
```
mauberti-bc: We should move any `CASE WHEN` business logic into the service layer, to make it more visible.

api/src/repositories/download-repository.ts
dwhogan: Good convention. Thanks!

api/src/models/download.ts
dwhogan: Moved to z.string().nullable(), very insightful, thanks.

api/src/repositories/download-repository.ts
dwhogan: Moved logic to service layer, thanks!

api/src/models/download.ts
dwhogan: Yes z.object({}) is a better code, thanks.

api/src/paths/download/{downloadId}/fragment/{fragmentIndex}/url/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  }),
+  getFragmentUrl()
+];
+
+GET.apiDoc = {
+  description: 'Get a signed URL to download a specific fragment',
+  tags: ['download'],
+  security: [
+    {
+      Bearer: []
```
mauberti-bc: For consideration:

To enable anonymous downloads, authentication should mirror the `cart` authetnication pattern: 
- `OptionalBearer` here to allow unauthenticated requests
- Unauthenticated download requests will have a null `system_user_id` (and can therefore be accessed by anyone, as long as you know the download UUID)
- Authenticated download requests will use the requestor's `system_user_id`, making only that user authorized to access the download

api/src/paths/download/{downloadId}/fragment/{fragmentIndex}/url/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+  ],
+  parameters: [
+    {
+      description: 'Download ID',
+      in: 'path',
+      name: 'downloadId',
+      schema: {
+        type: 'integer',
```
mauberti-bc: Suggestion: Use a UUID, especially if we support non-authenticated downloads. We'll also eventually have a frontend route for: /:download_id, which we'll send in an email for users to find the presigned URLs, so using UUIDs feels a bit safer.

api/src/repositories/download/download-fragment-repository.ts
```diff
@@ -0,0 +1,270 @@
...
+   */
+  async *streamFragmentFeaturesByType(
+    downloadFragmentId: number,
+    featureTypeName: string,
+    batchSize = 5000
+  ): AsyncGenerator<DownloadFeatureData[]> {
+    const cursorName = 'fragment_feature_cursor';
+
+    await this.connection.query(
+      `DECLARE ${cursorName} CURSOR FOR
```
mauberti-bc, line 144: Neat!

api/src/services/download/download-pipeline-service.ts
```diff
@@ -0,0 +1,644 @@
...
+   * @memberof DownloadService
+   */
+  async planFragments(downloadId: number, sizeEstimate: DownloadSizeEstimate): Promise<void> {
+    const features = sizeEstimate.features;
+
+    // Read configurable fragment size from the download record
+    const download = await this.downloadRepository.findDownloadById(downloadId);
+    const fragmentThreshold = Number(download?.fragment_size_bytes ?? FRAGMENT_SIZE_THRESHOLD);
+
+    // Greedy bin packing using fragment_size_bytes as the bin capacity
```
mauberti-bc, line 239: 👍

api/src/services/download/download-pipeline-service.ts
```diff
@@ -0,0 +1,644 @@
...
+    }
+    return lookup;
+  }
+
+  /**
+   * Stream a single feature type's data as a CSV into the archive.
+   *
+   * Returns file references for any artifact_key properties encountered.
+   */
+  private async streamFeatureTypeCsv(
```
mauberti-bc, line 562: Nit: Update the jsdocs to include params with descriptions of each

database/src/migrations/20260127000000_download.ts
```diff
@@ -0,0 +1,145 @@
...
+      'pending',     -- Queued for processing
+      'ready',       -- Package created, available for download
+      'downloaded',  -- User has downloaded the package
+      'failed'       -- Processing failed
+    );
+
+    --------------------------------------------------------------------------------
+    -- Create download table
+    --------------------------------------------------------------------------------
+    CREATE TABLE download (
```
mauberti-bc, line 30: Add `record_end_date`, which will let users (soft) delete download records from the UI

database/src/migrations/20260127000000_download.ts
```diff
@@ -0,0 +1,145 @@
...
+      download_id         INTEGER GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+      system_user_id      INTEGER NOT NULL,
```
mauberti-bc: When we approve a data request, I think we'll want to create a download package for the requested data. In this case, the download should reference a `team_id` instead of `system_user_id`?
 
Suggestion: 
- Add a separate `download_team` table 
- Create a team during the download request, and join it to the download
- For authorization, check whether the user has access to the download via team membership

This is thinking ahead to when downloads have URL routes, so the user could click "Share" on their download and enter emails/system_user_ids. The API would then add those users to the relevant `team` to grant access.

mauberti-bc (changes requested):

mauberti-bc:
Looks great! In the comments I left, the only one required for this PR is 
- Changing the **submission_id from an integer to UUID**. 
- Adding `record_end_date` to `download`

- The addition of `download_team` and authorization changes to enable anonymous downloads might be better for separate PRs (we will discuss).

Also, check the 11 code smells and address any valid issues.

database/src/migrations/20260127000000_download.ts
dwhogan: Thanks for the catch!

api/src/services/download/download-pipeline-service.ts
dwhogan: Yes JSDoc make for better code, thanks!

api/src/services/download/download-pipeline-service.ts
dwhogan: A Google Takeout inspired approach.

api/src/repositories/download/download-fragment-repository.ts
dwhogan: batchSize can be tuned to balance time and POD memory.

api/src/paths/download/{downloadId}/fragment/{fragmentIndex}/url/index.ts
dwhogan: `OptionalBearer`...very pythonic 👍

api/src/paths/download/{downloadId}/fragment/{fragmentIndex}/url/index.ts
dwhogan: Good call.
Thanks

mauberti-bc (dismissed):
Looks good

database/src/migrations/20260127000000_download.ts
dwhogan: OK added the feature.

mauberti-bc (dismissed):
Nice

mauberti-bc (approved):
👍

PR #318, opened 2026-01-31
SIMSBIOHUB-822: Frontend Cart for Downloads

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-822

## Description of Changes

- Adds a **shopping cart sidebar** and frontend `CartContextProvider` for centralized cart state management.
- Persists cart state via backend cart APIs; stores the active `cart_id` in **sessionStorage** to survive page refreshes within the same session.
- Implements **optimistic add / remove / clear** operations for immediate UI feedback, with rollback on API failure.
- Uses a **single-operation lock** to avoid race conditions from concurrent cart mutations.
- Automatically **claims an anonymous cart on login**, assigning the active cart to the authenticated user (idempotent backend operation).
- Recovers from invalid or unauthorized carts: on **401/403 during add**, a **new cart is created** and pending features are re-added.

## Testing Notes

- To add an item, click the `+` plus icon on a search result card
- To remove cart items, the `+` icon turns into a checkmark - click the checmark to remove, OR click the `X` icon in the cart.
- Anonymous flow: add items while logged out → refresh → cart persists and reloads correctly.
- Login flow: add items while logged out → log in → cart remains and is assigned to the user.
- Optimistic UI: verify immediate UI updates on add/remove/clear; simulate API failures to confirm rollback.
- Duplication guard: adding the same feature multiple times does not create duplicates.
- Access error recovery: simulate 401/403 during add → new cart is created and features are retained.


dwhogan (dismissed):
`app/src/contexts/cartContext.tsx:300-330`
await cartApiRef.current.assignCartToCurrentUser(cartId);
await cartApiRef.current.getCartById(cartId);
are concurrent HTTP requests, I think for the happy path were good, but maybe consider moving the claim into the load effect's success path to handle unforeseen events.
`api/src/paths/cart/{cartId}/index.ts:131`
Handle findCartById return null
`api/src/paths/cart/{cartId}/index.ts:97`
Do you need the ...paginationRequestQueryParamSchema?

dwhogan (approved):
console.error('Failed to claim cart for authenticated user:', error); is WIP?
findCartWithFeaturesById is unused?
operationInProgress.current returns without feedback to user
consider removing `state` from addToCart dependencies with a ref to reduce re-renders

PR #319, opened 2026-02-05
SIMSBIOHUB-861: Fix Winston Logger

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-861

## Description of Changes

- The logger was suppressing some errors messages when Zod schema validation failed
- Replaces `winston.format.prettyPrint` with `winston.format.printf`
- Masks any values suspected to be secrets, where the key contains 'secret', 'token', 'password', etc.

## Testing Notes

- Logs should display like before in the docker container, although no longer coloured

### DIFF
--- a/api/src/utils/logger.test.ts
+++ b/api/src/utils/logger.test.ts
@@ -1,13 +1,18 @@
 import { expect } from 'chai';
-import { describe } from 'mocha';
+import { describe, it } from 'mocha';
 import { getLogger } from './logger';
 
 describe('logger', () => {
   describe('getLogger', () => {
-    it('returns a winston logger', () => {
+    it('returns a logger wrapper with standard log methods', () => {
       const logger = getLogger('myLogger');
 
-      expect(logger).not.to.be.undefined;
+      expect(logger).to.be.an('object');
+      expect(logger).to.have.property('info').that.is.a('function');
+      expect(logger).to.have.property('error').that.is.a('function');
+      expect(logger).to.have.property('warn').that.is.a('function');
+      expect(logger).to.have.property('debug').that.is.a('function');
+      expect(logger).to.have.property('silly').that.is.a('function');
     });
   });
 });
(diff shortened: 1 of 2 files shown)


dylanrogowsky-oxd (changes requested):
## PR 319 review: Fix Winston Logger

Summary of review for **SIMSBIOHUB-861**.

### What looks good
- **Secret masking** with `SENSITIVE_KEY_REGEX` and `redactSecrets` is a solid security improvement.
- **Type safety**: Switching `CustomLoggerParams` from `any` to `unknown` is an improvement.
- **Stability**: Replacing `prettyPrint` with `printf` and adding `winston.format.errors({ stack: true })` should address Zod/error messages being suppressed.
- **Runtime log level**: `setLogLevel` / `setLogLevelFile` now select transports by type instead of by index, which is more robust.
- No remaining usages of the old exported helpers (`_getLoggerParameters`, etc.) in the repo.

### Critical: Error objects lost in logs

When callers use the usual pattern `log.error({ message: '...', error })`, the `error` (an `Error` instance) lives in **metadata**. Winston’s `format.errors({ stack: true })` only serializes errors passed **directly** to the logger, not nested inside an object, so it never sees `metadata.error`.

`redactSecrets` then receives that raw `Error`. On `Error`, `Object.entries(error)` is effectively `[]` because `message` and `stack` are non-enumerable, so the function returns `{}` and **error message and stack are dropped** from the log.

**Recommendation:** In `redactSecrets`, treat `Error` instances explicitly (e.g. `if (value instanceof Error) return { message: value.message, stack: value.stack };`) before the generic object branch, so nested errors are serialized and still redact any sensitive sub-properties if needed.

### Minor / suggestions
- **Tests:** Consider adding a test that logs an object with a `password` (or similar) key and asserts it’s masked, and one that logs `{ error: new Error('test') }` and asserts message/stack appear. That would lock in both secret redaction and error serialization.
- **PR description:** It says “no longer coloured” in Docker; the code still uses `winston.format.colorize({ all: true })` for console. If Docker strips or doesn’t show ANSI codes, that’s expected; otherwise the note might be worth a quick clarification.

Requesting changes only for the Error-handling bug; the rest is optional polish.

dylanrogowsky-oxd (approved):

PR #321, opened 2026-02-06
SIMSBIOHUB-868: Fix access policy seed duplicate key error in TEST

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-868](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-868)

## Description of Changes

- **Access policy seed is now idempotent.** The seed `07_access_policy.ts` was always inserting the same two policies ("Telemetry Access", "Secret Sampling Sites Access") and their teams. In test, when the seed runs against a DB that already had that data (e.g. re-run or existing test data), it blew up with:

  ```
  Error while executing ".../07_access_policy.ts" seed: insert into "policy" ("description", "name") values ($1, $2) returning * - duplicate key value violates unique constraint "policy_nuk1"
  ```

- **What changed:** We only insert when the row doesn’t already exist. For policies and teams we do a get-or-create (select by name; insert only if missing). For `team_policy`, `policy_statement`, and `policy_statement_condition` we check for an existing row with the same natural key before inserting. The bulk insert that adds non-system users to the Telemetry team now skips users who are already in that team via a `NOT EXISTS` check so we don’t hit the unique on `(system_user_id, team_id)`.

- **create_user:** The seed now resolves a valid `create_user` (first active system user, fallback 1) and passes it on all inserts so it still works in environments where the audit trigger doesn’t set it.

## Testing Notes

- Run the DB setup (migrate + seed) twice against the same database, or run it in the test environment where the access policy data may already exist. The seed should complete without the `policy_nuk1` (or related) duplicate key error. No change to behaviour when the DB is empty—the same policies and teams are created.

### DIFF
--- a/database/src/seeds/07_access_policy.ts
+++ b/database/src/seeds/07_access_policy.ts
@@ -1,7 +1,8 @@
 import { Knex } from 'knex';
 
 /**
- * Inserts default access policies for accessing secured data
+ * Inserts default access policies for accessing secured data.
+ * Idempotent: safe to run when policies/teams already exist (e.g. re-run in test).
  *
  * @export
  * @param {Knex} knex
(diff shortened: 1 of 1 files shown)


mauberti-bc (approved):
👍

PR #331, opened 2026-02-11
SIMSBIOHUB-870: Fix feature properties and feature-type assignments

dylanrogowsky-oxd:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-870

## Description of Changes

- New migration `20260211000000_fix_feature_properties_and_assignments.ts` that corrects several feature-property settings from the original insert migration:
  - **Computed flag:** `filename`, `method_attribute`, `method_value`, `file_type`, and `file_size` are no longer marked as computed (they’re now user-supplied where used).
  - **File feature type:** `filename` and `file_size` are now required.
  - **Report feature type:** Adds required properties `filename`, `file_type`, and `file_size` (reports now validate these like file does).
  - **Sample period:** Replaces the `sample_technique` property with `method_technique_id` (new number property); the old link is soft-deleted and the new one is optional.
  - **Measurement:** Adds optional `description`.
  - **Species observation:** Adds optional `count`.
- Down migration reverses all of the above (restores computed flags, required flags, report links, sample_technique link, and removes the new/added links and the `method_technique_id` property).

## Testing Notes

- Run migrations up on a DB that has already run `20251001000000_insert_feature_types` (and any later migrations). Confirm no errors and that `feature_property` and `feature_type_property` rows match the description above.
- Run migration down and confirm the state matches the pre-migration state (e.g. `sample_technique` back on sample_period, report no longer has filename/file_type/file_size, etc.).
- If you have validation or submission code that depends on these properties, spot-check file/report validation and any UI that uses sample period, measurement, or species observation.

### DIFF
--- a/database/src/migrations/20260211000000_fix_feature_properties_and_assignments.ts
+++ b/database/src/migrations/20260211000000_fix_feature_properties_and_assignments.ts
@@ -1,0 +1,215 @@
+import { Knex } from 'knex';
+
+/**
+ * Fix feature properties and feature-type-property assignments from 20251001000000_insert_feature_types.
+ *
+ * - Set calculated_value = false for filename, method_attribute, method_value, file_type, file_size
+ * - File: filename and file_size required
+ * - Report: add filename, file_type, file_size (all required, not computed)
+ * - Sample_period: replace sample_technique with method_technique_id
+ * - Measurement: add description (optional)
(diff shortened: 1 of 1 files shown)


mauberti-bc (approved):
👍

PR #332, opened 2026-02-12
SIMSBIOHUB-847: Storing Feature Content Array

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-870](https://quartech.atlassian.net/browse/SIMSBIOHUB-870)

## Description of Changes

- **Database:** New junction table `submission_feature_feature` with `source_feature_id` and `target_feature_id`, unique on the pair, FKs to `submission_feature`, and a check so source ≠ target. Unique index on `(LEAST(source_feature_id, target_feature_id), GREATEST(...))` to prevent inverse pairs (A→B and B→A). Indexes on both columns for lookups. Audit/journal triggers added to match other tables.
- **Ingestion:** When flat features are ingested, we clear any existing relationship rows for that submission (with the existing feature soft-delete) for idempotency. After inserting features and updating parent refs, we add a third pass: for each feature with a non-empty `content` array, we insert one row per referenced feature into the junction table (feature = source, each `content` ID = target). Features with no or empty `content` are unchanged.
- **Repository:** `SubmissionRepository` now has `deleteSubmissionFeatureRelationships(submissionId)`, `insertSubmissionFeatureRelationships(pairs)` (bulk insert with ON CONFLICT DO NOTHING; pairs use `source_feature_id` / `target_feature_id`), and `getRelatedSubmissionFeatureIds(submissionFeatureId)` returning `{ sourceIds, targetIds }` so you can reconstruct related features from the table.
- **Tests:** New/updated tests for the three repo methods and for ingestion (relationships created when `content` is present, not created when empty, delete-relationships step, and source-with-multiple-targets).

## Testing Notes

- Run migrations (`database`: `npm run migrate-latest`) before testing

### DIFF
--- a/database/src/migrations/20260212000000_submission_feature_relationships.ts
+++ b/database/src/migrations/20260212000000_submission_feature_relationships.ts
@@ -17,36 +17,36 @@ export async function up(knex: Knex): Promise<void> {
     --------------------------------------------------------------------------------
 
     CREATE TABLE submission_feature__feature (
-      parent_submission_feature_id integer NOT NULL,
-      child_submission_feature_id integer NOT NULL,
+      parent_feature_id integer NOT NULL,
+      child_feature_id integer NOT NULL,
       create_date timestamptz(6) DEFAULT now() NOT NULL,
       create_user integer NOT NULL,
       update_date timestamptz(6),
       update_user integer,
       revision_count integer DEFAULT 0 NOT NULL,
       CONSTRAINT submission_feature__feature_pk
-        PRIMARY KEY (parent_submission_feature_id, child_submission_feature_id),
+        PRIMARY KEY (parent_feature_id, child_feature_id),
       CONSTRAINT submission_feature__feature_parent_fk
-        FOREIGN KEY (parent_submission_feature_id)
+        FOREIGN KEY (parent_feature_id)
         REFERENCES submission_feature(submission_feature_id)
         ON DELETE CASCADE,
       CONSTRAINT submission_feature__feature_child_fk
-        FOREIGN KEY (child_submission_feature_id)
+        FOREIGN KEY (child_feature_id)
         REFERENCES submission_feature(submission_feature_id)
         ON DELETE CASCADE,
       CONSTRAINT submission_feature__feature_no_self_loop
-        CHECK (parent_submission_feature_id != child_submission_feature_id)
+        CHECK (parent_feature_id != child_feature_id)
     );
 
     CREATE INDEX submission_feature__feature_parent_idx
-      ON submission_feature__feature(parent_submission_feature_id);
+      ON submission_feature__feature(parent_feature_id);
 
     CREATE INDEX submission_feature__feature_child_idx
-      ON submission_feature__feature(child_submission_feature_id);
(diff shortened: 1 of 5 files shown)


database/src/migrations/20260212000000_submission_feature_relationships.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    SET SEARCH_PATH = biohub, public;
+
+    --------------------------------------------------------------------------------
+    -- SUBMISSION_FEATURE__FEATURE (many-to-many feature relationships)
+    --------------------------------------------------------------------------------
+
+    CREATE TABLE submission_feature__feature (
```
mauberti-bc: Is double underscore intentional?

database/src/migrations/20260212000000_submission_feature_relationships.ts
```diff
@@ -0,0 +1,77 @@
...
+      parent_feature_id integer NOT NULL,
```
mauberti-bc: This might be irrelevant semantics but:

Instead of `parent` and `child`, consider renaming to `source` (=parent) and `target` (=child). Parent/child seems to imply hierarchy between the features, but they're not necessarily hierarchical; they're just related in some way.

In the submission payload structure, the `content` array does imply hierarchy, so the current table definition works. But it seems valuable to broaden its scope to work for non-hierarchical relationships. 

We could add a `relationship` field to this table to drive that point home (options as ['parent', 'sibling'], for example).

database/src/migrations/20260212000000_submission_feature_relationships.ts
```diff
@@ -0,0 +1,77 @@
...
+        REFERENCES submission_feature(submission_feature_id)
+        ON DELETE CASCADE,
+      CONSTRAINT submission_feature__feature_child_fk
+        FOREIGN KEY (child_feature_id)
+        REFERENCES submission_feature(submission_feature_id)
+        ON DELETE CASCADE,
+      CONSTRAINT submission_feature__feature_no_self_loop
+        CHECK (parent_feature_id != child_feature_id)
+    );
+
```
mauberti-bc, line 42: Consider adding a constraint or check to prevent inverse relationships (e.g., A → B and B → A) to avoid potential cycles in recursive queries.

mauberti-bc (changes requested):

database/src/migrations/20260212000000_submission_feature_relationships.ts
dylanrogowsky-oxd: It was what was in [Jira](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-847), but happy to change

mauberti-bc (approved):
Looks great

database/src/migrations/20260212000000_submission_feature_relationships.ts
mauberti-bc: Sorry, typo! Do you mind replacing and then we'll approve

PR #333, opened 2026-02-12
SIMSBIOHUB 865: data_request api routes

J-pilon:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-865

## Description of Changes

Adds a full data-request feature: models and OpenAPI schemas for data requests, a repository and service (with tests), and API routes for creating, listing, getting by ID, updating, and deleting data requests. Implements authorization so that listing requires a system admin, and get/update/delete by ID require either a system admin or a member of the data request’s team, including a `ModelTeamMemberRecord` type and service methods to check team membership.

---

### Authorization

`POST /data-request` | Must be authenticated
`GET /data-request` | Must be authenticated and system admin
`GET /data-request/:dataRequestId` | Must be authenticated and (system admin or team member)
`PUT /data-request/:dataRequestId` | Must be authenticated and (system admin or team member)
`DELETE /data-request/:dataRequestId` | Must be authenticated and (system admin or team member)

#### List all data requests (system admin)
curl -X GET "http://localhost:6100/api/data-request" -H "Authorization: Bearer YOUR_ACCESS_TOKEN"

List with query params (date_from, date_to, requested_by, team_id)
curl -X GET "http://localhost:6100/api/data-request?date_from=2025-01-01&date_to=2025-12-31&requested_by=123&team_id=550e8400-e29b-41d4-a716-446655440000" -H "Authorization: Bearer YOUR_ACCESS_TOKEN"

Get single data request by ID (system admin or team member)
curl -X GET "http://localhost:6100/api/data-request/550e8400-e29b-41d4-a716-446655440000" -H "Authorization: Bearer YOUR_ACCESS_TOKEN"

Create data request (authenticated)
curl -X POST "http://localhost:6100/api/data-request" -H "Authorization: Bearer YOUR_ACCESS_TOKEN" -H "Content-Type: application/json" -d '{"team_id": "550e8400-e29b-41d4-a716-446655440000", "reason": "Research project data access"}'

Update data request (system or team member)
curl -X PUT "http://localhost:6100/api/data-request/550e8400-e29b-41d4-a716-446655440000" -H "Authorization: Bearer YOUR_ACCESS_TOKEN" -H "Content-Type: application/json" -d '{"reason": "Updated reason for the data request"}'

Delete data request (system admin or team member)
curl -X DELETE "http://localhost:6100/api/data-request/550e8400-e29b-41d4-a716-446655440000" -H "Authorization: Bearer YOUR_ACCESS_TOKEN"

Replace `YOUR_ACCESS_TOKEN` with a valid JWT and use a real `dataRequestId` (e.g. the UUID above) when testing get/update/delete.


api/src/models/data-request.ts
```diff
@@ -0,0 +1,92 @@
...
+// data_request table
+// ──────────────────────────────────────────────────────────────────────────────
+
+export const DataRequest = z.object({
+  data_request_id: z.string().uuid(),
+  reason: z.string(),
+  team_id: z.string().uuid(),
+  requested_by: z.number(),
+  record_end_date: z.string().nullable(),
+  create_date: z.string(),
```
mauberti-bc: Don't send the audit columns (create_*, update_*, revision_count) in the API response unless they're needed in the frontend.

api/src/models/data-request.ts
```diff
@@ -0,0 +1,92 @@
...
+export const UpdateDataRequest = z.object({
+  reason: z.string().optional()
+});
+export type UpdateDataRequest = z.infer<typeof UpdateDataRequest>;
+
+// ──────────────────────────────────────────────────────────────────────────────
+// data_request CRUD Responses
+// ──────────────────────────────────────────────────────────────────────────────
+
+export const DataRequestWithStatus = z.object({
```
mauberti-bc: Suggestion: use `DataRequest.extend({ data_request_status: DataRequestStatus })`

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,265 @@
...
+};
+
+/**
+ * Get a data request by ID.
+ *
+ * @returns {RequestHandler}
+ */
+export function getDataRequestById(): RequestHandler {
+  return async (req, res) => {
+    const isAuthenticated = !!req.keycloak_token;
```
mauberti-bc: If we use bearer authentication:
``` 
security: [
    {
      Bearer: []
    }
  ],
```
Then req.keycloak_token will also be defined, because the authentication middleware will throw before `getDataRequestId` is called. So you can remove isAuthenticated and change the connection to:
```
const connection = getDBConnection(req.keycloak_token)
```

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,265 @@
...
+      await connection.open();
+
+      const dataRequestId = req.params.dataRequestId;
+      const dataRequestService = new DataRequestService(connection);
+
+      const dataRequest = await dataRequestService.updateDataRequest(dataRequestId, req.body);
+
+      await connection.commit();
+
+      res.status(200).json(dataRequest);
```
mauberti-bc: nit: add `return` for consistency

`return res.status(200).json(dataRequest);`

api/src/paths/data-request/index.ts
```diff
@@ -0,0 +1,176 @@
...
+ */
+export function findDataRequests(): RequestHandler {
+  return async (req, res) => {
+    const isAuthenticated = !!req.keycloak_token;
+    const connection = isAuthenticated ? getDBConnection(req.keycloak_token) : getAPIUserDBConnection();
+
+    try {
+      await connection.open();
+
+      const { date_from, date_to, requested_by, team_id } = req.query;
```
mauberti-bc: See `parseQueryParams` in `search/index.ts` as an example util for parsing query params with correct types. The current code is fine but adding a util reduces clutter.

api/src/paths/data-request/index.ts
```diff
@@ -0,0 +1,176 @@
...
+ */
+export function createDataRequest(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const systemUserId = connection.systemUserId();
+      const { team_id: teamId, reason } = req.body;
```
mauberti-bc, line 172: We should make `team_id` optional in the body. If a team is not given, we create a team within  `DataRequestService.createDataRequest`.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+  CreateDataRequest,
+  DataRequest,
+  DataRequestStatus,
+  DataRequestStatusEnum,
+  DataRequestWithStatus,
+  UpdateDataRequest
+} from '../models/data-request';
+import { BaseRepository } from './base-repository';
+
+export type DataRequestFilters = {
```
mauberti-bc: Put interfaces and types in either `dataRequestService.interface.ts` or `models/data-request.ts`. We aren't consistent so the location doesn't really matter, but I'd suggest a separate interface file. The important piece is that we keep type definitions out of the service and repository .ts files.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+   *
+   * @param {object} [filters]
+   * @param {string} [filters.date_from]
+   * @param {string} [filters.date_to]
+   * @param {number} [filters.requested_by]
+   * @param {string} [filters.team_id]
+   * @return {Promise<DataRequest[]>}
+   * @memberof DataRequestRepository
+   */
+  async findDataRequests(filters?: {
```
mauberti-bc: Reference the existing type instead of redeclaring:

async findDataRequests(filters?: DataRequestFilters)

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+    date_from?: string;
+    date_to?: string;
+    requested_by?: number;
+    team_id?: string;
+  }): Promise<DataRequest[]> {
+    const knex = getKnex();
+    const query = knex('data_request').whereNull('record_end_date').select('*');
```
mauberti-bc: Because we don't want to return the audit columns, replace `select *` with selecting the specific fields for the zod schema.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+   */
+  async findDataRequestsByStatus({
+    status,
+    filters
+  }: {
+    status: DataRequestStatusEnum;
+    filters?: DataRequestFilters;
+  }): Promise<DataRequestWithStatus[]> {
+    const knex = getKnex();
+    const query = knex('data_request as dr')
```
mauberti-bc, line 109: Nit: Even though knex lets you chain SQL in whatever order, put select statements first so it reads like SQL, just for readability.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+      drs_request_status: DataRequestStatusEnum;
+      drs_record_end_date: string | null;
+      drs_create_date: string;
+      drs_create_user: number;
+      drs_update_date: string | null;
+      drs_update_user: number | null;
+      drs_revision_count: number;
+    }>;
+
+    return rows.map(
```
mauberti-bc: Don't manipulate the database response using javascript/typescript in the repository layer. To help give each layer in the API a clear purpose, we put maps and other manipulations in the service layer or in the SQL.

The three layers we usually talk about are the endpoints (/paths), services (/services), and repositories (/repositories). Endpoints should just focus on auth, delegating to services, and the req/res. Services are the core business logic. Repositories are for database interactions and should usually just return either response.rows or response.rows[0].

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+
+    if (createDataRequestResponse.rowCount !== 1) {
+      throw new ApiExecuteSQLError('Failed to create data request', [
+        'DataRequestRepository->createDataRequest',
+        'rowCount !== 1'
+      ]);
+    }
+
+    const dataRequest = createDataRequestResponse.rows[0];
+    const dataRequestStatusResponse = await this.createDataRequestStatus(
```
mauberti-bc: Every repository method should `await` only once. For the flow of creating a data_request and then data_request_status record, you should put that in the service layer. 

Another pattern we sometimes use for multi-table inserts is to use CTEs within a single repository method, eg. 

```sql
WITH w_data_request AS (INSERT INTO data_request...) 
INSERT INTO data_request_status...
```

but chaining in the service layer is preferred because it's easier to interpret.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,392 @@
...
+   * @return {Promise<DataRequest>}
+   * @memberof DataRequestRepository
+   */
+  async updateDataRequest(dataRequestId: string, payload: UpdateDataRequest): Promise<DataRequest> {
+    const knex = getKnex();
+    const query = knex('data_request')
+      .where('data_request_id', dataRequestId)
+      .whereNull('record_end_date')
+      .update(payload)
+      .returning('*');
```
mauberti-bc: Because we don't include the audit columns in the `DataRequest` zod object, we can't select * and use the existing zod object. We typically do:
```
.returning('data_request_id')

const response = await this.connection.knex(query, z.object({ data_request_id: z.uuid()} )
```

api/src/services/authorization/authorization-service.ts
```diff
@@ -144,6 +159,9 @@ export class AuthorizationService extends DBService {
         case 'Cart':
           authorizeResults.push(await this.authorizeByCart(authorizeRule));
           break;
+        case 'ModelTeam':
```
mauberti-bc: I don't think we want middleware like this. I'd suggest removing all the middleware changes in this PR, and then we'll make a new ticket for adding the `Team` authorization discriminator, which we can reuse for accessing multiple entity types (downloads, data requests, etc.).

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,265 @@
...
+export const GET: Operation = [
+  authorizeRequestHandler((req) => {
+    return {
+      or: [
+        {
+          validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN],
+          discriminator: 'SystemRole'
+        },
+        {
+          modelId: req.params.dataRequestId,
```
mauberti-bc: See comment below about removing the middleware changes. Leave the `SystemRole` discriminator so these endpoints are only accessible to admins, then we'll add the `Team` discriminator in a separate PR.

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Find all data requests, optionally by status (will default to REQUESTED).
+   *
+   * @param {string} [status]
+   * @return {Promise<DataRequestWithStatus[]>}
+   * @memberof DataRequestRepository
+   */
+  async findDataRequestsByStatus({
+    status,
+    filters
+  }: {
```
mauberti-bc: I don't think this method adds much. 

Status is another filter dimension, so add `status` to `DataRequestFilters` instead of having it as its own prop, and just use the `findDataRequests` method.

```
async findDataRequests(filters?: DataRequestFilters) => {...}
```

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  /**
+   * Checks if the provided user is a member of the team associated with the given data request.
+   *
+   * @param {string} dataRequestId - The data request id to check.
+   * @param {number} userId - The system user id to check for team membership.
+   * @returns {Promise<boolean>} - True if the user is a team member, false otherwise.
+   */
+  async isCurrentUserATeamMember(dataRequestId: string, userId: number): Promise<boolean> {
```
mauberti-bc: Remove this method as part of removing the middleware changes

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * Returns the team record if the given user is a member of the team associated with the data
+   * request; otherwise returns null. Used by authorizeByModelTeamMember.
+   *
+   * @param {string} dataRequestId - The data request id.
+   * @param {number} userId - The system user id to check for team membership.
+   * @return {Promise<ModelTeamMemberRecord | null>}
+   */
+  async findTeamMember(dataRequestId: string, userId: number): Promise<ModelTeamMemberRecord | null> {
+    const dataRequest = await this.findDataRequestById(dataRequestId);
```
mauberti-bc: Returning null if the given data request doesn't exist is misleading. I'd suggest:
* Add `getDataRequestMembers`, which should throw an error if the request isn't found.
* Keep `findDataRequestById` for use cases where we don't want to throw
* Update `findTeamMember` to look more like:
```
const dataRequest = await this.getDataRequestById(dataRequestId)
const members = await teamMemberService.getTeamMembers(dataRequest.team_id)

const member = members.find((member) => member.system_user_id === userId)

if (!member){
   return null
}

return member
```

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+   * @param {object} [filters]
+   * @param {string} [filters.date_from]
+   * @param {string} [filters.date_to]
+   * @param {number} [filters.requested_by]
+   * @param {string} [filters.team_id]
+   * @return {Promise<DataRequest[]>}
+   * @memberof DataRequestRepository
+   */
+  async findDataRequests(filters?: {
+    date_from?: string;
```
mauberti-bc: Reference the existing interface definition for DataRequestFilters instead of redefining

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+   * record identified by modelId; otherwise returns false.
+   *
+   * @param {string} modelId - The id of the model record (e.g. data request id).
+   * @param {number} systemUserId - The system user id to check for team membership.
+   * @return {Promise<boolean>}
+   */
+  isCurrentUserATeamMember(modelId: string, userId: number): Promise<boolean>;
+}
+
+type ModelTeamMemberRecord = {
```
mauberti-bc: Move to separate interface file or `models/data-request.ts`

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,185 @@
...
+import { TeamMemberService } from './access-policy/team-member-service';
+import { DBService } from './db-service';
+
+/**
+ * Service interface for models that are associated with a team. Used by authorizeByModelTeamMember to
+ * determine if the current user is a member of the team for the given model record.
+ *
+ * @export
+ */
+export interface IModelWithTeamMemberService {
```
mauberti-bc: I'd suggest removing this interface when removing the middleware.

mauberti-bc (changes requested):

api/src/models/data-request.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/1cf986c99866c0b72328c2bd6958da66b68790d5

api/src/models/data-request.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/82aeceb570e51a3a2c136f79d582998bbefd6253

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/16666be34201d7c9a276a8cc6c080cee69e5704d

api/src/paths/data-request/index.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/8ef84061e00431a1fc00a094c96d3cce129fca46

api/src/paths/data-request/index.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/f8600eadaebf6d066a7cdb3d2903912b0d3b374e

api/src/repositories/data-request-repository.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/02449821f4a226f40aa5775ad8df925b0e7271ef

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/0e7682393397169b2b39cc57ba298bdd8fe365d1

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/services/authorization/authorization-service.ts
J-pilon: Resolved.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/changes/cfe1c6727ccbbeae8cf35af2d3c0cff1a909c58e

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,242 @@
...
+import { getLogger } from '../../../utils/logger';
+
+const defaultLog = getLogger('paths/data-request/{dataRequestId}');
+
+export const GET: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser'
```
mauberti-bc: We should use the `SystemRole` discrminiator with `validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN]`. This `SystemUser` discriminator just checks that a user exists, but only admins are authorized for the data request endpoints.

Check out the middleware for endpoints under `/api/administrative/*`, like `administrative/policies/index.ts`,for examples.

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,242 @@
...
+    };
+  }),
+  getDataRequestById()
+];
+export const PUT: Operation = [
+  authorizeRequestHandler(() => {
+    return {
+      and: [
+        {
+          discriminator: 'SystemUser'
```
mauberti-bc: Same middleware change.

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -0,0 +1,242 @@
...
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const dataRequestId = req.params.dataRequestId;
+      const dataRequestService = new DataRequestService(connection);
+
+      const dataRequest = await dataRequestService.getDataRequestById(dataRequestId);
```
mauberti-bc, line 170: Make sure you call `connection.commit()` before returning, otherwise the DB changes won't apply.

api/src/paths/data-request/index.ts
```diff
@@ -0,0 +1,203 @@
...
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const filters = parseQueryParams(req);
+
+      const dataRequestService = new DataRequestService(connection);
+      const dataRequests = await dataRequestService.findDataRequests(filters);
```
mauberti-bc, line 144: Missing `connection.commit()`

api/src/paths/data-request/index.ts
```diff
@@ -0,0 +1,203 @@
...
+      const systemUserId = connection.systemUserId();
+      const { team_id: teamId, reason } = req.body;
+
+      const dataRequestService = new DataRequestService(connection);
+
+      const dataRequest = await dataRequestService.createDataRequest(systemUserId, { reason }, teamId);
+
+      await connection.commit();
+
+      res.status(201).json(dataRequest);
```
mauberti-bc: nit: `return res.status(201).json(dataRequest);` instead of just `res.status(201).json(dataRequest);`.

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,285 @@
...
+ */
+export class DataRequestRepository extends BaseRepository {
+  /**
+   * Find all data requests, optionally filtered by date range, requested_by, team_id, or status.
+   *
+   * @param {DataRequestFilters} [filters] - Optional filters (date_from, date_to, requested_by, team_id, status).
+   * @return {Promise<DataRequest[]>}
+   * @memberof DataRequestRepository
+   */
+  async findDataRequests(filters?: DataRequestFilters): Promise<DataRequest[]> {
```
mauberti-bc: Every repository method should `await` once, and for clarity/consistency, we typically define the queryBuilder and then conditionally apply filters.

```
const knex = getKnex()

const queryBuilder = knex.select('...').from(..) // the main query

if (filters){
    queryBuilder.where(...)
}

const response = await this.connection.knex(queryBuilder, <zodObject>)
```

I'd suggest restructuring these repository methods to define the base queryBuilder, and then just conditionally add `where` statements.

Example:
```
async getSourceTransformRecordBySystemUserId(systemUserId: number, version?: string): Promise<ISourceTransformModel> {
    const queryBuilder = getKnexQueryBuilder()
      .select('*')
      .from('source_transform')
      .where('system_user_id', systemUserId)
      .and.where('record_end_date', null);

    if (version) {
      queryBuilder.where('version', version);
    }

    const response = await this.connection.knex(queryBuilder);

    if (!response.rowCount) {
      throw new ApiExecuteSQLError('Failed to get submission source transform record', [
        'SubmissionRepository->getSourceTransformRecordBySystemUserId',
        'rowCount was null or undefined, expected rowCount != 0'
      ]);
    }

    return response.rows[0];
  }
```

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+   * Get all status records for a data request.
+   *
+   * @param {string} dataRequestId
+   * @return {Promise<DataRequestStatus[]>}
+   * @memberof DataRequestRepository
+   */
+  async getDataRequestStatuses(dataRequestId: string): Promise<DataRequestStatus[]> {
+    const knex = getKnex();
```
mauberti-bc: Every database table should have its own repository, so move these `data_request_status` methods to a new `data-request-status-repository.ts`

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,285 @@
...
+  /**
+   * Create a new comment.
+   *
+   * @param {string} comment
+   * @return {Promise<Comment>}
+   * @memberof DataRequestRepository
+   */
+  async createComment(comment: string): Promise<Comment> {
+    const knex = getKnex();
+    const query = knex('comment').insert({ comment }).returning(['comment', 'comment_id']);
```
mauberti-bc: Create a separate `comment` repository/service.

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,215 @@
...
+   * @param {number} requestedBy - system user id
+   * @param {CreateDataRequest} payload
+   * @param {string} [teamId]
+   * @return {Promise<DataRequestWithStatus>}
+   * @memberof DataRequestService
+   */
+  async createDataRequest(
+    requestedBy: number,
+    payload: CreateDataRequest,
+    teamId?: string
```
mauberti-bc: Move `teamId` into the `payload` as an optional param, instead of passing a third prop.

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  /**
+   * Returns true if the current user may act on this data request (system admin or team member); otherwise false.
+   *
+   * @param teamId
+   */
+  async _authorizeAccessForDataRequest(teamId: string): Promise<boolean> {
+    const userService = new UserService(this.connection);
+    const isSystemAdmin = await userService.isSystemUserAdmin();
+    if (isSystemAdmin) {
```
mauberti-bc: We shouldn't need any authorization methods here because the middleware is responsible for that. By using the `SystemRole` discriminator with `validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN]`, we already know the caller is an admin passed the middleware.

I suggest removing this method. We're intentionally not authorizing requests via `team` membership right now (until we get to the middleware ticket).

api/src/services/data-request-service.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    const userId = this.connection.systemUserId();
+    const dataRequestTeamMember = await this.findTeamMember(teamId, userId);
+
+    return dataRequestTeamMember !== null;
+  }
+}
+
+/**
+ * Generates a unique team name for auto-created data request teams.
```
mauberti-bc: Create a new util file in `api/src/utils` called `data-request.ts`

api/src/repositories/data-request-repository.ts
```diff
@@ -0,0 +1,285 @@
...
+      }
+      if (filters?.team_id) {
+        query.where('dr.team_id', filters.team_id);
+      }
+
+      const response = await this.connection.knex(query, DataRequest);
+      return response.rows;
+    }
+
+    const query = knex('data_request')
```
mauberti-bc: This method has 2 distinct pathways. It feels odd for the first pathway in `if (filters.status)` to return early. You should isolate the common base used for both pathways and always call that before the conditionals.

This method should be refactored so we always call `knex('data_request').select('data_request_id', team_id', 'reason', 'requested_by').whereNull('record_end_date)`, and then status just becomes a filter in the same way as other filters.

Something like:
```sql
 
const queryBuilder = knex('data_request as dr').select('data_request_id', team_id', 'reason', 'requested_by').whereNull('record_end_date')

if (filters?.status){
    queryBuilder.join('data_request_status as drs', 'drs.data_request_id', 'dr.data_request_id').whereNull('dr.record_end_date').whereNull('drs.record_end_date')
}
```

mauberti-bc (changes requested):

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Please correct me if I'm wrong because I'm a bit confused. The ticket says that 'This will return data if the authenticated user making the request is either a biohub system administrator, OR is a system_user belonging to the team_id that initiated this request.' but your comment says that 'only admins are authorized for data request endpoints'. 

Do you mean:
1. the ticket details are incorrect (SIMSBIOHUB-865)?
2. you mean that for this PR only admins are authorized because the team authorization check will come in a separate PR?

Your answer will help me correct this and make the right choice going forward.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: I mistakenly thought connection.commit wasn’t necessary for read-only queries, but I misunderstood the convention.

mauberti-bc: > I mistakenly thought connection.commit wasn’t necessary for read-only queries, but I misunderstood the convention.

It will work without commit, but committing ensures any locks are released.

api/src/paths/data-request/{dataRequestId}/index.ts
mauberti-bc: Good catch, the change to admins only was based on reviewing the PR and moving middleware changes into a different branch. I should have updated the ticket description to match: **only admins are authorized for data request endpoints** is true. I'll update the ticket.

J-pilon: Ok. I completely understand. Thank you!

J-pilon:
**PR Walkthrough Video - Please watch:** https://www.loom.com/share/2b385ccbbfe24b3eabaadf2be04bcadc

api/src/services/data-request-service.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/commits/6bbe7e8dd7ef7f6848e062d9af71db33f2b53572

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/paths/data-request/index.ts
J-pilon: Resolved.

api/src/paths/data-request/index.ts
J-pilon: Resolved.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Understood. I'll follow the conventions.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved: https://github.com/bcgov/biohubbc-platform/pull/333/commits/86c43aa28e8cce80a0b434bb1cf0243bc0cecff6

When viewing data requests or a single data request, it makes the most sense to also return the data request status details as that determines the state of the data request. So the base query always returns data request status details and then if 'status' is present in the filters object it will build on the base query.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: My reasoning for adding `SystemUser` discriminator:

Without the `SystemUser` discriminator, a malicious actor could potentially:

1. Have a valid Keycloak JWT token (properly authenticated)
2. Have the token claim to have `SYSTEM_ADMIN` role in Keycloak
3. **BUT** not actually exist as a registered user in the application's database

This could happen if:
- A user was deleted from the system database but their Keycloak account still exists
- Someone manipulates Keycloak roles directly without going through proper user registration
- There's a race condition between token issuance and user deletion

Please correct me if I'm wrong about this.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/repositories/data-request-repository.ts
J-pilon: Resolved.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: My reasoning for adding SystemUser discriminator for POST data_request:

Without the SystemUser discriminator, a malicious actor could potentially:

Have a valid Keycloak JWT token (properly authenticated)
Have the token claim to have SYSTEM_ADMIN role in Keycloak
BUT not actually exist as a registered user in the application's database
This could happen if:

A user was deleted from the system database but their Keycloak account still exists
Someone manipulates Keycloak roles directly without going through proper user registration
There's a race condition between token issuance and user deletion
Please correct me if I'm wrong about this.

J-pilon:
**Brief Video Walkthrough**
https://www.loom.com/share/9174f1f0d41a478884fbecc28c6a2796

mauberti-bc (approved):
Looks good!

PR #337, opened 2026-02-20
SIMSBIOHUB-862: Checking out Shopping Cart

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-862](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-862)

## Description of Changes

As a biohub client, I want to 'check out' my populated shopping cart, to indicate that I am ready for package download. 

## Acceptance Criteria 
WHEN I am finished adding items to my cart,  
THEN a "Checkout" button should be available.  
WHEN I click the "Checkout" button,  
THEN a new record should be inserted into the `download` table,  
AND each feature in the shopping cart should be linked to this download record in the `download_feature` table.

WHEN the download record is created,
THEN it should use a default fragment size of 50 MB,
AND the checkout endpoint should accept an optional `fragment_size_bytes` parameter to override the default.

## Testing Notes

Add to the cart and the "Checkout"

Confirm cart API call was successful:

Should see Request URL in browser dev tools:
http://localhost:6100/api/cart/{uuid}/checkout
Request Method
POST
Status Code
201 Created


mauberti-bc (approved):
👍

PR #339, opened 2026-02-25
SIMSBIOHUB 878: Team Middleware for Authorization

J-pilon:
## Links to Jira Tickets
https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-878

## Description of Changes

- Introduces team-scoped authorization: models and a dedicated repository and service (with tests) for resolving team-based access.
- Adds Zod types TeamMemberRecord, TeamPolicyRecord, and DataRequestRecord for query results.
- TeamAuthorizationRepository provides Knex query-builder methods to find active team membership by team ID, by policy ID, and by data request ID, and a SQL (CTE) method to find a team policy for a submission feature via URN matching on policy_statement.
- TeamAuthorizationService exposes isUserAuthorizedForTeamEntity, which handles team, policy, data request, and submission_feature entity types; for submission features it respects the secured flag, optional submission ID, and uses the new repository method to resolve access by submission feature ID.
- AuthorizationService is refactored so team rules are evaluated by delegating to TeamAuthorizationService instead of inlining logic.
- The GET submission feature route (/submission/{submissionId}/features/{submissionFeatureId}) is updated to authorize via the Team rule with entity type submission_feature (submissionFeatureId and submissionId).
- Includes unit tests for TeamAuthorizationRepository, TeamAuthorizationService, and updated AuthorizationService tests that stub TeamAuthorizationService.
- Data requests scoped by user’s teams: GET /data-requests is filtered by the requesting user’s teams. 
- TeamRepository gains a method to get teams by system_user_id (joining team_member and team); TeamService exposes getTeamIdsBySystemUserId; DataRequestRepository adds a team_ids filter; DataRequestService’s findDataRequests uses system_user_id to resolve team IDs and fetch only data requests for those teams.
- Discriminator handling is updated: adds system_user to confirm the user is in the system and removes the unimplemented ticket type. Adds tests for DataRequestService and for DataRequestRepository’s team_ids filter.

## Testing Instructions
Regular System User:
**Create data_request where your part of the team:**
```
curl --location 'http://localhost:6100/api/data-request' \
--header 'Authorization: Bearer <TOKEN> \
--header 'Content-Type: application/json' \
--data '{
    "reason": "Part of this team"
}'
```
**Create data_request where you are not part of the team:**
After creating the data_request, remove yourself from the team.
```
curl --location 'http://localhost:6100/api/data-request' \
--header 'Authorization: Bearer <TOKEN> \
--header 'Content-Type: application/json' \
--data '{
    "reason": "Not part of this team",
    "team_id": "abc123"
}'
```
**Request all data_request records**
```
curl --location 'http://localhost:6100/api/data-request/' \
--header 'Authorization: Bearer <TOKEN>'
```
You should see only one data_request record that you are part of the team. 

**Change system user role to ADMIN**
You should see both data_request records.


api/src/services/authorization/team-authorization-service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        const record = await this.teamAuthorizationRepository.findTeamPolicyBySubmissionFeature(
+          systemUserId,
+          entity.submissionFeatureId
+        );
+        return record !== null;
+      }
+    }
+
+    return false;
```
mauberti-bc: This `return false` is unreachable. Add a default case to the switch that returns false.

api/src/paths/data-request/{dataRequestId}/index.ts
```diff
@@ -11,38 +11,53 @@ import { getLogger } from '../../../utils/logger';
...
-      and: [
+      or: [
         {
           validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN],
           discriminator: 'SystemRole'
+        },
+        {
+          discriminator: 'Team',
+          entity: 'data_request',
+          dataRequestId: req.params.dataRequestId!
```
mauberti-bc: The bang operator (!) is unnecessary here (and in other endpoints that use it within this PR).

api/src/models/data-request.ts
```diff
@@ -14,6 +14,7 @@ export const DataRequestFilters = z.object({
   date_to: z.string().optional(),
   requested_by: z.number().int().optional(),
   team_id: z.string().uuid().optional(),
+  team_ids: z.array(z.string().uuid()).optional(),
```
mauberti-bc: I don't think the array `team_ids` is necessary, so we should drop it. 

In general, we shouldn't have a string or number property and then a separate array version for it. If we support arrays, we should always use arrays - but in this case, keep it simple and just use the singular string.

api/src/paths/data-request/index.ts
```diff
@@ -72,7 +71,17 @@ GET.apiDoc = {
       in: 'query',
       name: 'team_id',
       required: false,
-      schema: { type: 'string', format: 'uuid', description: 'Filter by team ID' }
+      schema: { type: 'string', format: 'uuid', description: 'Filter by a single team ID' }
+    },
+    {
+      in: 'query',
```
mauberti-bc: Remove the array `team_ids`

api/src/paths/data-request/index.ts
```diff
@@ -188,18 +197,31 @@ export function createDataRequest(): RequestHandler {
...
+/** Query params for list endpoint (team_ids may be string or string[] from request). */
+type DataRequestListQuery = Omit<DataRequestFilters, 'team_ids'> & { team_ids?: string | string[] };
+
 /**
  * Parses query params from the request into a filters object for data request list.
  * Returns empty object when no filter params are present.
  */
-function parseQueryParams(req: Request<unknown, unknown, unknown, DataRequestFilters>): DataRequestFilters {
-  const { date_from, date_to, requested_by, team_id, status } = req.query;
+function parseQueryParams(req: Request<unknown, unknown, unknown, DataRequestListQuery>): DataRequestFilters {
```
mauberti-bc: This should be `DataRequestFilters` so the output of the parse utility is exactly the filters object.

api/src/paths/data-request/index.ts
```diff
@@ -188,18 +197,31 @@ export function createDataRequest(): RequestHandler {
...
+  const { date_from, date_to, requested_by, team_id, team_ids, status } = req.query;
   const filters: DataRequestFilters = {
     ...(date_from && { date_from: String(date_from) }),
```
mauberti-bc: The params will be strings or undefined, because the openapi spec requires that, so this can be cleaned up and much less verbose. Look at parseQueryParams in `paths/search/index.ts`

api/src/repositories/authorization/team-authorization-repository.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  async findTeamMembership(systemUserId: number, teamId: string): Promise<TeamMemberRecord | null> {
+    const knex = getKnex();
+    const query = knex
+      .queryBuilder()
+      .select('tm.team_member_id')
+      .from('team_member as tm')
+      .where('tm.team_id', teamId)
+      .where('tm.system_user_id', systemUserId)
+      .whereNull('tm.record_end_date')
```
mauberti-bc: We want to know whether access is denied because the record does not exist or because it's expired, so we should remove the whereNull and put that check in the service layer.

api/src/repositories/authorization/team-repository.ts
```diff
@@ -124,6 +124,27 @@ export class TeamRepository extends BaseRepository {
...
+   * @memberof TeamRepository
+   */
+  async getTeamsBySystemUserId(systemUserId: number): Promise<Team[]> {
+    const knex = getKnex();
+    const query = knex('team_member as tm')
+      .select('t.team_id', 't.name', 't.description')
+      .join('team as t', 't.team_id', 'tm.team_id')
+      .where('tm.system_user_id', systemUserId)
+      .whereNull('tm.record_end_date')
+      .whereNull('t.record_end_date');
```
mauberti-bc: Remove this `whereNull` filters from the SQL and put them in the service layer, to return Access Denied if the record is expired

api/src/services/authorization/authorization-service.ts
```diff
@@ -35,16 +34,32 @@ export interface AuthorizeBySystemUser {
...
+      policyId: string;
+    }
+  | {
+      entity: 'data_request';
+      dataRequestId: string;
+    }
+  | {
+      entity: 'submission_feature';
+      submissionFeatureId: number;
+      submissionId?: number;
```
mauberti-bc: Submission ID should not be optional

api/src/services/authorization/team-authorization-service.ts
```diff
@@ -0,0 +1,76 @@
...
+   * @param {number} systemUserId
+   * @param {TeamAuthorizationEntity} entity
+   * @return {Promise<boolean>}
+   * @memberof TeamAuthorizationService
+   */
+  async isUserAuthorizedForTeamEntity(systemUserId: number, entity: TeamAuthorizationEntity): Promise<boolean> {
+    switch (entity.entity) {
+      case 'team': {
+        const record = await this.teamAuthorizationRepository.findTeamMembership(systemUserId, entity.teamId);
+        return record !== null;
```
mauberti-bc: Each case should check that the record is defined and that record_end_date is null or in the future.

api/src/services/authorization/team-authorization-service.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+
+      case 'policy': {
```
mauberti-bc: We don't need this case because users are not viewing policies, so you can remove it. Also remove the findTeamMembershipByPolicy

api/src/services/access-policy/team-service.ts
```diff
@@ -134,6 +134,18 @@ export class TeamService extends DBService {
...
   }
 
+  /**
+   * Retrieve all active team IDs for a given system user.
+   *
+   * @param {number} systemUserId - The ID of the system user.
+   * @return {Promise<string[]>} - List of team IDs the user belongs to.
+   * @memberof TeamService
+   */
+  async getTeamIdsBySystemUserId(systemUserId: number): Promise<string[]> {
```
mauberti-bc: Missing test coverage for this method

api/src/services/data-request-service.ts
```diff
@@ -58,14 +60,33 @@ export class DataRequestService extends DBService {
...
    * @return {Promise<DataRequestWithStatus[]>}
    * @memberof DataRequestService
    */
   async findDataRequests(filters?: DataRequestFilters): Promise<DataRequestWithStatus[]> {
-    const dataRequests = await this.dataRequestRepository.findDataRequests(filters);
+    const systemUserId = this.connection.systemUserId();
+    const userService = new UserService(this.connection);
+    const user = await userService.getUserById(systemUserId);
+
+    const isSystemAdmin = user.role_names.includes(SYSTEM_ROLE.SYSTEM_ADMIN);
```
mauberti-bc: This method is inconsistent with analogous methods for other services. We should keep this method scoped to returning only the data requests that the user has access to via team membership.
* Move the checks into the findDataRequests repository method's SQL
* Simplify by doing inner joins on `team_member`, `team, `data_request` and return the resulting set of data requests

Don't worry about admins needing to see everything.

mauberti-bc (changes requested):

api/src/models/data-request.ts
J-pilon: Just to clarify, a system_user will only be part of a single team? 
The reason for team_ids is if the user is part of multiple teams. Options:
1. Either we get all the team_ids for a user 
2. We allow the user to choose the team that they want to view data_requests for (but 'team' is not user facing)
3. We get the first team_id (this would be true if users can only be part of one team)

Recommendation:
I'll leave team_ids param and remove 'team_id' since having both would be added complexity.

mauberti-bc: A system user will belong to many teams. I think my last PR comment about pairing back the findDataRequests method and pushing the query logic into the SQL, so we're just doing inner joins to go from system_user_id -> data_requests, gives more context here.

We shouldn't need to find the team_ids that a user has access to as an intermediate step for finding data requests, if I understnad correctly.

mauberti-bc: We do not allow the user to choose the team they want. The user will never be shown the word 'team', we abstract that.

api/src/paths/data-request/{dataRequestId}/index.ts
J-pilon: Resolved.

api/src/services/authorization/team-authorization-service.ts
J-pilon: Resolved.

api/src/services/authorization/authorization-service.ts
J-pilon: Resolved.

api/src/paths/data-request/index.ts
J-pilon: Resolved.

api/src/models/data-request.ts
J-pilon: Resolved.

api/src/paths/data-request/index.ts
J-pilon: Resolved.

api/src/paths/data-request/index.ts
J-pilon: Resolved.

api/src/services/authorization/team-authorization-service.ts
J-pilon: Resolved.

api/src/services/access-policy/team-service.ts
J-pilon: Resolved.

api/src/repositories/authorization/team-authorization-repository.ts
J-pilon: Resolved.

api/src/repositories/authorization/team-repository.ts
J-pilon: Resolved.

api/src/services/authorization/team-authorization-service.ts
J-pilon: Resolved.

api/src/services/data-request-service.ts
J-pilon: Resolved.

api/src/repositories/authorization/team-authorization-repository.ts
J-pilon: I removed the where null check for team_member's record_end_date. Now the method queries the database for data_requests that haven't expired and the service checks that the team membership is active. If the membership is expired it throws a 401 unauthorized. If no data_request record is active then null is returned by the query and the service returns false to the authorization layer.

api/src/services/authorization/team-authorization-service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Check if a user is authorized for a team-scoped entity.
+   *
+   * @param {number} systemUserId
+   * @param {TeamAuthorizationEntity} entity
+   * @return {Promise<boolean>}
+   * @memberof TeamAuthorizationService
+   */
+  async isUserAuthorizedForTeamEntity(systemUserId: number, entity: TeamAuthorizationEntity): Promise<boolean> {
```
mauberti-bc, line 30: The name of this method implies it returns true or false, but then it also throws. Throwing should be the responsiblity of `authorization.ts`.

api/src/services/authorization/team-authorization-service.ts
```diff
@@ -0,0 +1,79 @@
...
+      default:
+        return false;
+    }
+
+    if (!record) {
+      return false;
+    }
+
+    if (record.record_end_date !== null && new Date(record.record_end_date) <= new Date()) {
+      throw new HTTP401('Access Denied');
```
mauberti-bc: Replace throw with return false

mauberti-bc (changes requested):
Looks good, just a couple comments to address

J-pilon:
@mauberti-bc 
- Refactored TeamAuthorizationService's 'isUserAuthorizedForTeamEntity' method to conform to its contract of only returning booleans
- Removed duplication for 'findDataRequests' and 'findDataRequestsByTeamMembership' in DataRequestRepository by creating a private method for applying filters following the pattern in TeamRepository

mauberti-bc (approved):
👍

PR #340, opened 2026-02-25
SIMSBIOHUB-884: Fix Malware Publish Edge Cases

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-884](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-884)[SIMSBIOHUB-884](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-884)

## Description of Changes

The malware scan publish flow has edge cases that pose a small risk of data inconsistency.

**Acceptance Criteria**

1. Throw on Missing Archive Record

WHEN `getUploadArchiveByArtifactId` is called with an `artifact_id` that has no `upload_archive` record,
THEN the repository should throw an `ApiExecuteSQLError` (matching the pattern in `getUploadArchive`),
AND the return type should be `Promise<UploadArchive>` (not `| null`),
AND callers should not need to null-check the result.

2. Update order

WHEN a malware scan completes with CLEAN status,
THEN `handleCleanScanResult` should validate the `upload_archive` record exists before the S3 promotion,
AND if the DB lookup throws, no irreversible S3 changes should be made.

3. Transactional Job Publishing

WHEN `publishMalwareScanJob` publishes a malware scan job,
THEN it should accept an `IDBConnection` parameter and use pg-boss's `db` option to enlist the job insert in the caller's transaction (matching the pattern in `publishProcessDownloadJob`),
AND this prevents ghost jobs and lost jobs.

## Testing Notes

- Verified via unit tests (operation ordering, throw-on-missing, transactional publish)
- **Missing archive now throws** instead of silently returning null

### DIFF
--- a/api/src/queue/publisher.ts
+++ b/api/src/queue/publisher.ts
@@ -181,11 +181,13 @@ export const publishProcessSubmissionFeaturesJob = async (
  *
  * Queues ClamAV scanning for an uploaded artifact.
  *
+ * @param {IDBConnection} connection Database connection for submission validation tracking
  * @param {IMalwareScanJobData} data Job data containing artifactSecurityId
  * @param {IPublishOptions} [options={}] Job options
  * @return {*}  {Promise<PublishJobResult>} Result indicating success, duplicate, or error
  */
 export const publishMalwareScanJob = async (
+  connection: IDBConnection,
   data: IMalwareScanJobData,
   options: IPublishOptions = {}
 ): Promise<PublishJobResult> => {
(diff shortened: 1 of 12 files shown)


api/src/queue/publisher.ts
```diff
@@ -181,11 +181,13 @@ export const publishProcessSubmissionFeaturesJob = async (
  *
  * Queues ClamAV scanning for an uploaded artifact.
  *
+ * @param {IDBConnection} connection Database connection for submission validation tracking
  * @param {IMalwareScanJobData} data Job data containing artifactSecurityId
  * @param {IPublishOptions} [options={}] Job options
  * @return {*}  {Promise<PublishJobResult>} Result indicating success, duplicate, or error
  */
 export const publishMalwareScanJob = async (
+  connection: IDBConnection,
```
mauberti-bc, line 191: What's the value in re-using the same DB connection? It seems like there's value in keeping jobs as independent as possible while still allowing chaining (eg. job A triggers job B), and using the same connection might go against that? On the flip side, I suppose there's also value in re-using connections in case there are many concurrent jobs that exceed the connection limit..

mauberti-bc (approved):
👍

api/src/queue/publisher.ts
dwhogan: The value of putting the job PUBLISHER ( publish  is a record insert into the pg boss table)  on the same transaction as completeArchiveUpload, is no Ghost jobs.  It becomes everything commits or everything is rolled back.  A  job  picked up from said table,  will of course run on a separate POD than the API server, so yes the worker is isolated.

PR #341, opened 2026-02-25
SIMSBIOHUB-883: Parse Feature Content into Search Tables

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-883](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-883)

## Description of Changes

As a BioHub background worker, I want submission features to be automatically indexed into the search tables after validation completes, so that new submissions become searchable without manual admin intervention.

Acceptance Criteria

1. Publishing the Indexing Job
WHEN the `process-submission-features` job handler completes validation successfully,
THEN a new `index-submission-features` job should be published with the `submissionId`,
AND the job should follow existing patterns (singleton key, retry config, DLQ).

2. Processing the Indexing Job
WHEN the `index-submission-features` job is picked up by the worker,
THEN `SearchFeatureService.indexFeaturesBySubmissionId()` should be called,
AND features should be inserted into `search_string`, `search_number`, `search_datetime`, and `search_spatial` tables.

3. Handling Indexing Failures
WHEN the indexing job fails after all retries are exhausted,
THEN the dead letter queue handler should log the failure with error details,
AND the validation job's success should NOT be affected (indexing failure is independent).

## Testing Notes

-  Internal pg-boss worker wiring
- Verified via unit tests 
- **Idempotency:** Re-indexing the same submission does not create duplicate search rows
- Existing search and admin workflows unchanged

### DIFF
--- a/api/src/repositories/search-feature-repository.ts
+++ b/api/src/repositories/search-feature-repository.ts
@@ -32,6 +32,34 @@ const defaultLog = getLogger('repositories/search-feature-repository');
  */
 export class SearchFeatureRepository extends BaseRepository {
   /**
+   * Deletes all existing search records (string, number, datetime, spatial) for features
+   * belonging to the given submission. Used before re-indexing to ensure idempotency —
+   * job retries and manual re-indexing would otherwise accumulate duplicate records
+   * because the search tables have no unique constraint on (submission_feature_id, feature_property_id).
+   *
+   * @param {number} submissionId - The submission whose search records should be cleared
+   * @return {Promise<void>}
+   */
+  async deleteSearchRecordsBySubmissionId(submissionId: number): Promise<void> {
+    defaultLog.debug({ label: 'deleteSearchRecordsBySubmissionId', message: 'start', submissionId });
+
+    const knex = getKnex();
+    const featureIdSubquery = knex
+      .select('submission_feature_id')
+      .from('submission_feature')
+      .where('submission_id', submissionId);
+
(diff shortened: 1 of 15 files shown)


api/src/repositories/search-feature-repository.ts
```diff
@@ -31,6 +31,34 @@ const defaultLog = getLogger('repositories/search-feature-repository');
...
+   * @param {number} submissionId - The submission whose search records should be cleared
+   * @return {Promise<void>}
+   */
+  async deleteSearchRecordsBySubmissionId(submissionId: number): Promise<void> {
+    defaultLog.debug({ label: 'deleteSearchRecordsBySubmissionId', message: 'start', submissionId });
+
+    const knex = getKnex();
+    const featureIdSubquery = knex
+      .select('submission_feature_id')
+      .from('submission_feature')
```
mauberti-bc, line 49: This looks good for this PR. Once we add `upload_id` to submission_feature, we'll replace this whole method with `deleteSearchRecordsByUploadId`.

mauberti-bc (approved):
👍

mauberti-bc:
My same comment form https://github.com/bcgov/biohubbc-platform/pull/340 about re-using the database connection between jobs, but it's non-blocking; more food for thought.

api/src/repositories/search-feature-repository.ts
dwhogan: Right...will set us up for 875, thanks.

PR #348, opened 2026-02-27
SIMSBIOHUB-893: Add Default Pagination & Increase Max Limit

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-893

## Description of Changes

- Changes the `per_page` property in the pagination response to always match the `limit` query param (previously, per_page was the number of rows returned)
- For API endpoints that support pagination, `limit` defaults to 25 and `page` defaults to 1
- Maximum limit of 2000 instead of 100.

### DIFF
--- a/api/src/utils/pagination.ts
+++ b/api/src/utils/pagination.ts
@@ -4,34 +4,36 @@ import { ApiPaginationOptions, ApiPaginationResults } from '../zod-schema/pagina
 export const DEFAULT_PAGINATION_PAGE = 1;
 export const DEFAULT_PAGINATION_LIMIT = 25;
 
-const toNumber = (value: unknown): number | undefined =>
-  typeof value === 'string' || typeof value === 'number' ? Number(value) : undefined;
+const toNumber = (value: unknown): number | undefined => {
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+
+  return typeof value === 'number' ? value : Number(value);
+};
 
 /**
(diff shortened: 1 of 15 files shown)


api/src/utils/pagination.ts
```diff
@@ -1,41 +1,58 @@
...
-    typeof source.order === 'string' && (source.order.toLowerCase() === 'asc' || source.order.toLowerCase() === 'desc')
-      ? (source.order.toLowerCase() as 'asc' | 'desc')
-      : undefined;
-
   const sort = typeof source.sort === 'string' ? source.sort : undefined;
+  const orderRaw = typeof source.order === 'string' ? source.order.toLowerCase() : undefined;
+  const order = orderRaw === 'asc' || orderRaw === 'desc' ? orderRaw : undefined;
 
-  return { page, limit, sort, order };
+  return ensureCompletePaginationOptions({ page, limit, sort, order });
```
dwhogan, line 30: now that `ensureCompletePaginationOptions` has been moved into `makePaginationOptionsFromBody', calling 

' const pagination = makePaginationOptionsFromBody(req);
const [features, count] = await Promise.all([
        service.searchFeatures(filters, ensureCompletePaginationOptions(pagination)' becomes redundant

mauberti-bc: Right, good call.

dwhogan (approved):
👍

PR #349, opened 2026-02-27
SIMSBIOHUB-894: Bug fix to clear invalid cached carts 

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-894

## Description of Changes

- Added stale-cart recovery: when loading a saved cart returns 401/403/404, we clear the cached cartId and reset cart state.
- Kept 5xx behavior safe: we do not clear the cart on server errors.
- Improved add-to-cart resilience: if adding features fails with 401/403/404, we create a new cart and retry once with those features.
- Removed an unnecessary extra request after cart creation: we now use the createCart response directly and skip the immediate follow-up getCartById.
- Tightened cart authorization rules on the API: only active, non-expired carts are accessible.
Added record_end_date to cart models/schemas/interfaces and updated mocks/tests accordingly.
- Cleaned up cart GET pagination: removed manual query mutation and relied on shared default pagination utility.
Expanded tests on both API and app sides to cover these stale-cart, retry, authorization, and pagination behaviors.

## Testing Notes

1) Log in
2) Add a feature to your cart to create a new cart Id
3) Log out
4) Navigate to the search result page -> should see 403 Access Denied on `GET {cartId}`
5) When you add an item to your cart, it should create a new cart

### DIFF
--- a/app/src/contexts/cartContext.test.tsx
+++ b/app/src/contexts/cartContext.test.tsx
@@ -148,11 +148,40 @@ describe('CartContextProvider', () => {
     });
   });
 
+  it.each([401, 403, 404])('clears stale cached cart id when getCartById returns a %s response', async (statusCode) => {
+    setStoredCartId('stale-cart-id');
+    mockCartApi.getCartById.mockRejectedValueOnce(makeApiError(statusCode, 'Cart unavailable'));
+
+    const { getContext } = await renderProvider(getMockAuthState({ base: SystemUserAuthState }));
+
+    await waitFor(() => {
+      expect(mockSetStoredCartId).toHaveBeenCalledWith(null);
+      expect(getContext().features).toEqual([]);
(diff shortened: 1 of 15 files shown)


app/src/contexts/cartContext.test.tsx
```diff
@@ -148,6 +148,36 @@ describe('CartContextProvider', () => {
     });
   });
 
+  it('clears stale cached cart id when getCartById returns a 404', async () => {
```
dwhogan: NIT: add a test for the 404 path

dwhogan (dismissed):

app/src/contexts/cartContext.test.tsx
mauberti-bc: Fixed

dwhogan (approved):
👍

PR #351, opened 2026-03-02
SIMSBIOHUB-875: Add CRUD endpoints for submissions

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-875: CRUD Operations on Submissions](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-875)
- [SIMSBIOHUB-876: Show Publish History in SIMS](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-876)

## Description of Changes

- Added `record_end_date` to `submission_upload` and introduced a `submission_upload_status` table (`submitted`, `approved`, `denied`, `deleted`) to support review workflows and soft-deletes.
- Added submission upload CRUD endpoints:
  - `POST /submission/:submissionId/upload` to create (append) a new upload.
  - `DELETE /submission/:submissionId/upload/:submissionUploadId` to soft-delete uploads only when status is `submitted`.
  - `PATCH /administrative/submission/:submissionId/upload/:submissionUploadId/status` for admins to set status to `approved` or `denied`.
  - `GET /submission/{submissionId}/history` returns publish history from `submission_upload_status` (newest first), with status, submissionUploadId, and createDate.  

## Testing Notes
- Requires [SIMS PR-1613](https://github.com/bcgov/biohubbc/pull/1613)
- Automated: `cd api && npm test` (all tests passing).
- Manual smoke tests against `http://localhost:6100/api`:
  - Verified POST create uploads and `submission_upload_status` rows with `submitted`.
  - Verified DELETE succeeds only for `submitted` uploads and soft-deletes the `submission_upload` row.
  - Verified PATCH as a system admin updates status to `approved`/`denied` and blocks DELETE once reviewed.

### DIFF
--- a/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
+++ b/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
@@ -95,7 +95,7 @@ export function updateSubmissionUploadReviewStatus(): RequestHandler {
       const { status } = req.body;
 
       const submissionUploadService = new SubmissionUploadService(connection);
-      await submissionUploadService.getSubmissionUploadForSubmissionOrThrow(submissionIdParam, submissionUploadId);
+      await submissionUploadService.getSubmissionUploadBySubmissionUuid(submissionIdParam, submissionUploadId);
 
       const reviewStatusService = new SubmissionUploadReviewStatusService(connection);
       const result = await reviewStatusService.updateSubmissionUploadReviewStatus(submissionUploadId, { status });
--- a/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
+++ b/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
@@ -102,7 +102,7 @@ export function updateSubmissionUploadReviewStatus(): RequestHandler {
 
       await connection.commit();
 
-      res.status(200).json(result);
+      return res.status(200).json(result);
     } catch (error) {
       defaultLog.error({
         label: 'updateSubmissionUploadReviewStatus',
--- a/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
+++ b/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
@@ -2,14 +2,12 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { SYSTEM_ROLE } from '../../../../../../../constants/roles';
 import { getDBConnection } from '../../../../../../../database/db';
-import { HTTP404 } from '../../../../../../../errors/http-error';
 import { defaultErrorResponses } from '../../../../../../../openapi/schemas/http-responses';
 import {
   SubmissionUploadReviewStatusResponseSchema,
   UpdateSubmissionUploadReviewStatusRequestSchema
 } from '../../../../../../../openapi/schemas/upload';
 import { authorizeRequestHandler } from '../../../../../../../request-handlers/security/authorization';
-import { SubmissionService } from '../../../../../../../services/submission-service';
 import { SubmissionUploadReviewStatusService } from '../../../../../../../services/upload/submission-upload-review-status-service';
 import { SubmissionUploadService } from '../../../../../../../services/upload/submission-upload-service';
 import { getLogger } from '../../../../../../../utils/logger';
@@ -96,22 +94,8 @@ export function updateSubmissionUploadReviewStatus(): RequestHandler {
       const { submissionId: submissionIdParam, submissionUploadId } = req.params;
       const { status } = req.body;
 
-      const submissionService = new SubmissionService(connection);
-      const byUuid = await submissionService.getSubmissionIdByUUID(submissionIdParam);
-      if (!byUuid) {
-        return res.status(404).json({ message: 'Submission not found.' });
-      }
-
       const submissionUploadService = new SubmissionUploadService(connection);
-      let submissionUpload;
-      try {
-        submissionUpload = await submissionUploadService.getSubmissionUpload(submissionUploadId);
-      } catch {
-        throw new HTTP404('Submission upload not found');
-      }
-      if (submissionUpload.submission_id !== byUuid.submission_id) {
-        throw new HTTP404('Submission upload not found');
-      }
+      await submissionUploadService.getSubmissionUploadForSubmissionOrThrow(submissionIdParam, submissionUploadId);
 
       const reviewStatusService = new SubmissionUploadReviewStatusService(connection);
       const result = await reviewStatusService.updateSubmissionUploadReviewStatus(submissionUploadId, { status });
--- a/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
+++ b/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
@@ -1,0 +1,110 @@
+import { RequestHandler } from 'express';
+import { Operation } from 'express-openapi';
+import { SYSTEM_ROLE } from '../../../../../../../constants/roles';
+import { getDBConnection } from '../../../../../../../database/db';
+import { defaultErrorResponses } from '../../../../../../../openapi/schemas/http-responses';
+import {
+  SubmissionUploadReviewStatusResponseSchema,
+  UpdateSubmissionUploadReviewStatusRequestSchema
+} from '../../../../../../../openapi/schemas/upload';
+import { authorizeRequestHandler } from '../../../../../../../request-handlers/security/authorization';
+import { SubmissionUploadReviewStatusService } from '../../../../../../../services/upload/submission-upload-review-status-service';
+import { getLogger } from '../../../../../../../utils/logger';
+
+const defaultLog = getLogger('paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status');
+
+export const PATCH: Operation = [
+  authorizeRequestHandler(() => ({
+    and: [
+      {
+        validSystemRoles: [SYSTEM_ROLE.SYSTEM_ADMIN],
+        discriminator: 'SystemRole'
+      }
+    ]
+  })),
+  updateSubmissionUploadReviewStatus()
+];
+
+PATCH.apiDoc = {
+  description:
+    'Update the review status of a submission upload to approved or denied. Only system administrators may perform this action.',
+  tags: ['admin'],
+  security: [{ Bearer: [] }],
+  parameters: [
+    {
+      description: 'Submission ID (UUID).',
+      in: 'path',
+      name: 'submissionId',
+      schema: {
+        type: 'string',
+        format: 'uuid'
+      },
+      required: true
+    },
+    {
+      description: 'Submission Upload ID (UUID).',
+      in: 'path',
+      name: 'submissionUploadId',
+      schema: {
+        type: 'string',
+        format: 'uuid'
+      },
+      required: true
+    }
+  ],
+  requestBody: {
+    required: true,
+    content: {
+      'application/json': {
+        schema: UpdateSubmissionUploadReviewStatusRequestSchema
+      }
+    }
+  },
+  responses: {
+    200: {
+      description: 'Submission upload review status updated successfully.',
+      content: {
+        'application/json': {
+          schema: SubmissionUploadReviewStatusResponseSchema
+        }
+      }
+    },
+    ...defaultErrorResponses
+  }
+};
+
+/**
+ * Updates the review status (approved or denied) for a submission upload.
+ * Only system administrators are authorized to call this endpoint.
+ *
+ * @returns {RequestHandler}
+ */
+export function updateSubmissionUploadReviewStatus(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const { submissionUploadId } = req.params;
+      const { status } = req.body;
+
+      const reviewStatusService = new SubmissionUploadReviewStatusService(connection);
+      const result = await reviewStatusService.updateSubmissionUploadReviewStatus(submissionUploadId, { status });
+
+      await connection.commit();
+
+      res.status(200).json(result);
+    } catch (error) {
+      defaultLog.error({
+        label: 'updateSubmissionUploadReviewStatus',
+        message: 'error updating submission upload review status',
+        error
+      });
+      await connection.rollback();
+      throw error;
+    } finally {
+      connection.release();
+    }
+  };
+}
--- a/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
+++ b/api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
@@ -2,13 +2,16 @@ import { RequestHandler } from 'express';
 import { Operation } from 'express-openapi';
 import { SYSTEM_ROLE } from '../../../../../../../constants/roles';
 import { getDBConnection } from '../../../../../../../database/db';
+import { HTTP404 } from '../../../../../../../errors/http-error';
 import { defaultErrorResponses } from '../../../../../../../openapi/schemas/http-responses';
 import {
   SubmissionUploadReviewStatusResponseSchema,
   UpdateSubmissionUploadReviewStatusRequestSchema
 } from '../../../../../../../openapi/schemas/upload';
 import { authorizeRequestHandler } from '../../../../../../../request-handlers/security/authorization';
+import { SubmissionService } from '../../../../../../../services/submission-service';
 import { SubmissionUploadReviewStatusService } from '../../../../../../../services/upload/submission-upload-review-status-service';
+import { SubmissionUploadService } from '../../../../../../../services/upload/submission-upload-service';
 import { getLogger } from '../../../../../../../utils/logger';
 
 const defaultLog = getLogger('paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status');
@@ -69,7 +72,11 @@ PATCH.apiDoc = {
         }
       }
     },
-    ...defaultErrorResponses
+    ...defaultErrorResponses,
+    404: {
+      description:
+        'Submission not found (invalid submissionId) or submission upload not found (submissionUploadId does not belong to this submission).'
+    }
   }
 };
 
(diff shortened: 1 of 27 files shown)


api/src/paths/submission/{submissionId}/history/index.ts
```diff
@@ -0,0 +1,101 @@
...
+export function getSubmissionHistory(): RequestHandler {
+  return async (req, res) => {
+    const connection = getDBConnection(req.keycloak_token);
+
+    try {
+      await connection.open();
+
+      const submissionUuid = req.params.submissionId as string;
+      const reviewStatusService = new SubmissionUploadReviewStatusService(connection);
+      const rows = await reviewStatusService.getSubmissionUploadStatusHistory(submissionUuid);
```
mauberti-bc: Move business logic like this into the service layer (eg. maps, significant data manipulation, etc.)

api/src/repositories/upload/submission-upload-review-status-repository.ts
```diff
@@ -0,0 +1,140 @@
...
+        status
+      FROM
+        submission_upload_status
+      WHERE
+        submission_upload_id = ${submissionUploadId};
+    `;
+
+    const response = await this.connection.sql(sqlStatement, SubmissionUploadReviewStatus);
+
+    if (response.rowCount !== 1) {
```
mauberti-bc, line 77: We just added the `ApiNotFoundError`, which we should use in the `get*` repository mehtods that query a single row by its primary key. This avoids returning a 5xx error if the resource is not found.

Look at `cartRepository.getCartById` for an example

api/src/paths/submission/{submissionId}/history/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      const history = rows.map((row) => ({
+        submissionUploadId: row.submission_upload_id,
+        status: row.status,
+        ...(row.create_date && {
+          createDate: row.create_date instanceof Date ? row.create_date.toISOString() : String(row.create_date)
```
mauberti-bc: Is this necessary? I think we can just return `rows` without manipulation (create_date will never be null because it's non-null in the database)

api/src/paths/submission/{submissionId}/history/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      }));
+
+      let submissionId: number;
+      if (rows.length > 0) {
+        submissionId = rows[0].submission_id;
+      } else {
+        const submissionService = new SubmissionService(connection);
+        const byUuid = await submissionService.getSubmissionIdByUUID(submissionUuid);
+        if (!byUuid) {
+          await connection.rollback();
```
mauberti-bc: throw a 404 error instead of rolling back, since we already roll back/propagate the error in the catch block?

api/src/paths/submission/{submissionId}/upload/index.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      const submissionUuid = req.params.submissionId as string;
+      const { bytes } = req.body;
+
+      const uploadIngestionService = new UploadIngestionService(connection);
+      const result = await uploadIngestionService.startArchiveUploadForExistingSubmissionByUuid(bytes, submissionUuid);
+
+      await connection.commit();
+
+      res.status(201).json(result);
```
mauberti-bc: nit: add `return` to all the `res` calls, for clarity/consistency

api/src/repositories/upload/submission-upload-review-status-repository.ts
```diff
@@ -0,0 +1,140 @@
...
+      INNER JOIN submission s ON s.submission_id = su.submission_id
+      WHERE
+        s.uuid = ${submissionUuid}
+      ORDER BY
+        sus.create_date DESC;
+    `;
+
+    const response = await this.connection.sql(sqlStatement, SubmissionUploadReviewStatusHistoryRow);
+
+    return response.rows ?? [];
```
mauberti-bc: Unnecessary fallback with `?? []`. Just remove that

api/src/services/upload/submission-upload-service.ts
```diff
@@ -34,6 +36,37 @@ export class SubmissionUploadService extends DBService {
...
+   * Retrieves a submission_upload record only if it belongs to the submission identified by the given UUID.
+   * Use this to validate path parameters (submissionId + submissionUploadId) before acting on an upload.
+   *
+   * @param {string} submissionUuid Submission UUID from path (submission.uuid)
+   * @param {string} submissionUploadId Submission upload ID from path
+   * @return {Promise<SubmissionUpload>} The submission upload record
+   * @throws {HTTP404} If the submission does not exist, or the upload does not exist, or the upload does not belong to the submission
+   * @memberof SubmissionUploadService
+   */
+  async getSubmissionUploadForSubmissionOrThrow(
```
mauberti-bc: Avoid naming like `*OrThrow`. Most of our `get*` methods throw if the resource isn't found, so it's implied. Just use do something like `getSubmissionUploadBySubmissionUUID` for consistency with the rest of the codebase.

api/src/services/upload/submission-upload-service.ts
```diff
@@ -34,6 +36,37 @@ export class SubmissionUploadService extends DBService {
...
+    submissionUuid: string,
+    submissionUploadId: string
+  ): Promise<SubmissionUpload> {
+    const submissionService = new SubmissionService(this.connection);
```
mauberti-bc: For this whole method, just push the logic into a repository method called `getSubmissionUploadBySubmissionUUID`? Then the service method just calls that repo method, like the established pattern typical of other services.

`WHERE submission_uuid = $1 AND submission_upload_id = $2` and throw a 404 if rowCount is 0.

api/src/services/upload/upload-ingestion-service.ts
```diff
@@ -47,21 +49,69 @@ export class UploadIngestionService extends DBService {
...
+   * @param {string} submissionUuid - Submission UUID (submission.uuid).
+   * @returns {Promise<PresignedUploadUrlResponse>}
+   * @throws {HTTP404} If no submission exists for the given UUID.
+   */
+  async startArchiveUploadForExistingSubmissionByUuid(
+    bytes: number,
+    submissionUuid: string
+  ): Promise<PresignedUploadUrlResponse> {
+    const byUuid = await this.submissionService.getSubmissionIdByUUID(submissionUuid);
+    if (!byUuid) {
```
mauberti-bc: This PR throws 404s in multiple service methods, so might be worth updating the `getSubmissionIdByUUID` repository method to throw a 404.

api/src/services/upload/upload-ingestion-service.ts
```diff
@@ -47,21 +49,69 @@ export class UploadIngestionService extends DBService {
...
+   *
+   * @param {number} bytes
+   * @param {number} submissionId - Integer PK for DB operations
+   * @param {string} submissionUuid - UUID returned to client as submissionId
+   * @returns {Promise<PresignedUploadUrlResponse>}
+   */
+  async _startArchiveUploadForSubmission(
+    bytes: number,
+    submissionId: number,
+    _submissionUuid: string
```
mauberti-bc: nit: Is the underscore necessary?

api/src/services/upload/upload-ingestion-service.ts
```diff
@@ -93,8 +143,13 @@ export class UploadIngestionService extends DBService {
     // 7. Persist S3 upload ID
     await this.uploadService.updateUpload(upload_id, { s3_upload_id: s3UploadId });
 
+    // 8. Submission UUID must come from submission table only (never upload_id or submission_upload_id)
+    const submissionRow = await this.submissionService.getSubmissionRecordBySubmissionId(submissionId);
+    const submissionIdForResponse = submissionRow.uuid;
+
     return {
-      submissionId: submission_id,
+      submissionId: submissionIdForResponse,
```
mauberti-bc: This will get confusing, since we're setting the UUID as the Submission ID, and one might expect the submission ID to always be the integer PK. I think we should always pass both for clarity.

mauberti-bc (changes requested):

api/src/paths/submission/{submissionId}/history/index.ts
dylanrogowsky-oxd: - `api/src/services/upload/submission-upload-review-status-service.ts`
    - New method `getSubmissionHistoryByUuid(submissionUuid)` encapsulates:
      - Fetching rows via `getStatusHistoryBySubmissionUuid`.
      - Resolving `submissionId` (from rows or via `SubmissionService.getSubmissionIdByUUID`).
      - Mapping to `{ submissionUploadId, status, createDate }` with `createDate` as an ISO string.
  - `api/src/paths/submission/{submissionId}/history/index.ts`
    - Handler now calls `getSubmissionHistoryByUuid` and returns `res.status(200).json(result)`, with no mapping or submission lookup inline.

api/src/repositories/upload/submission-upload-review-status-repository.ts
dylanrogowsky-oxd: - `api/src/repositories/upload/submission-upload-review-status-repository.ts`
    - Imports `ApiNotFoundError`.
    - `getSubmissionUploadReviewStatus` now:
      - Throws `ApiNotFoundError('Submission upload status not found', [...])` when `rowCount === 0`.
      - Throws `ApiExecuteSQLError('Unexpected row count', [...])` only when `rowCount !== 1`.
  - Callers rely on the global error-handler’s mapping of `ApiNotFoundError` to HTTP 404 (no ad‑hoc message matching).

api/src/paths/submission/{submissionId}/history/index.ts
dylanrogowsky-oxd: - `api/src/services/upload/submission-upload-review-status-service.ts`
    - In `getSubmissionHistoryByUuid`, mapping always sets `createDate`:
      - `createDate: row.create_date instanceof Date ? row.create_date.toISOString() : String(row.create_date)`.
    - The mapping is kept (to match the OpenAPI response schema and rename fields), but the conditional spread around `create_date` is removed and the logic moved out of the path layer.

api/src/paths/submission/{submissionId}/history/index.ts
dylanrogowsky-oxd: - `api/src/paths/submission/{submissionId}/history/index.ts`
    - No longer performs manual `rollback` + `return res.status(404)...` in the "no rows" case.
    - Delegates to `getSubmissionHistoryByUuid`, which calls `SubmissionService.getSubmissionIdByUUID` and relies on the repository’s `ApiNotFoundError` when the submission is missing; the path’s `catch` now handles rollback in a single place.

api/src/paths/submission/{submissionId}/upload/index.ts
dylanrogowsky-oxd: - `api/src/paths/submission/{submissionId}/upload/index.ts`
    - Handler now ends with `return res.status(201).json(result);` instead of just `res.status(201).json(result);`.

api/src/repositories/upload/submission-upload-review-status-repository.ts
dylanrogowsky-oxd: - `api/src/repositories/upload/submission-upload-review-status-repository.ts`
    - `getStatusHistoryBySubmissionUuid` now returns `response.rows` directly instead of `response.rows ?? []`.

api/src/services/upload/submission-upload-service.ts
dylanrogowsky-oxd: - `api/src/repositories/upload/submission-upload-repository.ts`
    - New method `getSubmissionUploadBySubmissionUuid(submissionUuid, submissionUploadId)`:
      - JOINs `submission_upload` with `submission` on `submission_id`.
      - Filters by `s.uuid = submissionUuid`, `su.submission_upload_id = submissionUploadId`, and `su.record_end_date IS NULL`.
      - Throws `ApiNotFoundError('Submission upload not found', [...])` when `rowCount === 0`.
      - Throws `ApiExecuteSQLError('Unexpected row count', [...])` when `rowCount !== 1`.
  - `api/src/services/upload/submission-upload-service.ts`
    - Replaced `getSubmissionUploadForSubmissionOrThrow` with `getSubmissionUploadBySubmissionUuid(submissionUuid, submissionUploadId)` that simply delegates to the repository method (no direct use of `SubmissionService` or `HTTP404`).
  - Path handlers:
    - `api/src/paths/submission/{submissionId}/upload/{submissionUploadId}/index.ts` (DELETE)
    - `api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts` (PATCH)
    - Both now call `submissionUploadService.getSubmissionUploadBySubmissionUuid(submissionIdParam, submissionUploadId)` for ownership validation.

api/src/services/upload/submission-upload-service.ts
dylanrogowsky-oxd: - `api/src/repositories/submission-repository.ts`
    - `getSubmissionIdByUUID(uuid)` now:
      - Returns `Promise<{ submission_id: number }>` (no `null`).
      - Throws `ApiNotFoundError('Submission not found', ['SubmissionRepository->getSubmissionIdByUUID', { uuid }])` when `rowCount === 0`.
  - `api/src/services/submission-service.ts`
    - `getSubmissionIdByUUID` now returns the non-null repository result and no longer exposes `null`.

api/src/services/upload/upload-ingestion-service.ts
dylanrogowsky-oxd: - `api/src/services/upload/upload-ingestion-service.ts`
    - `_startArchiveUploadForSubmission changed from `(..., _submissionUuid: string)` to `(..., submissionUuid: string)`.
    - The parameter is now used explicitly when building the response (based on the next feedback item), so the underscore prefix (unused) is no longer appropriate.

api/src/services/upload/upload-ingestion-service.ts
dylanrogowsky-oxd: - `api/src/services/upload/upload-ingestion-service.ts`
    - `_startArchiveUploadForSubmission(bytes, submissionId: number, submissionUuid: string)` now takes both identifiers.
    - The response uses:
      - `submissionId: submissionUuid` (explicitly documented in the JSDoc as "submission UUID; used when building response (API returns this as submissionId for client use)").
      - `submissionId` (integer PK) continues to be used for DB operations only.
    - Callers:
      - `startArchiveUpload` passes `submissionRecord.uuid` as `submissionUuid`.
      - `startArchiveUploadForExistingSubmissionByUuid` calls `getSubmissionIdByUUID` then passes the same `submissionUuid` string and the resolved integer `submission_id` into `_startArchiveUploadForSubmission`.

api/src/services/upload/upload-ingestion-service.test.ts
```diff
@@ -57,10 +58,18 @@ describe('UploadIngestionService', () => {
...
+        .resolves({ uuid: mockSubmission.uuid } as any);
       sinon.stub(UploadService.prototype, 'insertUpload').resolves({ upload_id: mockUploadId });
       sinon
         .stub(SubmissionUploadService.prototype, 'insertSubmissionUpload')
         .resolves({ submission_upload_id: 'submission-upload-id-1' });
+      sinon.stub(SubmissionUploadReviewStatusService.prototype, 'insertSubmissionUploadReviewStatus').resolves({
+        submission_upload_status_id: 1,
+        submission_upload_id: 'submission-upload-id-1',
+        status: 'submitted'
+      } as any);
```
mauberti-bc: Avoid use of `any`. We should include enough properties to match the expected return type.

api/src/services/upload/upload-ingestion-service.test.ts
```diff
@@ -117,10 +129,18 @@ describe('UploadIngestionService', () => {
 
     it('should throw if artifact creation fails', async () => {
       sinon.stub(SubmissionService.prototype, 'insertSubmissionRecord').resolves({ submission_id: 123 });
+      sinon
+        .stub(SubmissionService.prototype, 'getSubmissionRecordBySubmissionId')
+        .resolves({ uuid: mockSubmission.uuid } as any);
```
mauberti-bc: Same comment about avoiding 'any' applies to all test additions

api/src/paths/submission/{submissionId}/upload/{submissionUploadId}/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        );
+      }
+
+      await submissionUploadService.softDeleteSubmissionUpload(submissionUploadId);
+
+      await reviewStatusService.updateSubmissionUploadReviewStatus(submissionUploadId, { status: 'deleted' });
+
+      await connection.commit();
+
+      res.status(204).send();
```
mauberti-bc: `return` missing

api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+      const submissionUploadService = new SubmissionUploadService(connection);
+      await submissionUploadService.getSubmissionUploadBySubmissionUuid(submissionIdParam, submissionUploadId);
+
+      const reviewStatusService = new SubmissionUploadReviewStatusService(connection);
+      const result = await reviewStatusService.updateSubmissionUploadReviewStatus(submissionUploadId, { status });
+
+      await connection.commit();
+
+      res.status(200).json(result);
```
mauberti-bc: `return` missing

mauberti-bc (changes requested):

api/src/paths/submission/{submissionId}/upload/{submissionUploadId}/index.ts
dylanrogowsky-oxd: Added `return`

api/src/paths/administrative/submission/{submissionId}/upload/{submissionUploadId}/status/index.ts
dylanrogowsky-oxd: Added `return`

api/src/services/upload/upload-ingestion-service.test.ts
dylanrogowsky-oxd: Removed `any` in this file.

api/src/services/upload/upload-ingestion-service.test.ts
dylanrogowsky-oxd: Removed `any` in this file.

mauberti-bc (approved):
👍

PR #352, opened 2026-03-03
SIMSBIOHUB-882: Refactor Ingestion Job to support multiple uploads

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-882](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-882)

## Description of Changes

Update the ingestion jobs that currently take submission_id so they can take submission_upload_id instead.

This makes ingestion upload-scoped, so the same submission_id can be re-published by creating a new upload and running ingestion again without colliding with (or overwriting) prior runs.
Job logging and validation/processing status should be attributable to the specific upload being processed, not the parent submission.

## Testing Notes

Run `make test`

### DIFF
--- a/api/src/models/submission-upload.ts
+++ b/api/src/models/submission-upload.ts
@@ -34,11 +34,11 @@ export interface SubmissionUploadFilters {
 }
 
 /**
- * The upload + submission identifier pair used across the ingestion pipeline.
- * Both IDs travel together through publisher, job handler, validation tracking,
- * and feature insertion. The trigger resolves both before publishing.
+ * Job payload for the ingestion pipeline. Single identifier eliminates the risk
+ * of submissionId/uploadId getting out of sync — each consumer resolves what
+ * it needs from the submission_upload bridge table.
  */
 export interface IngestionJobData {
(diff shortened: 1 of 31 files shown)


api/src/models/submission-upload.ts
```diff
@@ -32,3 +32,13 @@ export type UpdateSubmissionUpload = z.infer<typeof UpdateSubmissionUpload>;
 export interface SubmissionUploadFilters {
   role?: UploadArtifactRoleEnum;
 }
+
+/**
+ * The upload + submission identifier pair used across the ingestion pipeline.
+ * Both IDs travel together through publisher, job handler, validation tracking,
+ * and feature insertion. The trigger resolves both before publishing.
+ */
+export interface IngestionJobData {
```
mauberti-bc: Can we simplify by just passing `submission_upload_id`?

mauberti-bc (approved):
👍

api/src/models/submission-upload.ts
dwhogan: done

PR #353, opened 2026-03-03
SIMSBIOHUB 866: Submitting a Data Request

J-pilon:
## Links to Jira Tickets
https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-866

## Description of Changes

- **API – Data request creation and policy:** `DataRequestService.createDataRequest` creates a data request and, when no `team_id` is provided, creates a team and adds the requester as a member. It then creates a wildcard access policy (expiring in 30 days) for that data request and links the team to the policy via `record_end_date`; policy creation uses `_getDataRequestPolicyExpiryDate()` so policies support an expiration date. New data requests are auto-approved (APPROVED status).
- **API – Data request helpers:** Adds `api/src/utils/data-request.ts` with `_generateDataRequestTeamName`, `_generateDataRequestPolicyName`, `_getDataRequestPolicyExpiryDate`, and `_transformFlatDataRequestToNested`. `DataRequestService` is refactored into private methods: `createAndReturnTeamForDataRequest`, `createPolicyForDataRequest`, and `linkTeamToPolicy`.
- **API – Policy expiry:** Policy creation for data requests passes `record_end_date` when creating a policy with an expiration date.
- **API – Tests:** Unit tests added for `DataRequestService` (data-request-service.test.ts).
- **App – Request access flow:** Adds client route `/search/request-access` (in `SearchRouter`) and `RequestAccessPage` with a Formik form (reason field, validation, submit). Success shows a message and a “Back to search results” link; search params are preserved for return navigation.
- **App – API integration:** Adds `useDataRequestApi` hook with `createDataRequest` calling `POST /api/data-request`; `RequestAccessPage` uses it and shows errors via `dialogContext.setErrorDialog`.
- **App – Secure-results banner:** Adds `hasSecureResults` in `utils/search-result-options.ts` to detect if any search result has `is_secured`. `SearchResultOptions` shows an `AlertBanner` when any results are secured, with a “Request Access” button that navigates to `/search/request-access` with current search params.
- **App – Reusable UI:** Extracts `AlertBanner` (MUI `Alert` with default `severity="info"`) in `components/notifications/AlertBanner.tsx` for reuse. Error handling on the request-access flow uses `dialogContext` to display errors in a dialog.

## Testing Instructions

**Submit a data request from the UI (with secured results):**
1. Run a search that returns at least one secured result (`is_secured: true`). **Remember to create a SubmissionFeatureSecurity record first**

Example search: 
- search string: custodia clibanus tot
- feature type: dataset

3. On the search results page, confirm the banner appears: “Some records in these results are secured. You can request access…”
4. Click **Request Access**; you should land on `/search/request-access` with search params preserved.
5. Enter a reason (required, max 2000 characters) and click **Submit Request**.
6. Confirm a success message and “Back to search results” link; following it returns to the same search list.

**Error handling:** Submit the form with an invalid or missing reason, or trigger a network/API error; confirm the error is shown in the dialog (via `dialogContext`), not only in the console.


J-pilon:
@mauberti-bc I'm not sure how to handle preventing duplicate data requests:
```
DataRequestService.createDataRequest still creates a new team, policy, and approved request on every call with no check for an existing active request by the same user. The frontend also doesn't hide/disable the banner if the user already has an approved request.
```

I'm thinking the best way to do this check is by policy statement URNs but that is out of scope for this PR. 

Should this be out of scope for this PR or is there another way that I'm not thinking of?

api/src/services/data-request-service.ts
```diff
@@ -87,41 +96,42 @@ export class DataRequestService extends DBService {
...
   async createDataRequest(requestedBy: number, payload: CreateDataRequest): Promise<DataRequestWithStatus> {
-    let resolvedTeamId = payload.team_id;
-    if (resolvedTeamId === undefined) {
-      const teamService = new TeamService(this.connection);
-      const team = await teamService.createTeam({ name: _generateDataRequestTeamName() });
+    let resolvedTeamId: string;
+    if (payload.team_id) {
+      resolvedTeamId = payload.team_id;
+    } else {
+      const team = await this.createAndReturnTeamForDataRequest(requestedBy);
```
mauberti-bc: For naming conventions, just use `createDataRequest`. The method names should be a simple verb (get, delete, update, sometimes list) followed by the name of the entity.

api/src/services/data-request-service.ts
```diff
@@ -87,41 +96,42 @@ export class DataRequestService extends DBService {
...
+   *
+   * Creates a team for the requester, a wildcard access policy expiring in 30 days
+   * linked to the team, and auto-approves the request.
    *
    * @param {number} requestedBy - system user id
    * @param {CreateDataRequest} payload
    * @return {Promise<DataRequestWithStatus>}
    * @memberof DataRequestService
    */
   async createDataRequest(requestedBy: number, payload: CreateDataRequest): Promise<DataRequestWithStatus> {
```
mauberti-bc: Move `requestedBy` into the `CreateDataRequest` interface as `requested_by`, since they're always sent together.

api/src/services/data-request-service.ts
```diff
@@ -87,41 +96,42 @@ export class DataRequestService extends DBService {
...
-    let resolvedTeamId = payload.team_id;
-    if (resolvedTeamId === undefined) {
-      const teamService = new TeamService(this.connection);
-      const team = await teamService.createTeam({ name: _generateDataRequestTeamName() });
+    let resolvedTeamId: string;
```
mauberti-bc: Nit: make this more readable by doing:
```
let teamId = payload.team_id

if (!teamId){
    const team = await this.createAndReturnTeamForDataRequest(requestedBy);
    teamId = team.teamId
}
```

api/src/services/data-request-service.ts
```diff
@@ -158,4 +168,52 @@ export class DataRequestService extends DBService {
...
+
+  /**
+   * Links a team to a policy by creating a team policy record.
+   *
+   * @param {string} teamId
+   * @param {string} policyId
+   * @return {Promise<void>}
+   * @private
+   */
+  private async linkTeamToPolicy(teamId: string, policyId: string): Promise<void> {
```
mauberti-bc: Replace the 2 args with a single interface, like: `linkTeamToPolicy(payload: {teamId: string, policyId: string})`, to prevent passing the wrong ids.

app/src/features/search/request-access/RequestAccessPage.tsx
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+
+  return (
+    <Container maxWidth="sm" sx={{ py: 6 }}>
+      <Paper sx={{ p: 4 }}>
+        <Typography variant="h2" sx={{ mb: 1 }}>
+          Request Access
+        </Typography>
+        <Typography variant="body1" color="text.secondary" sx={{ mb: 3 }}>
```
mauberti-bc: Nit: don't pass props that are the default, like `variant="body1"`, to keep code minimal.

app/src/hooks/api/useDataRequestApi.ts
```diff
@@ -0,0 +1,44 @@
+import { AxiosInstance } from 'axios';
+
+export interface CreateDataRequestPayload {
```
mauberti-bc: Move interfaces and types to `useDataRequestApi.interface.ts`, like the other api hooks.

app/src/hooks/api/useDataRequestApi.ts
```diff
@@ -0,0 +1,44 @@
...
+  reason: string;
+  team_id?: string;
```
mauberti-bc: Team ID should never be given from the frontend because the backend will create a new team for each data request, so we should just remove it from the `CreateDataRequestPayload` interface to avoid confusion.

app/src/router/search/SearchRouter.tsx
```diff
@@ -36,6 +37,16 @@ export const SearchRouter = () => {
         }
       />
 
+      <Route
+        path="/request-access"
```
mauberti-bc: Usually `request-access` implies system-level access, so we should rename this to `path="/data-request"`.

app/src/router/search/SearchRouter.tsx
```diff
@@ -36,6 +37,16 @@ export const SearchRouter = () => {
...
+        element={
+          <BaseLayout>
+            <PageTitle title="Request Access" description="Request access to secured data" />
```
mauberti-bc: Rename title to Data Request

app/src/router/search/SearchRouter.tsx
```diff
@@ -36,6 +37,16 @@ export const SearchRouter = () => {
         }
       />
 
+      <Route
```
mauberti-bc: Create a new router called DataRequestRouter, since this will probably bloat the SearchRouter over time.

app/src/utils/search-result-options.ts
```diff
@@ -0,0 +1,5 @@
+import { SearchFeatureResultWithRelevancy } from 'interfaces/useSearchApi.interface';
+
+export const hasSecureResults = (rows: SearchFeatureResultWithRelevancy[]): boolean => {
```
mauberti-bc: Nit: utils that are this small, very specific to one feature, and used only once should just be put in the React component where the util is used.

I would delete this util and just put this in the SearchResultOptions:
```
  // Check if any of the results are secured to conditionally show the alert banner
  const hasSecuredResults = useMemo(() => rows.some((row) => row.is_secured), [rows]);
```

mauberti-bc (changes requested):

app/src/hooks/api/useDataRequestApi.ts
J-pilon: @mauberti-bc Can 1 team have any data requests or 1 team for every data request (1 to 1 relationship). 

The reason I'm asking is that 'createDataRequest' method in DataRequestService accepts a team_id which would be passed in from the frontend when this route is called (api/src/paths/data-request/{dataRequestId}/index.ts). 

If we decided that data request to teams is 1:1, I could simplify the route handler and service method.

mauberti-bc (approved):
👍

PR #355, opened 2026-03-05
SIMSBIOHUB-833: Bulk Download Based on Query

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-833](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-833)

## Description of Changes

As a BioHub client, I want to download all features that match a search query, so that I can efficiently retrieve large sets of data without manually adding each feature to a shopping cart.

Acceptance Criteria
1. Performing a Search for Bulk Download  
WHEN a client enters search criteria (keywords, filters, etc.) and executes a search,  
  THEN the system should return all matching features and display the results.
2. Initiating Bulk Download  
WHEN the client clicks the "Download All" button,  
  THEN the current search query parameters should be sent to the server for processing.
3. Creating a Bulk Download Record  
WHEN the bulk download request is received by the server,  
  THEN a new record should be inserted into the `download` table,  
  AND all features matching the search query should be linked to this download record in the `download_feature` table.
4. Notes  
This bulk download bypasses the shopping cart and uses the search query directly to generate the download.  
The `download_feature` table should track the `feature_id` for each feature included in the bulk download.  
Search query parameters (filters, keywords) should be stored with the download record for traceability.

## Testing Notes

Navigate to the search page and click the "Download All" button, a download record should be created.

Note: the notification message is not precise for anonymous users, this we be improved in ticket 824.

### DIFF
--- a/api/src/paths/download/index.ts
+++ b/api/src/paths/download/index.ts
@@ -201,14 +201,12 @@ export function createDownload(): RequestHandler {
       // Create download with search filters stored for traceability.
       // teamId is null: downloads from search are user-owned, not team-based.
       const pipelineService = new DownloadPipelineService(connection);
-      const downloadId = await pipelineService.createDownloadRequest(
+      const downloadId = await pipelineService.createDownloadRequest({
         systemUserId,
-        null,
+        teamId: null,
         submissionFeatureIds,
-        undefined,
-        undefined,
-        filters
-      );
+        searchFilters: filters
+      });
 
       // Publish async processing job within the transaction
       await publisher.publishProcessDownloadJob(connection, { downloadId: downloadId.download_id });
--- a/api/src/paths/download/index.ts
+++ b/api/src/paths/download/index.ts
@@ -194,10 +194,15 @@ export function createDownload(): RequestHandler {
         throw new HTTP400('No features match the filter criteria');
       }
 
+      // Anonymous downloads: system_user_id is null (UUID is the credential).
+      // Matches the cart checkout pattern (cart/{cartId}/checkout/index.ts:98).
+      const systemUserId = isAuthenticated ? connection.systemUserId() : null;
+
       // Create download with search filters stored for traceability.
       // teamId is null: downloads from search are user-owned, not team-based.
       const pipelineService = new DownloadPipelineService(connection);
       const downloadId = await pipelineService.createDownloadRequest(
+        systemUserId,
         null,
         submissionFeatureIds,
         undefined,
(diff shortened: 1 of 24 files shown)


api/src/paths/download/index.ts
```diff
@@ -98,3 +103,122 @@ export function getDownloads(): RequestHandler {
...
+ */
+export function createDownload(): RequestHandler {
+  return async (req, res) => {
+    const isAuthenticated = !!req.keycloak_token;
+    const connection = isAuthenticated ? getDBConnection(req.keycloak_token) : getAPIUserDBConnection();
+
+    try {
+      await connection.open();
+
+      const filters: ISearchFeaturesFilters = req.body.filters;
```
mauberti-bc, line 185: nit: use the same property names between the openAPI schema and types. Pick either `filters` or `searchFilters`, but not both

api/src/paths/download/index.ts
```diff
@@ -98,3 +103,122 @@ export function getDownloads(): RequestHandler {
...
+
+      // Resolve filters to feature IDs via CTE intersection
+      const searchFeatureService = new SearchFeatureService(connection);
+      const submissionFeatureIds = await searchFeatureService.getSearchFeatureIds(filters);
```
mauberti-bc: Note: Out of scope for this PR but we probably want to support pagination here, to reduce memory issue? The requestor could give pagination params (ie. create a download package for page 1 of the query results), or we could loop through all pages internally.

api/src/repositories/search-feature-repository.ts
```diff
@@ -189,6 +189,30 @@ export class SearchFeatureRepository extends BaseRepository {
...
+    if (!filters || Object.keys(filters).length === 0) {
+      return [];
+    }
+
+    const knex = getKnex();
+    const query = this.buildSearchQuery(knex, filters);
+    const idsQuery = knex.from(query.as('sf_filtered')).select('submission_feature_id');
+    const response = await this.connection.knex(idsQuery);
+
+    return response.rows.map((row: { submission_feature_id: number }) => row.submission_feature_id);
```
mauberti-bc: nit: move map call to the service

mauberti-bc (changes requested):

api/src/repositories/search-feature-repository.ts
dwhogan: transform moved to the service

api/src/paths/download/index.ts
dwhogan: Right, as discussed we need a guard for MAX request size, and optimization for the download create.

One is temped to skip the download feature inserts for bulk download requests.  Instead use the `filters' in package pipeline.

For now I switched to `searchFeatureService.getSearchFeaturesCount`

api/src/paths/download/index.ts
dwhogan: renamed db column name to `filters`

mauberti-bc (approved):
👍

PR #357, opened 2026-03-06
SIMSBIOHUB 836: Public Feature Page

J-pilon:
## Links to Jira Tickets
https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-836

## Description of Changes

- **API – Submission feature by ID:** `GET /submission/:submissionId/features/:submissionFeatureId` now returns the submission feature plus `relatedFeatures`. The repository query for the feature includes `submission_name` and `feature_type_display_name` for breadcrumbs and display.
- **API – Related submission features:** Adds `RelatedSubmissionFeature` type and repository method `getRelatedSubmissionFeatures(submissionFeatureId)` (queries `submission_feature_feature` for source/target links). `SubmissionService.getRelatedSubmissionFeatures` delegates to the repository. The route fetches feature and related features in parallel and returns `{ feature, relatedFeatures }`.
- **API – OpenAPI & schema:** Adds `relatedFeatures` to the response schema (Zod) and OpenAPI spec for the submission feature endpoint.
- **API – Tests:** Route test updated to stub and assert `getRelatedSubmissionFeatures`; unit test added for `AuthorizationService`.
- **App – Submission feature page:** `SubmissionFeaturePage` is wrapped with `CartContextProvider` in `SearchRouter` so the page can add/remove the current feature from the cart. Page refactored into sections: **Properties** (`SubmissionFeatureProperties` – table of feature `data`, excluding geometry) and **Related** (`SubmissionFeatureRelated` – list of related features with links).
- **App – Related features UI:** Adds `IRelatedSubmissionFeature` and uses `relatedFeatures` from the feature API. `SubmissionFeatureRelated` shows related features as links to `/search/:submissionId/feature/:submissionFeatureId`; clicking a related feature navigates and the data loader refreshes (via `useEffect` on `submissionId`/`submissionFeatureId`) so the page shows the selected feature.

## Testing Instructions

**View a submission feature and related features:**
1. Go to a submission feature page (e.g. from search: `/search/:submissionId/feature/:submissionFeatureId`).
2. Confirm the page shows Properties (table) and Related (list), breadcrumbs (Search → submission name → feature type), and Add to Cart / In Cart.
3. If there are related features (from `submission_feature_feature`), click one; the URL should update and the page should reload to show that feature’s details and its related list.
4. Add the current feature to the cart and confirm the button switches to “In Cart”; remove and confirm it switches back.

**API:** Run the submission feature route test and the `AuthorizationService` test; both should pass.


app/src/features/search/result/content/option/SearchResultOptions.tsx
```diff
@@ -19,9 +19,10 @@ interface SearchResultOptionsProps {
   isLoading: boolean;
   view: SEARCH_RESULT_OPTION_VIEW;
   onDownload?: (result: SearchFeatureResultWithRelevancy) => void;
+  onClick?: (result: SearchFeatureResultWithRelevancy) => void;
```
mauberti-bc: nit: make onClick required, since we always want a click handler

app/src/features/search/result/layout/list/card/SearchResultCard.tsx
```diff
@@ -1,68 +1,86 @@
...
+  Chip,
+  IconButton,
+  Typography
+} from '@mui/material';
 import { SearchFeatureResultWithRelevancy } from 'interfaces/useSearchApi.interface';
 
 interface SearchResultCardProps {
   result: SearchFeatureResultWithRelevancy;
   isInCart: boolean;
+  onClick?: (result: SearchFeatureResultWithRelevancy) => void;
```
mauberti-bc: Make required

app/src/features/submissions/page/features/components/SubmissionFeatureProperties.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  return (
+    <Paper elevation={0} sx={{ border: '1px solid', borderColor: 'divider' }}>
+      <Box p={3}>
+        <Typography variant="h2" component="h2" mb={2}>
+          Properties
+        </Typography>
+        <LoadingGuard isLoading={isLoading} isLoadingFallback={<SkeletonList numberOfLines={3} />}>
+          <TableContainer>
```
mauberti-bc: Avoid MUI's table and use the `CustomDataGrid` component instead for tabular data.

app/src/features/submissions/page/features/components/SubmissionFeatureRelated.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Typography from '@mui/material/Typography';
+import { LoadingGuard } from 'components/loading/LoadingGuard';
+import { SkeletonList } from 'components/loading/SkeletonLoaders';
+import { IRelatedSubmissionFeature } from 'interfaces/useFeaturesApi.interface';
+import { ReactNode } from 'react';
+import { Link as RouterLink } from 'react-router-dom';
+
+interface SubmissionFeatureRelatedProps {
+  submissionId: number;
+  relatedFeatures: IRelatedSubmissionFeature[] | undefined;
```
mauberti-bc: relatedFeatures should not be optional. If a property is expected to be an array, especially when the API always returns it, it should be defined as a required array rather than potentially undefined.

app/src/features/submissions/page/features/components/SubmissionFeatureRelated.tsx
```diff
@@ -0,0 +1,53 @@
...
+  isLoading?: boolean;
+}
+
+export const SubmissionFeatureRelated = ({
+  submissionId,
+  relatedFeatures = [],
+  isLoading
+}: SubmissionFeatureRelatedProps) => {
+  const formatRelatedFeatures = (): ReactNode => {
```
mauberti-bc: nit: just inline this within the return html. Not much benefit to creating formatRelatedFeatures as its own function.

app/src/router/search/SearchRouter.tsx
```diff
@@ -41,8 +41,10 @@ export const SearchRouter = () => {
         path="/:submissionId/feature/:submissionFeatureId"
         element={
           <BaseLayout>
-            <PageTitle title="Submission Details" description="Details of a specific submission" />
-            <SubmissionFeaturePage />
+            <CartContextProvider>
```
mauberti-bc: Move the provider higher in the router tree so it's shared by the routes that need it. Right now, if you navigate the from the `search/list` page to the `search/:submissionId/feature/:submissionFeatureId` page, you would be recreating the cart context.

app/src/features/submissions/page/features/components/SubmissionFeatureProperties.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { SkeletonList } from 'components/loading/SkeletonLoaders';
+import { ReactNode } from 'react';
+
+interface SubmissionFeaturePropertiesProps {
+  data: Record<string, any>;
+  isLoading?: boolean;
+}
+
+export const SubmissionFeatureProperties = ({ data, isLoading }: SubmissionFeaturePropertiesProps) => {
+  const formatPropertyValue = (value: unknown): string => {
```
mauberti-bc: We have some utils that you can use: ie. `jsonStringifyObjectProperties`

mauberti-bc (changes requested):

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,123 @@
...
-              <pre>{JSON.stringify(featureDataLoader.data, null, 2)}</pre>
-            ) : (
-              'Loading submission feature...'
-            )}
-          </Box>
-        </Paper>
-      </Box>
-    </Container>
+    <>
+      {feature && (
```
mauberti-bc: Wrap the page in a `LoadingGuard` and use the `isLoading` prop with a skeleton for the fallback - search for other examples in the app.

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,123 @@
...
+import { SubmissionFeatureRelated } from './components/SubmissionFeatureRelated';
+
 export const SubmissionFeaturePage = () => {
   const biohubApi = useApi();
+  const { features: cartFeatures, addToCart, removeFromCart } = useCartContext();
 
-  // Get submission ID and submission feature ID from the URL
   const { submissionId, submissionFeatureId } = useParams<{ submissionId: string; submissionFeatureId: string }>();
 
   const featureDataLoader = useDataLoader(() =>
```
mauberti-bc: We should catch and redirect on 401/403 Access Denied errors. Currently, if the request is denied, the page just shows white, as if something broke.

The `useDataLoader` lets you pass an `onError` callback, so you can just pass the catch block as the `onError` instead of using `try...catch` inside `SubmissionFeaturePage`:

```
const featureDataLoader = useDataLoader(
  () =>
    biohubApi.features.getSubmissionFeatureById(
      Number(submissionId),
      Number(submissionFeatureId)
    ),
  (error: unknown) => {
    const status = (error as APIError)?.status
    if (status === 401 || status === 403) {
      navigate('/forbidden', { replace: true });
    }
  }
);
```

mauberti-bc:
There are a couple sonar cloud issues (see comment sonarqubecloud) that should be fixed

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,132 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [submissionId, submissionFeatureId]);
+
+  const { data, isLoading } = featureDataLoader;
+
+  const [feature, relatedFeatures] = useMemo(() => {
+    return [data?.feature, data?.relatedFeatures] as const;
+  }, [data]);
+
+  const isInCart = cartFeatures.some((f) => f.submission_feature_id === Number(submissionFeatureId));
```
mauberti-bc: I don't think this will work because cartFeatures is paginated. If the feature is not in the current page's rows, it will incorrectly show `not in cart`.

We probably want an endpoint like `GET /api/carts/:cartId/features/:submissionFeatureId`, which returns `{ inCart: boolean }`.

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,132 @@
...
+  if (!feature) {
+    return (
+      <Box display="flex" justifyContent="center" alignItems="center" minHeight={300} p={2}>
+        <Typography color="text.secondary">No data available</Typography>
+      </Box>
+    );
+  }
 
   return (
     <Container maxWidth="xl">
```
mauberti-bc: This container prevents the pageHeader from taking the page's width: 

Image

You have another maxWidth='xl' container below for the content, which is correct, so you can just replace the outermost container with a Box. We don't want any space between the left and right edges and the page header, like the Submissions page and Policies page.

app/src/features/submissions/page/features/components/SubmissionFeatureProperties.tsx
```diff
@@ -0,0 +1,45 @@
...
+    flex: 0.7
+  }
+];
+
+export const SubmissionFeatureProperties = ({ data }: SubmissionFeaturePropertiesProps) => {
+  const stringifiedProperties = jsonStringifyObjectProperties(data);
+
+  const rows = Object.entries(stringifiedProperties).map(([key, value]) => ({
+    id: key,
+    property: key.replace(/_/g, ' '),
```
mauberti-bc: Note code smell: Prefer `String#replaceAll()` over `String#replace()`.

app/src/features/search/result/layout/table/SearchResultTableLayout.tsx
```diff
@@ -8,6 +8,7 @@ interface SearchResultTableLayoutProps {
   results: SearchFeatureResultWithRelevancy[];
   cartFeatureIds: Set<number>;
   onRowSelectionModelChange: (rowSelectionModel: GridRowSelectionModel) => void;
+  onClick?: (result: SearchFeatureResultWithRelevancy) => void;
   onDownload?: (result: SearchFeatureResultWithRelevancy) => void;
```
mauberti-bc, line 12: You can ignore this code smell since onDownload isn't wired yet

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,132 @@
...
   useEffect(() => {
-    featureDataLoader.load();
-  }, [featureDataLoader]);
+    featureDataLoader.refresh();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [submissionId, submissionFeatureId]);
+
+  const { data, isLoading } = featureDataLoader;
+
+  const [feature, relatedFeatures] = useMemo(() => {
```
mauberti-bc: nit: this is a bit awkward. Cleaner to just memoize the whole response data object:
```
const { feature, relatedFeatures } = useMemo(() => featureDataLoader.data, [featureDataLoader])
```

app/src/features/submissions/page/features/SubmissionFeaturePage.tsx
```diff
@@ -1,43 +1,132 @@
...
+          feature_description: null,
+          submission_name: feature.submission_name,
+          is_secured: feature.secured,
+          relevancy_score: 0
+        }
+      ]);
+    }
+  };
+
+  if (!feature) {
```
mauberti-bc: You can remove this early return. Instead, pass the fallback to the `noDataFallback` prop in the `LoadingGuard`, and set `hasNoData={!feature}` prop.

mauberti-bc (changes requested):

api/src/paths/cart/{cartId}/feature/{cartSubmissionFeatureId}/index.ts
```diff
@@ -11,6 +11,88 @@ import { makePaginationOptionsFromRequest } from '../../../../../utils/paginatio
...
+  ],
+  parameters: [
+    {
+      in: 'path',
+      name: 'cartId',
+      required: true,
+      schema: { type: 'string', format: 'uuid' }
+    },
+    {
+      in: 'path',
```
mauberti-bc, line 44: The client will not know the `cartSubmissionFeatureId` when viewing a feature page. The only identifiers available to the client are:

- `cartId`
- `submissionFeatureId`

The requirement for the client is to know whether a feature is in a cart, so it would be cleaner to pass filters to the existing `GET /cart/{cartId}/feature` endpoint, like  `GET /cart/{cartId}/feature?submissionFeatureId=5`.

Keep this endpoint as-is because it does have a use case, but then add filters to the existing `GET /cart/{cartId}/feature`. Then wire up the feature page to use existing `GET /cart/{cartId}/feature` with filters instead.

mauberti-bc (changes requested):

mauberti-bc (approved):
👍

PR #359, opened 2026-03-06
SIMSBIOHUB 881: Adding Ticket ID to Data Request

J-pilon:
## Links to Jira Tickets

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-881

## Description of Changes

**Database** – Adds `ticket_id` to `data_request`: migration adds `ticket_id` (UUID, NOT NULL), FK to `ticket(ticket_id)`, and index `data_request_ticket_idx`.

**API – Models** – Data request models now include `ticket_id`: added to `DataRequest` (Zod), `FlatDataRequestWithStatus`, `DataRequestWithStatus`, and `CreateDataRequestPayload`.

**API – OpenAPI & schema** – `ticket_id` added to the data request response schema and OpenAPI spec (required, UUID, description: “ID of the ticket associated with this data request”).

**API – Data request creation** – When a data request is created, a ticket is created first via a new private `DataRequestService.createTicket()` (subject “Data Request”, description = request reason, priority “medium”). The new ticket’s `ticket_id` is passed into the repository so each data request is stored with a non-null `ticket_id`.

**API – Repository** – `DataRequestRepository` selects and returns `ticket_id` in all relevant queries; `createDataRequest` accepts and persists `ticket_id` in the payload.

**API – Utils** – `_transformFlatDataRequestToNested` in `data-request.ts` includes `ticket_id` in the nested response so all service responses expose it.

**API – Tests** – Data request service, repository, and route tests updated to use and assert `ticket_id`; seeds updated so data requests are created with linked tickets.

## Testing Instructions

- **Create data request**: Create a data request via the API; response and DB row should include a valid `ticket_id` pointing to a new ticket with subject “Data Request” and the request reason as description.
- **Get/list data requests**: GET by ID and list endpoints should return `ticket_id` for each data request.
- **API tests**: Run data request service, repository, and path tests; all should pass with the new `ticket_id` expectations.
- **Seeds**: Run seeds; data requests should be created with valid `ticket_id` values.

### DIFF
--- a/api/src/services/data-request-service.ts
+++ b/api/src/services/data-request-service.ts
@@ -11,6 +11,7 @@ import {
 import { DataRequestStatusEnum } from '../models/data-request-status';
 import { PolicyEffect } from '../models/policy-statement';
 import { Team } from '../models/team';
+import { CreateTicketRequest, Ticket } from '../models/ticket';
 import { DataRequestRepository } from '../repositories/data-request-repository';
 import {
   _generateDataRequestPolicyName,
@@ -24,6 +25,7 @@ import { TeamPolicyService } from './access-policy/team-policy-service';
 import { TeamService } from './access-policy/team-service';
 import { DataRequestStatusService } from './data-request-status-service';
 import { DBService } from './db-service';
+import { TicketService } from './ticket-service';
 
 /**
  * Service for managing data requests.
@@ -98,7 +100,7 @@ export class DataRequestService extends DBService {
   /**
    * Create a new data request.
    *
-   * Creates a team for the requester, a wildcard access policy expiring in 30 days
+   * Creates a team for the requester, at ticket, a wildcard access policy expiring in 30 days
    * linked to the team, and auto-approves the request.
    *
    * @param {CreateDataRequest} payload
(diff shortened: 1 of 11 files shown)


api/src/services/data-request-service.ts
```diff
@@ -167,6 +175,22 @@ export class DataRequestService extends DBService {
...
   }
 
+  /**
+   * Create a new ticket.
+   *
+   * @param {CreateTicketRequest} params
+   * @return {Promise<Ticket>}
+   * @memberof DataRequestService
+   */
+  private async createTicket(params: CreateTicketRequest): Promise<Ticket> {
```
mauberti-bc: Move this service method to TicketService and just have `dataRequestService.createDataRequest `call `ticketService.createTicket`. We try to keep the services focused on their domain, without thin wrappers for other services' methods.

mauberti-bc (changes requested):

api/src/services/data-request-service.ts
J-pilon: @mauberti-bc I may be misunderstanding you but this method does currently call TicketService.createTicket. 

I just extracted this logic to a private method so that the createDataRequest method was easier to read. I can remove this private method and have the logic in the createDataRequest method which would look like this:
```
  async createDataRequest(payload: CreateDataRequest): Promise<DataRequestWithStatus> {
    let teamId = payload.team_id;
    if (!teamId) {
      const team = await this.createTeam(payload.requested_by);
      teamId = team.team_id;
    }

    const ticketService = new TicketService(this.connection);
    const ticket = await ticketService.createTicket({
      subject: 'Data Request',
      description: payload.reason,
      priority: 'medium'
    });

    const payloadWithIds = { ...payload, team_id: teamId, ticket_id: ticket.ticket_id };

    const dataRequest = await this.dataRequestRepository.createDataRequest(payload.requested_by, payloadWithIds);

    const policy = await this.createPolicy(dataRequest.data_request_id);
    await this.createTeamPolicy({ teamId, policyId: policy.policy_id });

    const dataRequestStatusService = new DataRequestStatusService(this.connection);
    // initially defaults status to APPROVED for development
    const dataRequestStatus = await dataRequestStatusService.createDataRequestStatus(
      dataRequest.data_request_id,
      DataRequestStatusEnum.enum.APPROVED,
      undefined
    );

    return {
      ...dataRequest,
      data_request_status: dataRequestStatus
    };
  }
  ```

mauberti-bc (approved):
👍

PR #363, opened 2026-03-10
SIMSBIOHUB-901: Add new property_* tables for search

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-901

## Description of Changes

- The `property_*` tables need to use foreign keys where possible for performant queries
- Replaces the `search_*` tables with new `property_*` tables that are better designed for search
    - Foreign keys to the `taxon_id` 
- Changes how arrays are supported:
    - Rather than using 'array' as its own `feature_type_property`, we just set `allow_multiple` as true or false (default) in the `feature_property`
    - In other words, the definition of the property includes whether multiple is allowed 
## Testing Notes

- This PR adds net-new tables
- After this PR, we can switch application code to use `property_*` instead of `search_*`, and migrate existing data.
- We will drop the `search_*` tables after they are deprecated


database/src/migrations/20260309120000_add_property_tables.ts
```diff
@@ -0,0 +1,389 @@
...
+    CREATE TRIGGER audit_submission_feature_property_geometry
+      BEFORE INSERT OR UPDATE OR DELETE ON submission_feature_property_geometry
+      FOR EACH ROW EXECUTE PROCEDURE tr_audit_trigger();
+    CREATE TRIGGER journal_submission_feature_property_geometry
+      AFTER INSERT OR UPDATE OR DELETE ON submission_feature_property_geometry
+      FOR EACH ROW EXECUTE PROCEDURE tr_journal_trigger();
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
```
dwhogan, line 362: drop `allow_multiple` to `feature_type_property`

database/src/migrations/20260309120000_add_property_tables.ts
```diff
@@ -0,0 +1,389 @@
...
+    DROP TRIGGER IF EXISTS audit_submission_feature_property_code ON submission_feature_property_code;
+    DROP TRIGGER IF EXISTS journal_submission_feature_property_timestamp ON submission_feature_property_timestamp;
+    DROP TRIGGER IF EXISTS audit_submission_feature_property_timestamp ON submission_feature_property_timestamp;
+    DROP TRIGGER IF EXISTS journal_submission_feature_property_boolean ON submission_feature_property_boolean;
+    DROP TRIGGER IF EXISTS audit_submission_feature_property_boolean ON submission_feature_property_boolean;
+    DROP TRIGGER IF EXISTS journal_submission_feature_property_number ON submission_feature_property_number;
+    DROP TRIGGER IF EXISTS audit_submission_feature_property_number ON submission_feature_property_number;
+    DROP TRIGGER IF EXISTS journal_submission_feature_property_string ON submission_feature_property_string;
+    DROP TRIGGER IF EXISTS audit_submission_feature_property_string ON submission_feature_property_string;
+
```
dwhogan, line 390: drop indexes

database/src/migrations/20260309120000_add_property_tables.ts
```diff
@@ -0,0 +1,389 @@
+import { Knex } from 'knex';
```
dwhogan: nit: convention is `import type { Knex } from 'knex';`

dwhogan (commented):
a few adds the migrations down()

database/src/migrations/20260309120000_add_property_tables.ts
mauberti-bc: I believe DROP TABLE also drops indexes, constraints, etc. that depend on it?

database/src/migrations/20260309120000_add_property_tables.ts
mauberti-bc: Fixed

database/src/migrations/20260309120000_add_property_tables.ts
mauberti-bc: Fixed. Some `import { Knex } from 'knex';` snuck into old migrations but good convention to follow moving forward.

database/src/migrations/20260309120000_add_property_tables.ts
dwhogan: You are right.

dwhogan (approved):
👍

PR #367, opened 2026-03-10
SIMSBIOHUB-824: View and manage downloads

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-824](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-824)

## Description of Changes

As an authenticated BioHub user, I want to view the downloads I have requested, see which are ready, and access previously downloaded packages, so that I can manage and retrieve my data efficiently.

Acceptance Criteria
1. Viewing Downloads  
WHEN a user navigates to the downloads page,  
  THEN a table should display all download records associated with that user,  
  AND each record should include details such as request date, status, and number of features.
2. Download Button Availability  
WHEN a download record has status `ready` or `downloaded`,  
  THEN a "Download" button should be displayed next to that record,  
  AND clicking the button should allow the client to retrieve the zip package.
3. Status Indicators  
WHEN the download table is displayed,  
  THEN each record should clearly indicate its current status (`pending`, `ready`, `downloaded`) for transparency.
Notes  
The table should only show downloads belonging to the authenticated client.  
For records with `pending` status, no download button should be available, but the user can see the 'pending' status

## Testing Notes

on the search page click "download all" or add features to the cart and click "checkout"


api/src/repositories/download/download-repository.ts
```diff
@@ -102,59 +113,100 @@ export class DownloadRepository extends BaseRepository {
...
+
+    return { downloads, count };
+  }
+
+  /**
+   * Builds a CTE that resolves all download IDs a user is authorized to access.
+   *
+   * Three access paths are unioned:
+   * 1. Owner — user created or claimed the download (system_user_id matches)
+   * 2. Shared — user has a direct entry in download_share
```
mauberti-bc, line 185: I'm thinking we should better align the access model for downloads with the team concept:
* Replace `download_share` with `download_team`, which joins `download_id` and `team_id`
* Keep the `team_id` on `download`, and then this team's role more clearly becomes part of the mechanism for granting users access to that download (expanded on below)

--

Conceptually, downloads don't have owners, because they're just pre-packaged files that can be downloaded by anyone with access.
* If a set of features is downloaded 100 times per day, every day, we can avoid redundant download jobs by:
    * Check whether the requested feature set already exists in a download
    * Join the system_user_id to the `download.team_id`

We could go even further and re-use fragments, but that would really multiply scope. As follow up to this PR, I think trying to re-use downloads by checking if the requested feature set already exists in a download is reasonable.

api/src/services/object-storage/object-storage-service.ts
```diff
@@ -64,6 +69,39 @@ export class ObjectStorageService {
...
+            secretAccessKey: process.env.OBJECT_STORE_SECRET_KEY_ID!
+          },
+          forcePathStyle: true,
+          region: 'ca-central-1'
+        }),
+        name: process.env.OBJECT_STORE_BUCKET_NAME!
+      });
+    }
+
+    if (process.env.QUARANTINE_OBJECT_STORE_URL_PUBLIC) {
```
mauberti-bc: Ensure that we're never sending presigned GET urls for the quarantine bucket to clients. We only need to send presigned PUT urls.

app/src/features/search/result/sidebar/download/downloads/DownloadSidebarDownloads.tsx
```diff
@@ -0,0 +1,126 @@
...
+export const DownloadSidebarDownloads = () => {
+  const biohubApi = useApi();
+  const dialogContext = useDialogContext();
+  const [page, setPage] = useState(1);
+
+  const downloadsDataLoader = useDataLoader((pagination: ApiPaginationRequestOptions) =>
+    biohubApi.download.getDownloads(pagination)
+  );
+
+  downloadsDataLoader.load({ page: 1, limit: PAGE_SIZE });
```
mauberti-bc: React throws an error in the console if load is called outside of a useEffect.

compose.yml
```diff
@@ -144,6 +145,7 @@ services:
       - DB_IDLE_TIMEOUT=${DB_IDLE_TIMEOUT}
       # Object Store (S3)
       - OBJECT_STORE_URL=${OBJECT_STORE_URL}
+      - OBJECT_STORE_URL_PUBLIC=${OBJECT_STORE_URL_PUBLIC}
```
mauberti-bc, line 148: The presigned download URL that the client receives uses "minio:<port>" instead of "localhost:<port>". You can look at `generateMultipartUploadPresignedUrls` for how we solved this in the case of upload URLs.

Note this should only affect development, since the host will always be the remote object store when deployed.

Image

app/src/features/search/result/sidebar/download/feature/DownloadFeatureCard.tsx
```diff
@@ -0,0 +1,101 @@
+import { mdiChevronDown, mdiChevronUp, mdiDownload } from '@mdi/js';
+import Icon from '@mdi/react';
+import { Button, Card, Chip, ChipProps, Collapse, IconButton, Link, Stack, Typography } from '@mui/material';
+import { grey } from '@mui/material/colors';
+import { useState } from 'react';
+import { DownloadRecord } from 'interfaces/useDownloadApi.interface';
+import { isDownloadReady } from '../downloads/DownloadSidebarDownloads';
+
```
mauberti-bc, line 9: nit: move statusChipProps to constants. See `FEATURE_TYPE_CONFIG` in constants/feature-type.ts as an example

app/src/features/search/result/sidebar/download/downloads/DownloadSidebarDownloads.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * `pending` and `processing` are still building; `failed` has no artifact.
+ */
+export const isDownloadReady = (status: string): boolean => status === 'ready' || status === 'downloaded';
+
+const PAGE_SIZE = 10;
+
+export const DownloadSidebarDownloads = () => {
+  const biohubApi = useApi();
+  const dialogContext = useDialogContext();
+  const [page, setPage] = useState(1);
```
mauberti-bc, line 25: Possibly use `useServerPaginatedDataGrid` hook, even though we don't use an actual DataGrid, since it handles paginated requests? It will make the addition of search and filtering easier down the road.

mauberti-bc (changes requested):

api/src/repositories/download/download-repository.ts
dwhogan: I opened a ticket for this, but I will get it done here.

compose.yml
dwhogan: I can not reproduce, maybe make api

api/src/services/object-storage/object-storage-service.ts
dwhogan: removed

app/src/features/search/result/sidebar/download/downloads/DownloadSidebarDownloads.tsx
dwhogan: don't use useDataLoader outside effect, don't use useDataLoader outside effect...so I remeber

app/src/features/search/result/sidebar/download/feature/DownloadFeatureCard.tsx
dwhogan: 👍

app/src/features/search/result/sidebar/download/downloads/DownloadSidebarDownloads.tsx
dwhogan: i thought about extracting to `useServerPaginatedData`, but felt is was out of scope.  Settled on the simple implementation.

compose.yml
mauberti-bc: Oops, I'm probably missing an env var and it's falling back to minio instead of localhost

mauberti-bc (dismissed):
👍

mauberti-bc (approved):
👍

PR #369, opened 2026-03-11
SIMSBIOHUB-907: Optimistic data loader for updating UI state before promises resolve

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-907
- When an async request is made to update server-dependent state, we often want to update the client state immediately, as if the async call instantly succeeded

## Description of Changes

- Adds a reusable `useOptimisticDataLoader` that mirrors `useDataLoader`
- The optimistic data loader is initialized with a normal `dataLoader`
- The optimistic data loader exposes a `refresh` method that takes
    - The optimistic state, which is the expected state upon success
    - A mutation, which is the async request to update the state
    - An optional rollback, which runs if the mutation fails (eg. setting a snackbar)
    - An optional onSuccess callback, which runs after the mutation (eg. setting a snackbar, updating other state)
- Refactors the `cartContext` to use the new `useOptimisticDataLoader`

## Testing Notes

- Adding and removing features from the cart should behave as before with immediate state updates


app/src/hooks/useOptimisticMutation.ts
```diff
@@ -0,0 +1,60 @@
...
+        const result = await config.mutation(context);
+        config.onSuccess?.(result, context);
+        return result;
+      } catch (error) {
+        setup.setData(currentState);
+        config.onRollback?.(context);
+        throw error;
+      }
+    },
+    [setup]
```
dwhogan: Consider a useRef to remove setup from deps

app/src/contexts/useCartOptimisticActions.ts
```diff
@@ -0,0 +1,128 @@
...
+        return {
+          optimisticState,
+          mutation: () =>
+            cartApi.addCartFeatures(
+              cartId,
+              { features: optimisticAdds.map((feature) => feature.submission_feature_id) },
+              getPaginationParams(optimisticState.features.length)
+            ),
+          onSuccess: applyLoadSuccess,
+          onRollback: ({ currentState: previousState }) => {
```
dwhogan: nit: is rollback now handled by useOptimisticMutation?

dwhogan (commented):
Nice extracts, to make better UX a pattern👍

app/src/contexts/useCartOptimisticActions.ts
mauberti-bc: Thanks, good catch - onRollback doesn't need to be passed unless there are more side effects like snackbars.

app/src/hooks/useOptimisticMutation.ts
mauberti-bc: Done

dwhogan (approved):
👍

PR #370, opened 2026-03-11
SIMSBIOHUB 829: Public Submission Page

J-pilon:
**Links to Jira**

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-829

## Description of Changes

**App – Submission detail page**  
Adds **SubmissionDetailPage**, a public page showing submission metadata (name, description), breadcrumbs (Search → submission name), and a server-side paginated/sortable table of unsecured features (ID, Feature Type). Secured features are excluded; a chip and alert indicate when the submission has secured features. Row click goes to the submission feature page.

**App – Routing**  
- **SubmissionRouter** (new): Handles `/submission/*` with `/submission/:submissionId` (SubmissionDetailPage) and `/submission/:submissionId/feature/:submissionFeatureId` (SubmissionFeaturePage).  
- **SearchRouter**: Limited to `/search/` (search form) and `/search/list` (results). Submission detail and feature pages are no longer under `/search`.  
- **AppRouter**: Wraps Search and Submission routes in a single **CartContextProvider** so cart state is shared between search and submission flows.

**App – Navigation**  
- From search results, result click goes to `/submission/:submissionId` (no longer `/search/:submissionId`).  
- Submission detail table row click and feature-page back/related links use `/submission/:submissionId` and `/submission/:submissionId/feature/:submissionFeatureId` as above.

**API – Authorization**  
Submission route(s) for the submission record and features use optional bearer (unauthenticated access allowed). **getSubmissionFeatures** supports anonymous requests (e.g. uses API user connection when no token) so the public submission detail page can load submission and features data.

---

## Testing Instructions

1. From Search Results, click a result; confirm navigation to **/submission/:submissionId** and that the submission detail page shows name, description, and features table.  
2. Confirm breadcrumbs show Search → submission name.  
3. For a submission with secured features, confirm the “Contains secured features” chip and alert appear and secured features are not in the table.  
4. Use pagination and sorting on the features table; confirm server-side updates.  
5. Click a row in the features table; confirm navigation to **/submission/:submissionId/feature/:submissionFeatureId**.  
6. Confirm cart context is shared when moving between search and submission pages (e.g. add to cart on submission feature, then go back to search and verify cart).  
7. Run submission-related API route tests; confirm they pass (including unauthenticated access to submission/features where applicable).


app/src/router/AppRouter.tsx
```diff
@@ -1,28 +1,51 @@
...
     <Routes>
       {/* Redirect base to search router */}
       <Route path="/" element={<Navigate to="/search" replace />} />
 
-      {/* Search Routes  */}
+      {/* Search and Submission routes share CartContextProvider */}
       <Route
-        path="/search/*"
         element={
           <>
```
mauberti-bc: nit: remove unnecessary fragment

app/src/router/AppRouter.tsx
```diff
@@ -1,28 +1,51 @@
...
-            <PageTitle title="Search Data" description="Search and download data" />
-            <SearchRouter />
+            <CartContextProvider>
+              <Outlet />
```
mauberti-bc: Nice

app/src/features/search/result/SearchResultPage.tsx
```diff
@@ -284,7 +285,7 @@ export const SearchResultPage = () => {
               rows={rows}
               isLoading={isLoading}
               view={view}
-              onClick={(result) => navigate(`/search/${result.submission_id}/feature/${result.submission_feature_id}`)}
+              onClick={(result) => navigate(`/submission/${result.submission_id}${location.search}`)}
```
mauberti-bc: Clicking on a feature in the search results should navigate to that feature, like before. It currently navigates to the submission, which we don't want.

There is currently no way to search submissions directly, which is intentional to start.

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+    }
+    submissionDataLoader.load(Number(submissionId));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [submissionId]);
+
+  useEffect(() => {
+    if (!submissionId) {
+      return;
+    }
+    featureDataLoader.load(Number(submissionId), paginationOptions);
```
mauberti-bc: Server-side pagination does not work, presumably because this `load`. The `dataLoader.load` method is designed to run only once. To make another request, use the `refresh` method instead.

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const api = useApi();
+
+  const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({ page: 0, pageSize: 10 });
+  const [sortModel, setSortModel] = useState<GridSortModel>([{ field: 'submission_feature_id', sort: 'asc' }]);
+
+  const paginationOptions: ApiPaginationRequestOptions = useMemo(() => {
+    return toApiPagination(paginationModel, sortModel);
+  }, [paginationModel, sortModel]);
+
+  const submissionDataLoader = useDataLoader((id: number) => api.submissions.getSubmissionRecordWithSecurity(id));
```
mauberti-bc: Don't use `id` as a prop because it's not immediately clear what the id represents. Submission ID? Feature ID? Prefer using the name of that property in the database, camel case, like `submissionId` or `submissionFeatureId`.

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+ * Public-facing submission detail page.
+ * Displays unsecured features in a paginated table and indicates when secured features exist.
+ */
+export const SubmissionDetailPage = () => {
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { submissionId } = useParams<{ submissionId: string }>();
+  const api = useApi();
+
+  const [paginationModel, setPaginationModel] = useState<GridPaginationModel>({ page: 0, pageSize: 10 });
```
mauberti-bc: You can simplify this component by using the `useServerPaginatedDataGrid` hook, which handles the refresh logic and orchestrating the data loader and local state. `TicketsPage` and `ManagePoliciesPage` show example usage.

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+      return;
+    }
+    featureDataLoader.load(Number(submissionId), paginationOptions);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [submissionId, paginationOptions]);
+
+  const submission: SubmissionRecordWithSecurity | undefined = submissionDataLoader.data;
+  const featureResponse: ISubmissionFeatureForReviewResponse | undefined = featureDataLoader.data;
+
+  const rows = useMemo(
```
mauberti-bc: The frontend does not need to do any filtering based on security, because the API will only return what the client is authorized to see. You can remove the `.filter((f) => !f.secured)`

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const hasSecuredFeatures = featureResponse?.features.some((f) => f.secured) ?? false;
+
+  const handleRowClick = (params: GridRowParams<SubmissionRow>, _event: MuiEvent<React.MouseEvent>) => {
+    const featureId = params.row.submission_feature_id;
+    navigate(`/submission/${submissionId}/feature/${featureId}${location.search}`);
+    return;
+  };
+
+  return (
+    <Box>
```
mauberti-bc: nit: unnecessary box?

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  ISubmissionFeatureForReviewResponse,
+  SubmissionRecordWithSecurity
+} from 'interfaces/useSubmissionsApi.interface';
+import { useEffect, useMemo, useState } from 'react';
+import { Link as RouterLink, useLocation, useNavigate, useParams } from 'react-router-dom';
+import { ApiPaginationRequestOptions } from 'types/pagination';
+import { toApiPagination } from 'utils/pagination';
+
+type SubmissionRow = {
+  id: number;
```
mauberti-bc: Rename `id` to `submissionId`

app/src/features/submissions/SubmissionDetailPage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  submission_feature_id: number;
+  feature_type_name: string;
+};
+
+const columns: GridColDef[] = [
```
mauberti-bc: To prevent this single component from getting big, I'd suggest creating a prop-driven `SubmissionFeatureTable` that will own the data grid and column definitions, and it would receive callbacks like onRowClick. You could define the columns inside that component's functional component instead of as a constant.

Then this `SubmissionDetailPage` can focus on the business logic and handlers, while the SubmissionFeatureContainer just displays what it receives.

mauberti-bc (changes requested):

app/src/features/submissions/SubmissionDetailPage.tsx
J-pilon: @mauberti-bc The reason for the filter is to display a banner or chip if there is secure features:
```
 Secured Features Indicator  
WHEN a submission contains secured features,  
  THEN the page should indicate that some features are secured and not visible to the user.
  ```
  
  Should this be removed?

Comment (mauberti-bc) [reply app/src/features/submissions/SubmissionDetailPage.tsx]: `const hasSecuredFeatures = featureResponse?.features.some((f) => f.secured) ?? false;`
That line sets the flag for whether to show the "has secured features" banner. You might be interpreting "...and not visible to the user" as "we should remove those rows", but we just want the banner to imply that those secured rows are not visible to the user.

But this is not wanted:
```
featureResponse?.features
        .filter((f) => !f.secured)
        .map((f) => ({
          id: f.submission_feature_id,
          submission_feature_id: f.submission_feature_id,
          feature_type_name: f.feature_type_name
        })) ?? [],
```
If `rows` gets passed to the data grid, and `rows` removes records where `secured = true`, then once the backend starts sending secured records that a user is authorized to see, we'll be removing them from the UI?

mauberti-bc (approved):
👍

PR #374, opened 2026-03-16
SIMSBIOHUB-863: Caching secured records

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-863](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-863)

## Description of Changes

As an authenticated BioHub client who has been granted access to secured records via one or more policies, I want to use the front-end search page in BioHub and see my secured records in the results. 

Acceptance Criteria 

1. Secured Cache Table

WHEN a policy is created
THEN the secured records made available under that policy should be cached in a new table called team_feature, that joins team_id and the submission_feature_id 

2. Front-End Search 

WHEN a client is logged into BioHub
THEN The search should query both the submission_feature table as well as the cached secured submissions table for any team_id that they belong to
WHEN The client has an active data policy 
THEN Those records secured under that policy should be displayed in the search results alongside the unsecured features. 

## Testing Notes

### DIFF
--- a/api/src/repositories/authorization/team-feature-repository.ts
+++ b/api/src/repositories/authorization/team-feature-repository.ts
@@ -40,14 +40,6 @@ export class TeamFeatureRepository extends BaseRepository {
    * need cache entries. Uses decomposed URN columns on policy_statement
    * (urn_submission_id, urn_feature_type, urn_feature_id) for indexed matching.
    *
-   * Uses EXISTS semi-join so each submission_feature drives the outer scan exactly
-   * once — PostgreSQL short-circuits on the first matching policy statement. This
-   * avoids the hash aggregate that SELECT DISTINCT would build over the full
-   * (multiplied) result set when JOINing through team_policy → policy →
-   * policy_statement. At 10M features × 5 policies, EXISTS prevents 50M
-   * intermediate rows from materializing. Rows are unique by construction, so no
-   * ON CONFLICT clause is needed.
-   *
    * @param {string} teamId - The team UUID to populate cache for.
    * @return {Promise<void>}
    * @memberof TeamFeatureRepository
--- a/api/src/repositories/authorization/team-feature-repository.ts
+++ b/api/src/repositories/authorization/team-feature-repository.ts
@@ -32,42 +32,21 @@ export class TeamFeatureRepository extends BaseRepository {
   }
 
   /**
-   * Populate the team_feature cache for a specific team by resolving all active
-   * policy statement URNs (including wildcards) to concrete submission_feature_ids.
+   * Get submission IDs that contain secured features matching a team's policy
+   * grants. Used to partition the cache rebuild into per-submission batches
+   * so each recursive walk stays small.
    *
-   * Only secured features (those with an active submission_feature_security row)
-   * are cached — unsecured features are already visible to everyone and don't
-   * need cache entries. Uses decomposed URN columns on policy_statement
-   * (urn_submission_id, urn_feature_type, urn_feature_id) for indexed matching.
-   *
-   * @param {string} teamId - The team UUID to populate cache for.
-   * @return {Promise<void>}
+   * @param {string} teamId - The team UUID to resolve grants for.
+   * @return {Promise<number[]>} Submission IDs with secured features the team may access.
    * @memberof TeamFeatureRepository
    */
-  async populateTeamFeatureCache(teamId: string): Promise<void> {
+  async getSecuredSubmissionIds(teamId: string): Promise<number[]> {
     const sql = SQL`
-      WITH RECURSIVE secured_features AS (
-        -- Base: features with a direct security row
-        SELECT sf.submission_feature_id
-        FROM submission_feature sf
+      SELECT DISTINCT sf.submission_id
+      FROM submission_feature sf
         JOIN submission_feature_security sfs
           ON sfs.submission_feature_id = sf.submission_feature_id
           AND sfs.record_end_date IS NULL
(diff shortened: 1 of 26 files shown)


api/src/services/access-policy/policy-service.ts
```diff
@@ -248,6 +267,10 @@ export class PolicyService extends DBService {
       })
     );
 
+    // Refresh team_feature cache so secured search results reflect the updated policy statements
+    const teamFeatureService = new TeamFeatureService(this.connection);
+    await teamFeatureService.refreshCacheForPolicy(policyId);
```
mauberti-bc: Nice, keeping the cache updates event-driven. Move this refresh to background pgboss job.

api/src/repositories/authorization/team-feature-repository.ts
```diff
@@ -0,0 +1,85 @@
...
+   * need cache entries. Uses decomposed URN columns on policy_statement
+   * (urn_submission_id, urn_feature_type, urn_feature_id) for indexed matching.
+   *
+   * @param {string} teamId - The team UUID to populate cache for.
+   * @return {Promise<void>}
+   * @memberof TeamFeatureRepository
+   */
+  async populateTeamFeatureCache(teamId: string): Promise<void> {
+    const sql = SQL`
+      WITH RECURSIVE secured_features AS (
```
mauberti-bc: To keep this table small, we could move the recursive part of the query into the run-time search query, and just cache the root-level features.

- If policy has `urn:*:*:1`, we just insert submission_feature_id = 1 into the cache. Then during search, we recursively find children to include.
- If policy has `urn:*:*:*`, we insert all submission_feature_id records that don't have a parent (parent_submission_feature_id = null)
- If policy has `urn:1:*:*`, we again insert all submission_feature_id records without a parent AND from submission 1

The mechanism would be that we cache the root-level features that a team has access to, plus any features explicitly referenced by a policy (eg. `urn:*:*:10`). 

Cache becomes more like a pre-filter that we calculate as a performance boost.

api/src/services/access-policy/policy-service.ts
```diff
@@ -98,12 +99,26 @@ export class PolicyService extends DBService {
...
+    const teamFeatureService = new TeamFeatureService(this.connection);
+
+    // Must fetch affected teams BEFORE the soft-delete removes the team_policy association
+    const teamIds = await teamFeatureService.getTeamIdsForPolicy(policyId);
+
+    await this.policyRepository.deletePolicy(policyId);
+
+    // Refresh cache for each affected team after the policy is gone
+    for (const teamId of teamIds) {
+      await teamFeatureService.refreshCacheForTeam(teamId);
```
mauberti-bc: Move the `refreshCacheForTeam` into a background pgboss job, so that system admins are not waiting for the query to complete when trying to delete a policy.

api/src/services/access-policy/team-policy-service.ts
```diff
@@ -27,14 +28,25 @@ export class TeamPolicyService extends DBService {
 
     if (existingPolicies.length > 0) {
       const existingPolicy = existingPolicies[0];
+
+      // Refresh cache even on idempotent return — the cache may be stale
+      const teamFeatureService = new TeamFeatureService(this.connection);
```
mauberti-bc: Is this necessary? We need to refresh cache when we:
- create policy
- delete policy
- update policy
- create team_policy
- delete team_policy

Idempotency doesn't seem like a concern here because there is a unique constraint on (team_id, policy_id). 

It seems cleaner to just let the database handle duplicates and actually just remove the `
    if (existingPolicies.length > 0) {`block entirely, and instead just throw an error if the incoming team already has the incoming policies? I realize that conditional block wasn't your addition in the first place!

database/src/migrations/20260310165917_team_feature_cache.ts
```diff
@@ -0,0 +1,82 @@
...
+    SET SEARCH_PATH = biohub, public;
+
+    --------------------------------------------------------------------------------
+    -- Create team_feature cache table
+    --------------------------------------------------------------------------------
+    CREATE TABLE team_feature (
+      team_feature_id       integer GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
+      team_id               uuid NOT NULL,
+      submission_feature_id integer NOT NULL,
+      create_date           timestamptz(6) DEFAULT now() NOT NULL,
```
mauberti-bc: Like you suggested, drop audit triggers and probably even drop the audit colums (create_*, update_*, revision_count) to keep it small

mauberti-bc (changes requested):

api/src/services/access-policy/team-policy-service.ts
dwhogan: right

api/src/repositories/authorization/team-feature-repository.ts
dwhogan: addressed in 914

api/src/services/access-policy/policy-service.ts
dwhogan: 👍

database/src/migrations/20260310165917_team_feature_cache.ts
dwhogan: 👍

api/src/services/access-policy/policy-service.ts
dwhogan: 👍

dwhogan:
moving to 914

PR #390, opened 2026-03-24
SIMSBIOHUB-914: Search Results Include Secured Records for Authorized Users

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-914](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-914)

## Description of Changes

- Scope lifecycle hooks: creating/updating/deleting policy statements and team-policies automatically maintains scope tables
- Background job (`compute-scope-anchors`) resolves anchor features for new scopes
- Search query uses scope anchors + recursive CTE for authenticated users
- Lock icon on search results for secured features (list and table views)
- Download authorization uses scope walk-up instead of direct policy match

## Testing Notes

 Open app:

| Scenario | Datasets (1,4,7,10,13) | Telemetry (3,6,9) | Sample sites (2,5,8) | Unapproved (12) |                                                             
  |---|---|---|---|---|                                                                                                           
  | Authenticated dev | ✅ Visible | ✅ Visible (team grant) | ❌ Hidden (no team) | ❌ Hidden |                                                                 
  | Anonymous | ✅ Visible | ❌ Hidden (secured) | ❌ Hidden (secured) | ❌ Hidden |   

Run tests to validate **Policy mutation hooks**


api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+   *
+   * One policy statement maps to exactly one scope. Multiple statements
+   * can share the same scope (via scope_hash deduplication).
+   *
+   * @param policyStatementId UUID of the policy statement
+   * @param securityScopeId UUID of the security scope
+   */
+  async insertPolicyStatementScope(policyStatementId: string, securityScopeId: string): Promise<void> {
+    const sqlStatement = SQL`
+      INSERT INTO policy_statement_scope (policy_statement_scope_id, policy_statement_id, security_scope_id)
```
mauberti-bc: major nit: don't set the primary key field and let the field's default (gen_random_uuid()) populate it. Just one less distraction in the SQL

api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+   * anchor, not from the anchor down.
+   *
+   * Only the topmost secured features (those with no secured ancestor at any
+   * depth) are anchors. A secured feature whose parent or grandparent is also
+   * secured is not an anchor — only the highest secured node in each chain
+   * qualifies. This prevents duplicate walk-up hits when both an ancestor and
+   * descendant are secured by the same scope. Uses a recursive CTE to walk the
+   * full parent chain — submission_feature trees can be arbitrarily deep.
+   *
+   * Uses a server-side cursor to avoid materializing the full result set in
```
mauberti-bc: Excellent documentation

api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+   * dataset roots that expand to 10M telemetry features. The expansion never
+   * happens — the walk-up search strategy checks from the candidate up to the
+   * anchor, not from the anchor down.
+   *
+   * Only the topmost secured features (those with no secured ancestor at any
+   * depth) are anchors. A secured feature whose parent or grandparent is also
+   * secured is not an anchor — only the highest secured node in each chain
+   * qualifies. This prevents duplicate walk-up hits when both an ancestor and
+   * descendant are secured by the same scope. Uses a recursive CTE to walk the
+   * full parent chain — submission_feature trees can be arbitrarily deep.
```
mauberti-bc: If the policy statement's urn is `urn::*::*::55`, don't we have to insert `submission_feature_id = 55` as an anchor? In that case, we don't want to walk up the tree and include any of 55's parents as an anchor. We only want to include 55, which should then expand to all of 55's descends in the search query.

api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+      return [];
+    }
+
+    const knex = getKnex();
+    const query = knex
+      .distinct('security_scope_id')
+      .from('policy_statement_scope')
+      .whereIn('policy_statement_id', policyStatementIds);
+
+    const response = await this.connection.knex<{ security_scope_id: string }>(query);
```
mauberti-bc: nit: use zod objects instead of type hints in `this.connection.knex`, for the benefit of also validating the output.

api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+   *
+   * Why wipe-and-re-derive instead of surgical removal: a team can reach the same
+   * scope through multiple policies. Removing scopes from a deleted policy requires
+   * a graph reachability check (is the scope still reachable through another policy?).
+   * At ~30 rows per team, DELETE + INSERT completes in < 1ms — faster than the
+   * reachability query and guaranteed correct.
+   *
+   * @param teamId UUID of the team to rebuild scopes for
+   */
+  async rebuildTeamSecurityScopes(teamId: string): Promise<void> {
```
mauberti-bc: Suggestion: move the orchestration of `DELETE` then `INSERT` into the service layer. 

Repositories should usually have just CRUD methods - it's the service's responsibility to define how "rebuild" maps to insert, update, delete, etc.

api/src/services/security-service.ts
```diff
@@ -330,18 +331,49 @@ export class SecurityService extends DBService {
...
+   * Only features with zero remaining rules should stop being scope roots — deleting
+   * anchors for still-secured features would create a coverage gap where the walk-up
+   * search can't find a matching anchor.
+   *
+   * @param {SecurityScopeService} securityScopeService - Scope service instance (reuses caller's connection)
+   * @param {number[]} featureIds - Feature IDs that had rules removed
+   * @return {Promise<void>}
+   */
+  private async deleteAnchorsForFullyUnsecuredFeatures(
+    securityScopeService: SecurityScopeService,
```
mauberti-bc: Not sure what's preferred but put the service in the constructor so it can be accessed on `this`, instead of passing services around?

api/src/services/security-service.ts
```diff
@@ -359,12 +391,29 @@ export class SecurityService extends DBService {
     if (applyRuleIds.length > 0) {
       await this.securityRepository.applySecurityRulesToSubmissionFeatures(submissionFeatureIds, applyRuleIds);
     }
+
+    // Update security scope anchors after rule mutations
+    const securityScopeService = new SecurityScopeService(this.connection);
+
+    if (removeRuleIds.length > 0) {
```
mauberti-bc: This is repeated
```
if (removeRuleIds.length > 0) {
      await this.deleteAnchorsForFullyUnsecuredFeatures(securityScopeService, submissionFeatureIds);
    }

    if (applyRuleIds.length > 0) {
      await securityScopeService.triggerAnchorComputationForSubmission(submissionId);
    }
```

although the duplicate callsite uses `if (applyRuleIds?.length)`. Create a reusable method like `refreshSecurityScopeAnchors(submissionId, submissionFeatureIds)`?

api/src/services/security-service.ts
```diff
@@ -330,18 +331,49 @@ export class SecurityService extends DBService {
...
+  private async deleteAnchorsForFullyUnsecuredFeatures(
+    securityScopeService: SecurityScopeService,
+    featureIds: number[]
+  ): Promise<void> {
+    const remainingRules = await this.securityRepository.getSecurityRulesForSubmissionFeatures(featureIds);
+    const stillSecuredIds = new Set(remainingRules.map((r) => r.submission_feature_id));
+    const fullyUnsecuredIds = featureIds.filter((id) => !stillSecuredIds.has(id));
+
+    if (fullyUnsecuredIds.length > 0) {
+      await securityScopeService.deleteAnchorsForFeatures(fullyUnsecuredIds);
```
mauberti-bc: Is it necessary to call `deleteAnchorsForFeatures`, or should we just let the refresh pgboss job delete old anchors? 

Treating the `security_scope_anchor` table as the source-of-truth for which secured features a team is authorized to access:
* Scenario: A system admin is revoking your access to feature A and is instead giving you access to feature A's direct child, which is feature B.
* You are authorized to access feature B
* When we delete feature A as an anchor, unless there are other security scopes giving you access, you will immediately lose access to A and all of its children.
* You will regain access to feature B once the anchor refresh job completes
* If the anchor refresh job is delayed, you will lose access for some amount of time, despite being authorized.

It seems maybe easier to understand and reason about if we say that the pgboss job "owns" the anchor table, and all changes to it come from the pgboss job?

api/src/repositories/authorization/security-scope-repository.ts
```diff
@@ -0,0 +1,395 @@
...
+   *
+   * Uses a server-side cursor to avoid materializing the full result set in
+   * memory. Wildcard scopes (`urn:*:*:*`) can match millions of secured features;
+   * batching prevents memory/WAL/lock pressure from a single massive INSERT.
+   * Each batch inserts via `unnest` (two parameters regardless of batch size)
+   * rather than a multi-row VALUES clause.
+   *
+   * @param securityScopeId UUID of the security scope to compute anchors for
+   */
+  async computeAnchorsForScope(securityScopeId: string): Promise<void> {
```
mauberti-bc: Ensure we are only including features from submitted features that are "active". If I apply security to a feature while reviewing the submission, before I have accepted the `submission_upload`, that should not affect the anchors.

mauberti-bc (changes requested):
Wonderful! Main points are:
- Ensure that if a policy statement references a specific `submission_feature_id`, that feature is included as an anchor, eg. `urn::*::*::10`
- Ensure we're only including features that are "ready" to be included (`record_effective_date` of the feature?), so that we're not recalculating anchors while adding security to features that are inaccessible/unsearchable.

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: fixed less distraction

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: thanks

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: The walk-up is used to determine "if any ancestors are also anchor candidates", if there are, the descendants are pruned from the anchor candidate set.  Candidates are derived by the urn.  With a narrow urn such as urn::*::*::55, there is only 1 candidate in the set, therefore ancestors will not be found in the anchor candidate set, resulting in the 1 anchor insert for urn::*::*::55,.  That said the was a bug in the pruning SQL which is now fixed.

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: resolved

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: addressed, with a refactor for mutation hook logic to the anchor compute job.

api/src/services/security-service.ts
dwhogan: dried up the code

api/src/services/security-service.ts
dwhogan: refactored to job

api/src/services/security-service.ts
dwhogan: pgboss job now "owns" the anchor table, much cleaner conceptually and less logic spread in hooks

api/src/repositories/authorization/security-scope-repository.ts
dwhogan: sql now consider submission_upload_status

api/src/repositories/search-feature-repository.ts
```diff
@@ -361,11 +367,37 @@ export class SearchFeatureRepository extends BaseRepository {
...
+            UNION ALL
+            SELECT p.submission_feature_id, p.parent_submission_feature_id
+            FROM submission_feature p
+            JOIN ancestors a ON p.submission_feature_id = a.parent_submission_feature_id
+          )
+          SELECT EXISTS (
+            SELECT 1
+            FROM ancestors a
+            INNER JOIN submission_feature_security sfs ON sfs.submission_feature_id = a.ancestor_id
+            WHERE sfs.record_end_date IS NULL
```
mauberti-bc, line 385: Add filter for `WHERE sfs.record_effective_date > now()`, so we're excluding features that have been ingested but not yet reviewed?

app/src/features/search/result/layout/table/SearchResultTableLayout.tsx
```diff
@@ -28,6 +30,27 @@ export const SearchResultTableLayout = ({
...
+        sortable: false,
+        filterable: false,
+        disableColumnMenu: true,
+        renderCell: (params: GridCellParams) => {
+          if (!params.value) {
+            return null;
+          }
+          return (
+            <Stack flexDirection="row" alignItems="center" color="error.main" height="100%" gap={1}>
+              <Icon path={mdiLock} size={0.75} />
```
mauberti-bc, line 46: Wrap icon in a `Box` so it doesn't shrink, and remove Stack/Typography (show only the icon): 

Image

mauberti-bc (changes requested):
Awesome. All search queries for features should only include features where `record_effective_date`, and we probably want to add a DB index for that.

Also see 3 SonarQube issues

mauberti-bc (dismissed):
👍

mauberti-bc (approved):
👍

PR #392, opened 2026-03-24
SIMSBIOHUB-924: Extract Feature Properties

mauberti-bc:
## Links to Jira Tickets

* https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-924

## Description of Changes

* Extracts feature properties using session-scoped temp tables 
* Aims to keep intermediate records in the database rather than API memory
* Feature extraction job executes the following steps:
    * Hard deletes `submission_feature_property_*` records for the given `submissionUploadId`, for idempotency
    * Initializes temp database tables that hold intermediate values, resolving foreign keys.
    * Expands features' `data.properties` into temp table staging rows
    * Creates a set of temp tables for mapping each property to a `feature_type_property_id`
    * Uses the temp tables to validate properties, and writes any errors to a temp error table
    * Queries the temp error table for the API, then throws an `IngestionValidationError` if error count > 0
    * Sets `parent_submission_feature_id` using `data.properties.id` and `data.properties.parent`
    * Inserts into `submission_feature_feature` using `data.properties.content` and `data.properties.id`
    * Inserts into `submission_feature_artifact`, to link `file` and `report` features to their associated artifacts
* Splits `submission_feature_property_timestamp.value` into `date_value` and `time_value`
* Replaces `array` property_feature_type with an `allow_multiple` boolean field, so we can validate the scalar type (eg. `number[]`)
* Removes dead code related to old ingestion flow
* Adds a `database/procedure` file for validating that an object is GeoJSON
* Updates `knexfile` to run procedures in all environments

Note: we cannot run mulitple SQL commands (eg. 2 `CREATE TABLE` statements in a single prepared statement), so the `submission-feature-property-ingestion-service` calls `await this.connection.sql` many times. Ideas for doing this more cleanly would be worthwhile!

## Testing Notes

- After merging https://github.com/bcgov/biohubbc-platform/pull/393, we can test this with >1Gb tarballs.
- Upload tarball to Create Submission page
- Confirm that `submission_feature_property_*` tables have data


api/src/repositories/submission-repository.ts
```diff
@@ -477,91 +477,97 @@ export class SubmissionRepository extends BaseRepository {
...
    *
-   * @param {number} submissionId The submission ID.
-   * @return {Promise<void>}
+   * @param {string} submissionUploadId
+   * @param {string[]} sourceIds
+   * @return {Promise<Array<{ source_id: string; submission_feature_id: number }>>}
    * @memberof SubmissionRepository
    */
-  async deleteSubmissionFeatures(submissionId: number): Promise<void> {
+  async getSubmissionFeatureIdMapBySourceIds(
```
dwhogan: getSubmissionFeatureIdMapBySourceIds has no call sites

api/src/services/ingestion/submission-feature-property-ingestion-service.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    // Phase 10: relationships from `data.content`.
+    await Promise.all([
+      this.submissionFeaturePropertyIngestionRepository.insertFeatureRelationshipsBySubmissionUploadId(
+        submissionUploadId
+      ),
+      this.submissionFeaturePropertyIngestionRepository.recordReferenceErrorsBySubmissionUploadId(submissionUploadId)
+    ]);
+
+    // Phase 11: fail-at-end with aggregated diagnostics after full-upload processing.
```
dwhogan: with an all-or-nothing to the db commit via the `IngestionValidationError` throw, could phase 11 move up before row inserts?

dwhogan (changes requested):

PR #394, opened 2026-03-26
SIMSBIOHUB-938: Sanitize ClamAV results

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-938

## Description of Changes

- Default ClamAV behaviour is to skip large files and report success, as if the file was scanned
- ClamAV scan job now throws an error if file size exceeds `CLAMAV_MAX_SCAN_SIZE`
- Sanitizes ClamAV results: 
    - Fixes a bug where bad Unicode was failing to insert into `artifact_security_scan.results` because `jsonb` does not support that character, causing ClamAV job to fail

## Testing Notes

- You can set CLAMAV_MAX_SCAN_SIZE to be very large (eg. 10gb), allowing big files to be uploaded, but they won't be scanned


dylanrogowsky-oxd (commented):
Should `CLAMAV_MAX_SCAN_SIZE` be added to:
- infrastructure/api/templates/deployment.yaml (add env var injection)
- infrastructure/queue/templates/deployment.yaml (add env var injection)
- infrastructure/api/values.yaml (define value, e.g. app.clamav.maxScanSize)
- infrastructure/queue/values.yaml (define value, e.g. app.clamav.maxScanSize)

Umbrella chart (where env-specific values are actually set):
- infrastructure/biohub-platform/values.yaml (default/global umbrella value)
- infrastructure/biohub-platform/values-dev.yaml
- infrastructure/biohub-platform/values-test.yaml
- infrastructure/biohub-platform/values-prod.yaml
- infrastructure/biohub-platform/values-pr.yaml

dylanrogowsky-oxd (approved):

PR #403, opened 2026-03-30
SIMSBIOHUB-947: Team membership update should support removing members

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-947](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-947)

## Description of Changes

When editing a team, the update only appends members — there is no way to remove a member. The edit dialog starts with an empty member list, so saving without re-selecting every member either does nothing or silently loses the intended state. We want the edit flow to pre-populate current members and sync subtractively: remove absent, add new, keep existing.

### Backend

1. **Subtractive sync**
   - WHEN `system_user_ids` is provided in the update payload
   - THEN members absent from the list are removed, new IDs are added, and existing matches are unchanged

2. **Metadata-only update**
   - WHEN `system_user_ids` is omitted from the payload
   - THEN member sync is skipped and only name/description are updated

### Frontend

3. **Pre-populate members on edit**
   - WHEN the user opens the edit team dialog
   - THEN current members are fetched and shown in the form

4. **Fetch failure blocks dialog**
   - WHEN the member fetch fails
   - THEN an error dialog is shown and the edit dialog does not open

## Testing Notes

1. Go to admin/policies
2. Edit a team
3. Remove members


app/src/features/admin/policies/components/TeamsContainer.tsx
```diff
@@ -377,13 +403,14 @@ export const TeamsContainer = (props: ITeamsContainerProps) => {
         open={openEditTeamDialog}
         dialogSaveButtonLabel="Save"
         component={{
-          element: <AddTeamForm />,
+          element: <AddTeamForm initialUsers={editingTeamMembers} />,
```
mauberti-bc: It's a bit confusing to pass editingTeamMembers to `initialUsers` prop and formik's `initialValues`. AcCleaner approach might be to modify the form interface to this:
```
export interface IAddTeamFormValues {
  name: string;
  description: string;
  system_users: IAvailableUser[];
}
```

then set system_users with `editingTeamMembers`:

```
const getEditTeamInitialValues = (): IAddTeamFormValues => {
    if (!editingTeam) {
      return AddTeamFormInitialValues;
    }
    return {
      name: editingTeam.name,
      description: editingTeam.description || '',
      system_users: editingTeamMembers
    };
  };
```

and then extract the `system_user_ids` in the submit handler: 

```
await biohubApi.teams.updateTeam(editingTeam.team_id, {
        name: values.name,
        description: values.description ?? null,
        system_user_ids: values.system_users.map((user) => user.system_user_id)
      });
```

mauberti-bc (changes requested):

app/src/features/admin/policies/components/TeamsContainer.tsx
dwhogan: fix made, thanks

mauberti-bc (approved):
👍

PR #408, opened 2026-04-09
SIMSBIOHUB-953: Store Artifact for Downloads

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-953](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-953)

## Description of Changes

Associate downloads with artifacts by creating an artifact record at download creation time and linking it via new download, export, and export-artifact tables.

Data Model

1. Create table download_artifact

download_id (FK to download)
artifact_id (FK to artifact)
format (e.g. parquet)
We will only support parquet, but we could theoretically support other file types where parquet isn't ideal, so worth adding the column (e.g. exporting spatial data to tif, zarr, etc.)

2. Create table download_export

download_export_id
download_id (FK to download)
format (e.g. csv)
status

3. Create table download_export_artifact

download_export_id (FK to download_export)
artifact_id (FK to artifact)
chunk_id

I'm thinking this might be useful for restarting failed export jobs from a checkpoint? if we export all parquet chunks in order, and we have chunk_id = 10, we can then restart exporting at chunk 11.

**Acceptance Criteria**

1. Create and link artifact on download creation

WHEN a download is requested
THEN create a new artifact with status = 'pending'
This let's us create the artifact record without the file existing yet
AND join the artifact to the download

2. Constraints

Add a unique constraint on (download_id, artifact_id) WHERE record_end_date is null
Add a unique constraint on (download_export_id, artifact_id) WHERE record_end_date is null

**Notes**
For this ticket's scope, because we currently write CSV fragments, we cannot create a single completed artifact for the final export, so the download-level artifact will remain in pending status.

We'll add the new `download_export` and `download_export_artifact` tables now, then drop the old `download_fragment` tables when we update the API to use the new tables.

## Testing Notes

Backend regression tests only
Run tests (make test, make test-db, make test-sys)


api/src/services/download/download-service.ts
```diff
@@ -37,24 +40,45 @@ export class DownloadService extends DBService {
...
    * @return {Promise<DownloadId>} The created record ID.
    * @memberof DownloadService
    */
   async createDownload(payload: CreateDownload): Promise<DownloadId> {
-    return this.downloadRepository.createDownload(payload);
+    const downloadId = await this.downloadRepository.createDownload(payload);
+
+    const artifact = await this.artifactService.insertArtifact({
+      bucket: getObjectStoreBucketName(),
+      object_key: `downloads/${downloadId.download_id}/download.parquet`,
```
mauberti-bc: suggestion: add timestamp like `download-2026-01-01T000000Z.parquet`, to prevent collisions if a client downloads the file locally (if they already have a `download.parquet` in their downloads folder)

database/src/migrations/20260408224113_download_artifact_tables.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    --------------------------------------------------------------------------------
+    -- DOWNLOAD_ARTIFACT
+    --------------------------------------------------------------------------------
+
+    CREATE TABLE download_artifact (
+      download_artifact_id INTEGER GENERATED ALWAYS AS IDENTITY NOT NULL,
+      download_id UUID NOT NULL,
+      artifact_id UUID NOT NULL,
+      format VARCHAR(50) NOT NULL,
```
mauberti-bc: This `format` looks unnecessary, after some more thought. The _actual_ format of the file belongs to the `artifact` table, and the _requested_ format should belong to the `download` table. 

I'd suggest removing `download_artifact.format` and instead adding `download.format`

mauberti-bc (changes requested):

database/src/migrations/20260408224113_download_artifact_tables.ts
```diff
@@ -0,0 +1,183 @@
...
+  await knex.raw(`
+    SET SEARCH_PATH = biohub, public;
+
+    --------------------------------------------------------------------------------
+    -- DOWNLOAD_ARTIFACT
+    --------------------------------------------------------------------------------
+
+    -- Add requested format to the download table (the artifact table tracks actual file format)
+    ALTER TABLE download ADD COLUMN format VARCHAR(50) NOT NULL DEFAULT 'csv';
+    ALTER TABLE download ALTER COLUMN format DROP DEFAULT;
```
mauberti-bc, line 13: Clever!

mauberti-bc (approved):

PR #409, opened 2026-04-16
SIMSBIOHUB-960: Add Data Request to Tickets (Backend)

mauberti-bc:
## **Links to Jira Tickets**

https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-960

## **Description of Changes**

- Use policy.status as the single source of truth (requested, reviewed, approved, denied)
- Data request status now comes from its linked policy (`data_request.policy_id`)
- Remove `data_request.status`, `data_request_status`, and `request_status` (prioritize simplicity; it was getting complex trying to sync tables)
- Deletes all related code for deprecated database tables
- Simplify services and repositories
- Prevent deleting tickets with active (requested or reviewed) data requests

## **Testing Notes**

- Create a ticket
- Create a data request in the ticket sidebar and verify:
  - `policy` is created with correct `status`
  - `data_request.policy_id` is set correctly
- Verify status is returned via policy join (no derived TS status)
- Attempt to delete a ticket with open data requests (`requested` / `reviewed`) and confirm 400 response
- Run existing repository/service/path tests to confirm updated behavior


api/src/services/data-request-service.ts
```diff
@@ -1,226 +1,178 @@
...
+        description: 'Auto-generated policy for ticket-linked data request',
+        status: 'requested'
+      },
+      [{ effect: PolicyEffect.DENY, submission_feature_urn: 'urn:*:*:*' }]
     );
 
-    return {
-      ...dataRequest,
-      data_request_status: dataRequestStatus
+    const payloadWithIds: CreateDataRequest = {
```
dwhogan, line 164: do we need to link the team to the policy here?

api/src/services/data-request-service.ts
```diff
@@ -1,226 +1,178 @@
...
-   * @return {Promise<Team>} team to use for the data request
-   * @private
-   */
-  private async createTeam(requestedBy: number): Promise<Team> {
-    const teamService = new TeamService(this.connection);
-    const team = await teamService.createTeam({ name: _generateDataRequestTeamName() });
-    const teamMemberService = new TeamMemberService(this.connection);
-    await teamMemberService.createTeamMember({ system_user_id: requestedBy, team_id: team.team_id });
-    return team;
+    await this.dataRequestRepository.deleteDataRequest(dataRequestId);
```
dwhogan, line 197: `if (!dataRequest) {
      throw new HTTP404('Data request not found');
    }`  guard was removed, intentional?

dwhogan (changes requested):
Nice refactor to Single source of truth 👍

api/src/services/data-request-service.ts
mauberti-bc: Yes, the repository method will throw a 500 still. Just trying to keep it simple

api/src/services/data-request-service.ts
mauberti-bc: Thanks, here is the intended outcome (now fixed). When a data request is created,
1. Create a team for the data request, which just gives access for viewing the request
2. Create a team for the ticket, which just gives access for viewing the ticket
3. Create a team for the policy, which defines the actual recipients of the request.

We intentionally create 3 separate teams so admins can give clients appropriate access without side effects.

dwhogan (approved):
👍

PR #411, opened 2026-04-16
SIMSBIOHUB-955: Write Download to Parquet File

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-955](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-955)[SIMSBIOHUB-955](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-955)

## Description of Changes

 Generate a single artifact for each download by writing the dataset to Parquet, partitioned by feature_type, and associate it with the download.

**Acceptance Criteria**

1. Write Parquet artifact for download

GIVEN a download has been created and is associated with an artifact in "pending" status
WHEN the download processing job executes
THEN write the features to a Parquet file partitioned by "feature_type" (see below)
The parquet file may be larger than memory, so we should write row groups in chunks (eg. initialize the parquet file, then write 10k rows at a time until all features are included) 
Parquet column headers / structure should match current .csv behaviour (transforming feature properties to column headers, etc.)
AND persist the Parquet file in object storage
THEN update the parquet file's artifact record's status to completed
AND update the download's status to completed

Example partitioned parquet

```
downloads/{downloadId}/download-{timestamp}.parquet/
  dataset/
    data.parquet
  sample_site/
    data.parquet
  telemetry/
    data.parquet
```

## Testing Notes

The PR effects the download backend Job, therefore run all test (unit and integration)

*I have keep the original CSV branch in the download job as we transition to parquet write and export to CSV.


api/src/repositories/download/download-repository.ts
```diff
@@ -608,4 +634,318 @@ export class DownloadRepository extends BaseRepository {
...
+   * Same purpose as listDownloadFeatureTypesByCartId but uses a Knex subquery
+   * from the search pipeline instead of a cart join.
+   *
+   * @param {Knex.QueryBuilder} searchSubquery - Unexecuted Knex subquery
+   *   returning submission_feature_id rows (from SearchFeatureRepository).
+   * @return {Promise<string[]>} Ordered list of feature type names.
+   * @memberof DownloadRepository
+   */
+  async listDownloadFeatureTypesBySearchQuery(searchSubquery: Knex.QueryBuilder): Promise<string[]> {
+    const FeatureTypeName = z.object({ feature_type_name: z.string() });
```
mauberti-bc: nit: Move to `models/download.ts` if repeated, or just inline it within this.connection.knex to avoid the const?

api/src/repositories/download/download-repository.ts
```diff
@@ -608,4 +634,318 @@ export class DownloadRepository extends BaseRepository {
...
+   * @param {string} cartId - The cart ID (from download.cart_id).
+   * @param {string} featureTypeName - The feature type to stream.
+   * @param {number} [batchSize=5000] - Rows per FETCH.
+   * @yields {BaseFeatureRow[]} Batches of base feature rows.
+   * @memberof DownloadRepository
+   */
+  async *streamFeatureBaseByCartIdAndType(
+    cartId: string,
+    featureTypeName: string,
+    batchSize = 5000
```
mauberti-bc: Suggestion: Move all batch size constants to env vars/Helm, so that we can tune these without codebase changes. Establish a prefix like `DOWNLOAD_*`, such as `DOWNLOAD_FEATURE_MAX_ROWS`, `DOWNLOAD_FEATURE_MAX_BYTES`. We can do the same for `EXPORT_*` down the road, too.

It will be useful to see all the memory-impacting config values together in Helm.

mauberti-bc (changes requested):

api/src/repositories/download/download-repository.ts
dwhogan: made fix for batchSize and made not for future tuneable env vars

api/src/repositories/download/download-repository.ts
dwhogan: fixed, thanks

mauberti-bc (dismissed):

mauberti-bc (approved):
👍

PR #418, opened 2026-04-21
SIMSBIOHUB-965: Parquet Download UI

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-965](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-965)[SIMSBIOHUB-965](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-965)

## Description of Changes

1. **Backend:** `POST /api/download` creates Parquet-only downloads; the pg-boss worker writes one Parquet file per feature type to S3 at `downloads/{downloadId}/{featureTypeName}/data.parquet`. No fragment ZIPs, no CSV runtime path, no backward compatibility.

2. **Frontend:** every download row renders as a minimal card — create date + status chip only. No action buttons, no feature count, no extra copy.

### Acceptance Criteria

1. **Backend: Parquet-only download job**
   - `POST /api/download` sets `download.format = 'parquet'` and enqueues a pg-boss job on `process-download`. No `artifact` / `download_artifact` rows at request time — the worker owns all artifact writes.
   - The worker writes **one Parquet file per feature type** to S3 at key `downloads/{downloadId}/{featureTypeName}/data.parquet` (deterministic across retries, no multi-part splits).
   - The worker writes one `artifact` row per Parquet file (`format='parquet'`, `artifact_status='uploaded'`, real `byte_size` + streaming `checksum_sha256` + `uploaded_at`) and one `download_artifact` join row per file.
   - Spatial feature types get file-level GeoParquet 1.0 metadata (WKB geometry column, CRS, geometry types). Non-spatial feature types (e.g. `dataset`) are plain Parquet.
   - `artifact_key` attachments map to a Parquet `filePath` column only — binary delivery is out of scope.
   - `download.download_status` transitions `pending` → `processing` → `ready`. `ready` means "Parquet snapshot is on S3" — no fragment ZIPs, no user-downloadable output. No enum or schema change.
   - CSV/ZIP runtime paths (`csv-utils.ts`, `archiver`-based CSV bundling, `processFragment`'s CSV-to-zip assembly, `download_fragment_repository`) have no runtime caller but stay in the repo so the follow-up can reuse `archiver` for Parquet-in-ZIP packing.
   - Retries are idempotent: deterministic S3 keys + `artifact UNIQUE (bucket, object_key)` + `download_artifact ON CONFLICT … WHERE record_end_date IS NULL DO NOTHING`; S3 overwrite semantics make partial-success retries safe.
   - Backend tests cover: `format='parquet'` on create with no artifact rows; worker produces one Parquet + one `artifact` + one `download_artifact` per feature type; S3 key shape; GeoParquet metadata on spatial types; status transitions; retry produces identical keys.

2. **Backend integration tests: inspect the actual Parquet objects on S3**
   - Download each uploaded object from S3 (MinIO in CI), open it with a Parquet reader, and assert against the file — not just metadata.
   - S3 object keys match **exactly** `downloads/{downloadId}/{featureTypeName}/data.parquet` — no timestamps, no suffixes, one key per feature type.
   - Parquet schema per feature type matches the expected columns (typed properties + `uuid` + `parent_uuid`). Column types match the feature_property_type (string/number/timestamp/code→string/taxon→string/spatial→binary WKB).
   - Spatial feature types carry file-level **GeoParquet 1.0** key-value metadata: the `geo` key is present, parses as JSON, `version === '1.0.0'`, names the primary geometry column, geometry column encoding is `WKB`, CRS is recorded, geometry types are listed. Non-spatial feature types have **no** `geo` key.
   - `artifact_key` properties appear in the Parquet as the S3 key string in a `filePath` column. No binary payload is embedded.
   - The `artifact.checksum_sha256` field matches SHA-256 re-hashed over the downloaded Parquet bytes. `byte_size` matches the actual byte length. `uploaded_at` is non-null.
   - Retry: running the `process-download` job a second time for the same `downloadId` leaves the S3 key set and `artifact` / `download_artifact` row counts unchanged, and the Parquet bytes still round-trip identically.

3. **Frontend: placeholder rendering for all downloads**
   - `DownloadFeatureCard` renders **only** the create date + status chip (the existing Row 1). No feature count, no download buttons, no "Download all N parts", no individual-parts collapse, no extra copy. No branching on `format`.
   - No new hooks in `useDownloadApi.ts`.
   - Frontend tests (`DownloadSidebarCart.test.tsx`, `DownloadSidebarDownloads.test.tsx`, `DownloadFeatureCard.test.tsx`) cover the placeholder rendering.

## Testing Notes


api/src/repositories/download/download-repository.ts
```diff
@@ -115,28 +115,21 @@ export class DownloadRepository extends BaseRepository {
   /**
    * Link an artifact to a download via the download_artifact join table.
    *
-   * Downloads are tracked as formal artifacts from the moment they're requested.
-   * The artifact is created as 'pending' (no file yet) and linked here.
-   *
-   * @param {string} downloadId - The download ID.
```
mauberti-bc, line 121: Suggestion: keep all the `@params`, `@returns`, etc. for the jsdocs. Applies to all methods where comments were updated.

api/src/repositories/download/download-repository.ts
```diff
@@ -170,6 +163,26 @@ export class DownloadRepository extends BaseRepository {
...
+   *
+   * `get*` throws on missing row (codebase convention — companion to `findDownloadById`).
+   *
+   * @throws {ApiExecuteSQLError} when no download matches the given ID.
+   */
+  async getDownloadById(downloadId: string): Promise<DownloadRecord> {
+    const download = await this.findDownloadById(downloadId);
+
+    if (!download) {
+      throw new ApiExecuteSQLError('Download not found', [
```
mauberti-bc: ApiNotFoundError?

api/src/paths/download/index.ts
```diff
@@ -220,7 +220,7 @@ export function createDownload(): RequestHandler {
       }
 
       // Store filters on download — features re-derived at pipeline time via search re-run
-      const downloadId = await downloadService.createDownload({ filters, format: 'csv' });
+      const downloadId = await downloadService.createDownload({ filters, format: 'parquet' });
```
mauberti-bc, line 223: Cart checkout still creates downloads with format: 'csv' but generates a parquet. Downloads should always be parquet (configurable via `format` but we should only be using parquet)

```
const downloadId = await this.downloadService.createDownload({
      fragmentSizeBytes,
      cartId,
      format: 'csv'
    });
```

api/src/queue/pg-boss-service.ts
```diff
@@ -11,26 +11,40 @@ export const pgBossDependencies = {
...
+const setBoss = (value: PgBoss | null): void => {
+  globalRef[GLOBAL_KEY] = value;
+};
 
 /**
  * Initialize the pg-boss singleton instance.
  *
  * @return {*}  {Promise<PgBoss>}
  */
 export const initPgBoss = async (): Promise<PgBoss> => {
```
mauberti-bc, line 47: Possible race condition: if `initPgBoss` is called twice at the same time, there's no lock/guard to ensure a single instance?

mauberti-bc (changes requested):

api/src/repositories/download/download-repository.ts
dwhogan: thanks for catching my refactor miss!

api/src/repositories/download/download-repository.ts
dwhogan: better, thanks!

api/src/paths/download/index.ts
dwhogan: good eye, thanks

api/src/queue/pg-boss-service.ts
dwhogan: added guard, thanks

mauberti-bc (approved):
👍

PR #419, opened 2026-04-21
SIMSBIOHUB-969: Assign Tickets to Users 

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-969

## Description of Changes

- Adds explicit ticket assignee management so tickets can be assigned to individual users.
- Introduces an assignee lifecycle status model (`requested`, `started`, `blocked`, `resolved`) across UI/API/DB.
- Renames the terminal assignee status from `done` to `resolved` and includes database enum migration handling.
- Enables status updates for assignees from the ticket sidebar with role-aware controls.
- Keeps a direct “Accept” action for the current assignee to move from `requested` to `started`.
- Allows system administrators to remove assignees.
- Improves assignee context menu UX by grouping status actions separately from destructive actions (`Delete`) with visual separation.
- Enhances shared dropdown/context menu components to support grouped menu items and icons for clearer action presentation.

## Testing Notes

- Create a ticket, invite assignees, remove assignees, change status


app/src/features/admin/tickets/detail/sidebar/TicketSidebarSystemUserCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+      [TicketSystemUserStatus, (typeof TICKET_SYSTEM_USER_STATUS_PRESENTATION)[TicketSystemUserStatus]]
+    >
+  ).map(([value, presentation]) => ({
+    value,
+    label: presentation.label,
+    icon: presentation.icon
+  }));
+
+  const displayLabel = ticketSystemUser.system_user.display_name ?? ticketSystemUser.system_user.user_identifier;
+  const canUpdateStatus = isSystemAdmin || isCurrentUserAssignment;
```
dwhogan: Maybe check up stream TicketSystemUserForm which hits api.teams.getAvailableUsers is expected behaviour, both PATCH and DELETE routes both gate on `[SYSTEM_ROLE.SYSTEM_ADMIN]` anyway.

database/src/migrations/20260421120000_ticket_system_user.ts
```diff
@@ -0,0 +1,129 @@
...
+      CONSTRAINT ticket_system_user_system_user_fk FOREIGN KEY (system_user_id) REFERENCES "system_user"(system_user_id)
+    );
+
+    ----------------------------------------------------------------------------------------
+    -- 3. Create lookup indexes and enforce one active assignment per ticket/user.
+    ----------------------------------------------------------------------------------------
+    CREATE INDEX ticket_system_user_ticket_idx ON ticket_system_user(ticket_id);
+    CREATE INDEX ticket_system_user_system_user_idx ON ticket_system_user(system_user_id);
+    CREATE UNIQUE INDEX ticket_system_user_nuk1
+      ON ticket_system_user(ticket_id, system_user_id, (record_end_date IS NULL))
```
dwhogan: wondering if we should reconsider the redundant (record_end_date IS NULL)

api/src/services/ticket-system-user-service.ts
```diff
@@ -0,0 +1,127 @@
...
+    const existing = await this.ticketSystemUserRepository.getTicketSystemUserByTicketAndSystemUserId(
+      ticketId,
+      ticketSystemUserId
+    );
+
+    if (!existing) {
+      throw new HTTP404('Ticket system user not found');
+    }
+
+    if (existing.status === payload.status) {
```
dwhogan, line 94: The context menu already disables the current status (`disabled: statusOption.value === ticketSystemUser.status`), so this branch is mostly unreachable. But PATCH conventionally is idempotent — when it does get hit (double-click race, stale UI state, retry) the user sees an error snackbar for what is effectively a no-op.

**Suggestion:** Return `existing` unchanged instead of throwing.

```suggestion
    if (existing.status === payload.status) {
      return existing;
    }
```

dwhogan (changes requested):
Clean domain separation👍

app/src/features/admin/tickets/detail/sidebar/TicketSidebarSystemUserCard.tsx
mauberti-bc: Good point. I removed the frontend guard since the whole page is gated.

database/src/migrations/20260421120000_ticket_system_user.ts
mauberti-bc: Agreed, switched to:
```
CREATE UNIQUE INDEX ticket_system_user_nuk1
      ON ticket_system_user(ticket_id, system_user_id)
      WHERE record_end_date IS NULL;
```

api/src/services/ticket-system-user-service.ts
mauberti-bc: Fixed, thanks

dwhogan (approved):
👍

PR #420, opened 2026-04-23
SIMSBIOHUB-954: Export Parquet Fragments

dwhogan:
## Links to Jira Tickets

- [SIMSBIOHUB-954](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-954)

## Description of Changes

Generate CSV exports for a download by reading the Parquet artifact and writing one logical CSV per feature type. Track each export's lifecycle via `download_export` and `download_export_artifact`.

The Parquet artifact is the "master" copy — optimized for storage efficiency and downstream processing. CSV exports are **user-triggered transformations** of that master, produced on demand when an authenticated user requests a tabular view of an existing download.

This ticket ships a single export shape:

- **Per-feature-type (star):** one logical CSV per feature type, each containing that type's own properties plus system columns (`uuid`, `parent_uuid`). Consumers join across types client-side via `parent_uuid = uuid`.

Each logical CSV is split into **size-bounded chunks** so users can reliably download over constrained connections and reconstruct the full CSV client-side by concatenating chunks in `chunk_id` order. Chunks are grouped into part-zips for delivery.

### 1. Migration: add `max_part_size_bytes` and `mode` to `download_export`

Add two columns to `biohub.download_export`:

- `max_part_size_bytes bigint not null default 524288000` (500 MB) — per-export part-size threshold, independent of `download.fragment_size_bytes`.
- `mode varchar(32) not null default 'per_feature_type'` with `check (mode in ('per_feature_type', 'denormalized'))` — export-shape discriminator. This ticket always writes `'per_feature_type'`; the denormalized follow-up starts writing `'denormalized'`.

Extend `DownloadExportRecord` in `api/src/models/download-export.ts` with `max_part_size_bytes: z.string()` and `mode: z.enum(['per_feature_type', 'denormalized'])`.

### 2. Request a CSV export

WHEN an authenticated user POSTs to the export endpoint for a download they can access (via `download_team` membership)
AND the underlying download is `"ready"` (Parquet finalized)
THEN create a `download_export` record with `format = "csv"`, `mode = "per_feature_type"`, and `max_part_size_bytes = payload.max_part_size_bytes ?? 524288000`
AND set the export status to `"pending"`
AND accept a request body of shape `{ "max_part_size_bytes"?: number }` (optional). When omitted, defaults to 500 MB. Value must be a positive integer ≤ 5 GB; out-of-range values return 400.
AND publish a `processExportJob` (singleton on `download_export_id`)
AND return the `download_export` identifier

Anonymous UUID holders cannot trigger exports; they must authenticate and claim the download first.

**Note on export shape:** this ticket ships per-feature-type (star) output as the only shape. The `mode` column added by AC 1 carries the discriminator with a `check` constraint bounding the set to `'per_feature_type' | 'denormalized'`. This ticket always writes `'per_feature_type'`; the denormalized follow-up starts writing `'denormalized'`.

### 3. pg-boss infrastructure: dedicated queue for exports

WHEN this ticket lands
THEN a new pg-boss queue and handler are registered, sibling to (not branching off of) the existing `process-download` queue:

- **Queue name:** `process-download-export` (and paired DLQ `process-download-export-failed`) — registered in `api/src/queue/worker.ts` alongside the existing queues. Same retry pattern as `process-download` (retryLimit 3, retryDelay 60s, retryBackoff true). Expiration: 1 hour (exports are bounded by one Parquet read + zip write per feature type — shorter than Parquet's 4-hour budget).
- **Job handler file:** `api/src/queue/jobs/process-download-export-job.ts`. Exports two functions: `processDownloadExportJobHandler` (main) and `processDownloadExportFailedHandler` (DLQ). Handler body is a thin orchestrator that delegates to a `DownloadExportPipelineService` — matches the `process-download-job.ts` / `DownloadPipelineService` split.
- **Handler status transitions:** `pending` → `processing` → `ready` (main) or `pending`/`processing` → `failed` (DLQ only, after all retries exhausted). The main handler never writes `failed` — terminal failure is owned by the DLQ handler.
- **Input shape:** `{ downloadExportId: string }`. No union, no discriminator.
- **Publisher method:** `publisher.publishProcessDownloadExportJob(connection, { downloadExportId })` in `api/src/queue/publisher.ts`. Same signature style as `publishProcessDownloadJob`, called from the POST export handler inside its transaction.
- **Singleton key:** `export-${downloadExportId}`. Prevents duplicate jobs for the same export from queueing concurrently. Queue uses `policy: 'short'` (same as `PROCESS_SUBMISSION_FEATURES`) so the singleton key is enforced while a job is queued, not just while active.
- **Retry budget isolation:** exports have their own queue, so a storm of export retries cannot starve Parquet-write jobs and vice versa. Matches the lead's existing convention (5 per-domain queues, each with its own DLQ).
- **Tests:** unit tests for the handler mirror `process-download-job.test.ts` — stub the pipeline service, construct `PgBoss.Job` fixtures, assert status transitions + DLQ behaviour.

### 4. Export CSV content from Parquet

WHEN the export processing job executes
THEN read the Parquet dataset partitioned by `"feature_type"` from object storage using **`ParquetReader.openS3(s3Client, { Bucket, Key })`** so that row groups are fetched on-demand via S3 range GETs — the full Parquet file is NEVER buffered into memory
AND iterate rows via the reader's async iterator (`for await (const row of reader)`) or `getCursor()` so peak memory is bounded to a single row group (Parquet default ≤ 128 MB) regardless of the Parquet file's total size
AND emit one logical CSV per feature type present in the download; each contains the feature type's own properties plus the system columns `uuid` and `parent_uuid`.

AND apply the same value transformations as the legacy CSV pipeline:

- Spatial properties → `decimalLatitude` / `decimalLongitude` columns, first-point only
- `artifact_key` → `filePath` column, valued as the zip-relative path `files{N}/{submission_feature_id}_{original_filename}` (where `N` is the part index of the zip that owns the chunk)
- CSV escaping per RFC 4180
- Array / object properties flattened per existing `csv-utils.ts` semantics

AND order rows within each feature type by `uuid` for reproducibility across retries.

**De-risked for OOM.** Feature types write their CSV chunks sequentially (one feature type fully processed before the next begins) so at most one Parquet reader + one archive pipeline are alive at a time. Peak heap per job ≈ one row group (~128 MB) + archiver compression buffer + 64 MB multipart upload buffer ≈ 200–300 MB — well within typical container limits, and independent of total dataset size. Tests MUST cover this: an integration test exports a Parquet file deliberately larger than available heap (e.g. 512 MB fixture) and asserts the job completes without OOM.

### 5. Split logical CSVs into size-bounded chunks

WHEN CSV output exceeds `download_export.max_part_size_bytes`
THEN split each logical CSV across multiple chunk files at a row boundary
AND include the CSV header on the first chunk only; subsequent chunks are header-less so byte-concatenation in `chunk_id` order reconstructs a valid CSV
AND assign each chunk a sequential `chunk_id` within its logical CSV (1..N)

The threshold is **per-export**, not per-download — it lives on the `download_export` row (AC 1), so the same download can be exported multiple times at different part sizes. Default 500 MB, matching the legacy fragment pipeline. Users supply a smaller value at export-creation time via AC 2's `max_part_size_bytes` payload field.

Each logical CSV lives in its own folder inside the zip, named after the feature type. Chunks live under that folder as `chunk{chunk_id}.csv`:

- `{feature_type}/chunk{chunk_id}.csv` (e.g. `observation/chunk1.csv`, `observation/chunk2.csv`)

Binary attachments referenced by `artifact_key` properties are bundled under a sibling `files{N}/` folder at the zip root, where `N` is this zip's part index (1..). Binaries are named `{submission_feature_id}_{original_filename}`. The `filePath` column in the CSV points to this zip-relative path (e.g. `files1/10042_photo-01.jpg`), so a user opening the CSV has a working link into the extracted folder. Using `files{N}/` (not a shared `files/`) avoids any chance of filename collision between parts and mirrors the legacy `-part-` pattern.

### 6. Package chunks into part-zips

WHEN chunks are written
THEN group them into one zip per part, named `biohub-{exportId}-part-{N}.zip`
AND assign each zip a sequential part index `N` starting at 1

Part grouping rule: **zip N contains `chunk{N}.csv` from every logical CSV that has a chunk at that index**, plus the `files{N}/` subset referenced by that zip's rows.

- A small export (every logical CSV fits in one chunk) ships as a single `biohub-{exportId}-part-1.zip`.
- An export with uneven chunk counts (e.g. observation has 3 chunks, dataset has 1) ships as three zips: `part-1.zip` contains all three types' `chunk1.csv`, `part-2.zip` contains `observation/chunk2.csv` only, `part-3.zip` contains `observation/chunk3.csv` only. The `dataset/` folder is absent from later parts.

All parts extract to the **same root folder** `biohub-export-{exportId}/`. Because chunk filenames are unique (`chunk1.csv`, `chunk2.csv`, …) the extracted trees merge cleanly, producing the layout in the Example Export Output below.

### 7. Link part-zip artifacts to export

WHEN each part-zip is written to object storage
THEN create an `artifact` record for the part-zip with `format = "zip"` (physical file format on disk, distinct from `download_export.format = "csv"` which is the requested export format)
AND set the artifact's real `byte_size`, streaming `checksum_sha256`, `uploaded_at`, and `artifact_status = "uploaded"` at write time (same pattern the Parquet pipeline uses)
AND join the artifact to the `download_export` via a `download_export_artifact` row
AND persist the zip's part index on the join row (reusing `chunk_id` to mean the part index across the export)

**S3 key is deterministic per `(downloadId, exportId, N)`:**

```
downloads/{downloadId}/exports/{exportId}/biohub-{exportId}-part-{N}.zip
```

Retries overwrite the same key idempotently — matches the Parquet pipeline's approach (`downloads/{downloadId}/{featureTypeName}/data.parquet`). Safety relies on:

- S3 multipart upload atomicity — partial uploads abort and leave nothing behind; successful uploads replace prior content in a single commit.
- `artifact UNIQUE (bucket, object_key)` with `ON CONFLICT … DO NOTHING` on `download_artifact` / `download_export_artifact` so retried DB inserts don't double-link.

No per-attempt subkeys, no retry counters, no orphan accumulation.

### 8. Complete or fail the export

WHEN all part-zips have been written successfully
THEN update all associated artifact records to status `"uploaded"`
AND update the `download_export` status to `"ready"`
AND make each part-zip retrievable via its own presigned URL (one URL per `download_export_artifact` row)

IF any feature type or part-zip fails
THEN set the `download_export` status to `"failed"` (whole-export-fails — no partial delivery)
AND record the failure context in `download_export.error_message`

(Any part-zips that were successfully uploaded before the failure stay on S3 under their deterministic keys. They're not garbage — a subsequent retry overwrites them with fresh content. No cleanup pass needed.)

### 9. List exports for a download

WHEN an authenticated user GETs `/api/download/:downloadId/export`
THEN return all `download_export` rows for that download (id, format, mode, status, max_part_size_bytes, started_at, completed_at, error_message)

### 10. Retrieve an export's detail

WHEN an authenticated user GETs `/api/download-export/:exportId`
THEN return the export's metadata
AND when `status = "ready"`, include one presigned URL per part-zip in `chunk_id` (part index) order
AND surface the part count so clients can pre-size progress indicators

Endpoints in ACs 9 and 10 both enforce `download_team` membership; unauthorized requests return 401/403.

### 11. Deprecate legacy fragment-feature usage

WHEN the new export pipeline runs
THEN it MUST NOT read from or write to `download_fragment_feature`

The legacy CSV-zip pipeline may continue to exist for backward compatibility, but the new export path must be fully independent of `download_fragment_feature`. DDL removal is a follow-up ticket.

### 12. Frontend: trigger a CSV export from the download card

WHEN a download is `"ready"` and the current user is a member of a `download_team` that can access it
THEN `DownloadFeatureCard` renders an **"Export CSV"** action (icon button in the card's right-hand meta area, adjacent to the status chip)
AND clicking it POSTs to `/api/download/:downloadId/export` with an empty body (accepts server default of 500 MB chunks), then transitions the card into the in-flight state (AC 13)

There is no dialog in this ticket — per-feature-type is the only shape, it always exports every property of every feature type, and there's nothing for the user to configure. The 4xx error surface is rendered inline on the card (toast or inline banner; decide during implementation). The `max_part_size_bytes` payload field is wired through the hook but not exposed in the card UI — it's reserved for future UX (e.g. a "part size" selector alongside the export action).

Anonymous UUID holders (no `download_team` membership) do not see the action — this matches AC 2's server-side guard.

The denormalized follow-up (`docs/csv-export-denormalized-mode/ticket.md`) adds the dialog, the mode toggle, the column-selection tree, and the schema endpoint that feeds it — this ticket deliberately leaves all of that out.

### 13. Frontend: render export state on the download card

WHEN a download has one or more `download_export` rows
THEN `DownloadFeatureCard` expands to show an **Exports** section listing each export in reverse-chronological order (most recent first)

Per-export row state, driven by `download_export.status`:

- **`pending` / `processing`** — row shows relative `started_at` and a spinner. No download affordances.
- **`ready`** — row shows relative `completed_at` and **N download links** (one per part-zip) labelled `Part 1`, `Part 2`, …, ordered by `chunk_id`. Clicking a part link fetches the presigned URL from the detail endpoint (AC 10) and opens it. A single-part export renders as a single "Download" link rather than "Part 1".
- **`failed`** — row shows relative `completed_at` and a compact error surface (icon + tooltip containing `error_message`). No retry affordance in this ticket — the user re-triggers via the Export CSV action.

(When the denormalized follow-up ships, a mode label — "Per feature type" / "Denormalized" — is added to each row. This ticket only has one mode, so the label is omitted.)

Refresh model: no polling. The card reflects whatever state `listExports` returned on the last fetch. Users pick up in-flight → ready transitions via the existing refresh button on the downloads sidebar (`DownloadSidebarDownloads.tsx`), which re-fires `listExports` as part of the paginated list refresh. Kicking off a new export via the Export CSV action also triggers a local refresh on success so the new `pending` row appears immediately without the user needing to hit the refresh button.

Empty state: when a download has no exports, the Exports section is omitted entirely — only the Export CSV action is rendered.

### 14. Frontend: hooks, types, and tests

- Add `useDownloadExportApi.ts` under `app/src/hooks/api/` with: `createExport(downloadId, payload?: { max_part_size_bytes?: number })`, `listExports(downloadId)`, `getExport(exportId)`. Mirrors the shape of the existing `useDownloadApi.ts` hooks.
- Add `useDownloadExportApi.interface.ts` under `app/src/interfaces/` with the `DownloadExport`, `DownloadExportArtifact`, and `CreateExportPayload` (currently just `{ max_part_size_bytes?: number }`) types — Zod-inferred on the backend, manually typed on the frontend to match. The denormalized follow-up extends `CreateExportPayload` with `mode` + `columns`.
- Tests:
  - `DownloadFeatureCard.test.tsx` — gated rendering of the Export action (visible only when download is `ready`); clicking the action fires `createExport` and triggers a local refresh so the new `pending` row renders; exports list renders per-status variants (pending spinner, ready part links, failed error tooltip).
  - `useDownloadExportApi.test.ts` — request/response shapes for all three methods.

## Example Export Output

The user downloads **one part-zip per `download_export_artifact`**, all extracted into the same root folder `biohub-export-{exportId}/`. The merged tree mirrors the logical output.

Two part-zips (observation has two chunks; dataset and sample_site fit in one each):

```
biohub-{exportId}-part-1.zip
  biohub-export-{exportId}/
  ├── dataset/
  │   └── chunk1.csv
  ├── observation/
  │   └── chunk1.csv
  ├── sample_site/
  │   └── chunk1.csv
  └── files1/
      ├── 10042_photo-01.jpg
      └── 10042_fieldnotes.pdf

biohub-{exportId}-part-2.zip
  biohub-export-{exportId}/
  ├── observation/
  │   └── chunk2.csv
  └── files2/
      └── 10089_track.gpx
```

Extracted side-by-side into the same parent directory, they merge into:

```
biohub-export-{exportId}/
├── dataset/
│   └── chunk1.csv
├── observation/
│   ├── chunk1.csv
│   └── chunk2.csv
├── sample_site/
│   └── chunk1.csv
├── files1/
│   ├── 10042_photo-01.jpg
│   └── 10042_fieldnotes.pdf
└── files2/
    └── 10089_track.gpx
```

## Testing Notes

Backend

- Run all tests

Export CSV flow (manual)

1. In the app, search for features, add them to the cart, and check out to create a download. Wait for the download card to reach **Ready**.
2. From the download card, click **Export CSV**.
3. Wait for the queue to process the export (`download_export` status: pending → processing → ready).
4. Click each **Part N** link to download the zips.
5. Extract all part-zips into the same parent folder and verify:
   - One folder per feature type, with `chunk{N}.csv` files.
   - Header row on `chunk1.csv` only; later chunks are header-less and byte-concatenate into a valid CSV.
   - `files{N}/` folders contain the referenced binary attachments, prefixed with `{submission_feature_id}_`.
   - `filePath` column in each CSV resolves to the correct `files{N}/…` path.


api/src/paths/download/{downloadId}/export/index.ts
```diff
@@ -0,0 +1,169 @@
...
+import { publishProcessDownloadExportJob } from '../../../../queue/publisher';
+import { authorizeRequestHandler } from '../../../../request-handlers/security/authorization';
+import { DownloadExportService } from '../../../../services/download/download-export-service';
+import { DownloadService } from '../../../../services/download/download-service';
+import { getLogger } from '../../../../utils/logger';
+
+const defaultLog = getLogger('paths/download/{downloadId}/export');
+
+export const POST: Operation = [
+  authorizeRequestHandler(() => ({ and: [{ discriminator: 'SystemUser' }] })),
```
mauberti-bc, line 19: To allow exports of anonymous downloads (ie. to let unauthenticated users export non-secured data), I think we probably want to remove the `authorizeRequestHandler` and use `OptionalBearer: []` for the security instead of `Bearer`. 

The downstream service already has `const download = await this.downloadService.getAuthorizedDownload(downloadId, systemUserId);` so it is properly gated.

api/src/queue/publisher.ts
```diff
@@ -374,6 +390,77 @@ export const publishProcessDownloadJob = async (
...
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+
+    defaultLog.error({
+      label: 'publishProcessDownloadExportJob',
+      message: 'Failed to publish job',
+      downloadExportId: data.downloadExportId,
+      error
+    });
+
+    return { status: 'error', message: errorMessage };
```
mauberti-bc: Should we throw here, so the endpoint doesn't return 200 if the job failed to enqueue? Currently, if we fail to enqueue, I believe the endpoint returns 200 but there is no recovery mechanism, so the job will never be processed.

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,712 @@
...
+    const download = await this.downloadRepository.getDownloadById(exportRecord.download_id);
+
+    const schemaLookup = await this.buildSchemaLookup();
+    const featureTypes = await this.listExportFeatureTypes(download.download_id);
+
+    const maxPartSizeBytes = BigInt(exportRecord.max_part_size_bytes);
+    const archiverByPart = new Map<number, PartArchiverBundle>();
+    let currentPart = 1;
+
+    archiverByPart.set(currentPart, this.createPartArchiverBundle(exportId, download.download_id, currentPart));
```
mauberti-bc, line 648: Consider validating that the export will produce at least one row, so we're not writing empty export files?

mauberti-bc (changes requested):
Testing locally, I keep getting empty zip files (22 KB, and computer throws an error when trying to unzip because archive is empty). I am authenticated and tried both the cart and bulk download paths.

api/src/queue/publisher.ts
mauberti-bc: Fixed by https://github.com/bcgov/biohubbc-platform/pull/424

api/src/paths/download/{downloadId}/export/index.ts
dwhogan: Anonymous users was moved to a follow-up ticket.

api/src/services/download/download-export-pipeline-service.ts
dwhogan: fixed, thanks

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,752 @@
...
+    let currentEntry: PassThrough | null = null;
+
+    try {
+      while (true) {
+        const next = (await cursor.next()) as Record<string, unknown> | null;
+        if (next === null) {
+          break;
+        }
+
+        const submissionFeatureId = typeof next.submission_feature_id === 'number' ? next.submission_feature_id : 0;
```
mauberti-bc: I don't think the parquet always has `submission_feature_id` beacuse `buildParquetSchema` excludes it? We should always keep the `submission_feature_id` in the parquet and exported csvs.

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,752 @@
...
+
+        fileRefs.push(...this.collectArtifactKeyRefs(next, artifactKeyProperties, submissionFeatureId, currentPart));
+
+        const line = headers.map((h) => escapeCsvField(flattened[h] ?? '')).join(',') + '\n';
+        const lineBytes = BigInt(Buffer.byteLength(line, 'utf8'));
+
+        // Lazy-open the chunk entry on the first row of this call. Header
+        // rule: written only on chunk 1 of a feature type. Later chunks
+        // (within the same feature type, in later part-zips) start directly
+        // with data rows so `cat chunk1.csv chunk2.csv ...` reconstructs a
```
mauberti-bc: We should write headers for each chunk. Most users won't be familiar with `cat` and instead will just open the csvs independently, or a script will iterate through and read them.

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,752 @@
...
+
+    archiverByPart.set(currentPart, this.createPartArchiverBundle(exportId, download.download_id, currentPart));
+
+    // Binary file references are collected as rows stream by, each tagged
+    // with the part index the referring row landed in. Each part's binaries
+    // must stream into that part's archiver BEFORE the part is finalized —
+    // whether that finalize happens on roll-over (mid-stream, for memory
+    // bounds) or at the end. Otherwise a finalized part's CSV would reference
+    // `files{N}/...` entries that never land in the zip.
+    const allFileRefs: FileFeatureRef[] = [];
```
mauberti-bc: Reset `allFileRefs` per part? Currently, `allFileRefs` grows and each parts filters over the entire array.

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,752 @@
...
+          if (chunkIndex === 1) {
+            await writeLine(currentEntry, headerLine);
+            currentBundle().byteCount += BigInt(Buffer.byteLength(headerLine, 'utf8'));
+          }
+        }
+
+        await writeLine(currentEntry, line);
+        currentBundle().byteCount += lineBytes;
+
+        if (shouldRollPart(currentBundle().byteCount, maxPartSizeBytes)) {
```
mauberti-bc, line 369: There's an edge case where the threshold for a new part is reached, but that row was the last, so the next new part is empty. I'd suggest looking ahead to ensure the next part has at least one row.

api/src/services/download/download-export-pipeline-service.ts
```diff
@@ -0,0 +1,752 @@
...
+
+    try {
+      while (true) {
+        const next = (await cursor.next()) as Record<string, unknown> | null;
+        if (next === null) {
+          break;
+        }
+
+        const submissionFeatureId = typeof next.submission_feature_id === 'number' ? next.submission_feature_id : 0;
+        const flattened = flattenFeatureBySchema(next, properties, submissionFeatureId, `files${currentPart}`);
```
mauberti-bc, line 337: Parquet stores spatial columns as WKB for GeoParquet, but the CSV export path currently treats spatial values like GeoJSON and tries to split them into `decimalLatitude` / `decimalLongitude`, so geometry can come out empty or malformed.

For the Parquet-to-CSV export, we should decode the WKB and emit a single geometry column using the original property name, serialized as WKT. This preserves points, lines, and polygons instead of forcing everything into lat/long columns.

examples:
```
POINT (-123.1207 49.2827)
LINESTRING (-123.1 49.2, -123.2 49.25, -123.3 49.3)
POLYGON ((-123.1 49.2, -123.3 49.2, -123.3 49.4, -123.1 49.4, -123.1 49.2))
MULTIPOINT ((-123.1 49.2), (-123.2 49.3))
MULTILINESTRING ((-123.1 49.2, -123.2 49.3), (-123.4 49.5, -123.5 49.6))
MULTIPOLYGON (((-123.1 49.2, -123.3 49.2, -123.3 49.4, -123.1 49.4, -123.1 49.2)))
```

mauberti-bc (changes requested):
Nice, very comprehensive!

api/src/services/download/download-export-pipeline-service.ts
dwhogan: added header to each chuck

api/src/services/download/download-export-pipeline-service.ts
dwhogan: added submission_feature_id

api/src/services/download/download-export-pipeline-service.ts
dwhogan: fixed per part, thanks

api/src/services/download/download-export-pipeline-service.ts
dwhogan: fixed with a look ahead, nice catch, thanks

api/src/services/download/download-export-pipeline-service.ts
dwhogan: moved from lat/lon to wkt

mauberti-bc (dismissed):
Nice!

mauberti-bc (changes requested):
👍

mauberti-bc (approved):
👍

PR #423, opened 2026-04-24
SIMSBIOHUB-975: Bug Fix: Preserve Original Error when Thrown

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-975

## Description of Changes
Regression where “cart not found” responses were returned as `500` instead of `404`.
Repository/business logic still threw the correct not-found error.
Root cause was custom error type identity loss after the TypeScript/toolchain upgrade.
Error mapping depended on type checks; failed checks caused fallback to generic `500 Unexpected Error`.
Fix was applied in the shared base error class to preserve custom error identity and proper stack behavior.
Result: not-found errors now map correctly to `404` again.
Verified with targeted error-mapping and cart repository tests.

### DIFF
--- a/api/src/errors/base-error.test.ts
+++ b/api/src/errors/base-error.test.ts
@@ -1,0 +1,42 @@
+import { expect } from 'chai';
+import { describe } from 'mocha';
+import { ApiError, ApiNotFoundError } from './api-error';
+import { BaseError } from './base-error';
+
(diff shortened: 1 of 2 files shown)


dwhogan:
Replacing with PR Brand-symbol pattern for ESM-resilient instanceof

PR #424, opened 2026-04-24
SIMSBIOHUB-973:  Queue publishers should throw on failure

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-973](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-973)

## Description of Changes

**Acceptance criteria**

1. Publishers in `api/src/queue/publisher.ts` throw (do not swallow) when `boss.send` rejects; the existing structured error log is emitted before the rethrow.
2. PublishJobResult` is narrowed to `published | duplicate`; the `error` variant is removed from the type and from every call site.
4. Caller `services/upload/upload-ingestion-service.ts:232` is `await`ed.
5. Publisher unit tests assert that each publisher throws when `boss.send` rejects (prior `{ status: 'error' }` expectations flipped). Existing `published` / `duplicate` assertions still pass.
6. Each of the seven production call sites has an integration or route test asserting the DB row is not committed when its publisher throws, following the pattern at `api/src/paths/download/index.test.ts:260`.

## Testing Notes

- API internal changes
- Run all tests.

### DIFF
--- a/api/src/__integration__/db/cart-checkout.integration.ts
+++ b/api/src/__integration__/db/cart-checkout.integration.ts
@@ -3,11 +3,14 @@
 // works correctly against the real database.
 //
 // Uses a transaction that is ROLLED BACK after each test, so no data is persisted.
+// pg-boss is not running in the make test-db environment, so CartService.dependencies.publishProcessDownloadJob
+// is stubbed in beforeEach to resolve without attempting a real enqueue.
 //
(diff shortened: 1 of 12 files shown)


mauberti-bc (dismissed):
Nice!

mauberti-bc (approved):
👍

PR #425, opened 2026-04-28
SIMSBIOHUB-975: Brand-symbol pattern for ESM-resilient instanceof

dwhogan:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-975

## Description of Changes

 After the ESM migration, `instanceof CustomError` checks became unreliable. The express `errorMiddleware` dispatcher in `app.ts:128` calls `ensureHTTPError(error)`, which uses `instanceof` to map error classes to HTTP statuses. When the dispatcher's reference to a class object differs from the one used at the throw site (a normal consequence of ESM URL-keyed module caching), every `instanceof` branch returns false and errors fall through to `isAjvError`, which rewraps them with the wrong name and message.

Originally reported as "Cart not found returns 500 instead of 404." The status code is actually correct in most cases; the visible defect is the **malformed body** — `name` and `message` get clobbered to `"Bad Request"` and `"Request Validation Error"`.

## Acceptance criteria

1. `instanceof CustomErrorClass` returns the correct result regardless of how many copies of the class exist in the module graph.
2. `errorMiddleware` returns response bodies whose `name`, `status`, and `message` match the thrown error class — not AJV-error placeholders.
3. The fix covers all classes used in `instanceof` dispatch: `HTTPError` and subclasses, `ApiError` and subclasses, `IngestionValidationError`.
4. No unit-test regressions; new tests cover the cross-module identity case explicitly.
5. Live `make web` API: hitting an endpoint that throws `HTTP404` returns a correctly-formed 404 body, before and after editing files (no malformed-body window during tsx-watch reloads).
6. `AsyncRequestStorage` survives ESM module duplication. Request-id log correlation remains intact across tsx-watch reloads — no silent fallback to `'SYSTEM'` when middleware writes through one module copy and downstream code reads from another.

## Testing notes

```bash
make test
```

Covers AC 1, 3, 4, and 6 (unit tests, including the cross-module `instanceof` cases).

For AC 2 and 5, run `make web` and hit an endpoint that throws an `HTTPError` subclass — body should match the thrown error, not `"Request Validation Error"`:

```bash
curl -s http://localhost:6100/api/download/00000000-0000-0000-0000-000000000000
# {"name":"Not Found","status":404,"message":"Download not found","errors":[]}
```

### DIFF
--- a/api/src/errors/api-error.ts
+++ b/api/src/errors/api-error.ts
@@ -10,6 +10,8 @@ export enum ApiErrorType {
 }
 
 export class ApiError extends BaseError {
+  static readonly brand: symbol = Symbol.for('@biohub/error/ApiError');
+
   constructor(name: ApiErrorType, message: string, errors?: (string | object)[], stack?: string) {
     super(name, message, errors, stack);
   }
@@ -23,6 +25,8 @@ export class ApiError extends BaseError {
  * @extends {ApiError}
  */
 export class ApiGeneralError extends ApiError {
+  static readonly brand: symbol = Symbol.for('@biohub/error/ApiGeneralError');
+
   constructor(message: string, errors?: (string | object)[]) {
     super(ApiErrorType.GENERAL, message, errors);
   }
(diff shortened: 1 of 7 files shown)


mauberti-bc (dismissed):
👍

mauberti-bc (approved):
👍

PR #426, opened 2026-04-29
SIMSBIOHUB-968: Search with Expressions (Backend)

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-968

## Description of Changes

- Added property-driven expression predicates using `feature_property_id` with optional `feature_type_property_id` narrowing.
- Added normalized expression tree validation, hashing, persistence, and read-back support.
- Added typed predicate payload handling for string, number, boolean, datetime, taxon, spatial, and code properties.
- Switched internal timestamp predicates to scalar string values; date/time splitting now happens only at DB/query boundaries.
- Added expression-tree feature search support with recursive graph projection.
- Added property search metadata for supported operators and value input types.
- Cleaned up expression/predicate types to reduce redundancy and keep shared types in model/interface files.

Expression search works like this:

1. **Validate and enrich the expression**
   - The API receives an `ExpressionTree`.
   - Each predicate is resolved to property metadata:
     - shared `feature_property_id`
     - optional `feature_type_property_id`
     - property type id/name
     - typed internal predicate value

2. **Find evidence features**
   - Each predicate queries the correct typed property table.
   - It always filters by shared property:
     ```sql
     ftp.feature_property_id = clause.feature_property_id
     ```
   - If `feature_type_property_id` is **non-null**, it also filters to that exact property:
     ```sql
     p.feature_type_property_id = clause.feature_type_property_id
     ```
   - If `feature_type_property_id` is **null**, it matches that shared property across all active feature-type properties.

3. **Project evidence to target features**
   - Evidence features may not be the route’s target feature type.
   - The query walks feature relationships to find connected features of the requested `anchorFeatureType`.
   - Traversal uses:
     - explicit feature-to-feature links
     - parent/child links
     - dataset-to-same-submission links
   - It is recursive, depth-limited, and cycle-safe.

4. **Combine expression clauses**
   - Each predicate produces a target feature id set.
   - `AND` combines child sets with `INTERSECT`.
   - `OR` combines child sets with `UNION`.

Example:

```ts
AND [
  { feature_property_id: 10, feature_type_property_id: null, value: 'red' },
  { feature_property_id: 20, feature_type_property_id: 55, value: 5 }
]
```

This means:
- Predicate 1 matches shared property `10` across feature types.
- Predicate 2 matches only exact `feature_type_property_id = 55`.
- Each match set is projected to the target feature type.
- Final results are targets present in both sets.

Security is applied to evidence before traversal and to final target results before response.


database/src/migrations/20260501090000_property_driven_predicate_anchor.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Predicates are now keyed by the required shared feature_property_id, while feature_type_property_id becomes the only
+ * nullable predicate anchor. This lets search predicates target a property across feature types, while persisted
+ * expression predicates can still be narrowed to one feature type property when needed.
+ *
+ * @param {Knex} knex - Knex database client.
+ * @return {Promise<void>}
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    --------------------------------------------------------------------------------
```
dwhogan, line 17: nit: should we drop the code convention to use `SET SEARCH_PATH' in migration files as you have done here.  It is not request as knex has `biohob`?

api/src/services/expression-predicate-semantic-validator.ts
```diff
@@ -0,0 +1,356 @@
...
+        { featurePropertyTypeName: metadata.feature_property_type_name }
+      ]);
+    }
+
+    const propertyType: PredicatePropertyTypeName = metadata.feature_property_type_name;
+    const allowedOperators: readonly PredicateOperator[] = OperatorsByPropertyType[propertyType];
+
+    // Operator compatibility is semantic because it depends on the resolved property type.
+    if (!allowedOperators.includes(predicate.operator)) {
+      throw new ApiGeneralError('Predicate operator is not valid for property type', [
```
dwhogan: nit: reminder that ApiGeneralError maps to http 500, http 400 may align better

api/src/repositories/property-search-repository.ts
```diff
@@ -115,96 +94,107 @@ export class PropertySearchRepository extends BaseRepository {
...
-      .distinct('ss.feature_property_id', 'ftp.name as property_name', knex.raw('1.0 as relevancy_score'))
-      .from('search_string as ss')
-      .innerJoin('feature_property as ftp', 'ss.feature_property_id', 'ftp.feature_property_id');
+      .select(
+        'fp.feature_property_id',
+        'fp.name as property_name',
+        'fp.display_name as property_display_name',
+        'fpt.name as feature_property_type',
+        knex.raw(this.buildOperatorCaseSql()),
+        'fpt.name as value_input',
```
dwhogan: look like a duplicate `fpt.name as value_input`

api/src/utils/timestamp.ts
```diff
@@ -0,0 +1,45 @@
...
+  time_value: string | null;
+};
+
+/**
+ * Parses a timestamp-like scalar string into separate date/time parts.
+ *
+ * Supports date-only, time-only, and datetime strings. Returns null when the input cannot be split safely.
+ */
+export const parseTimestamp = (value: string): ParsedTimestamp | null => {
+  const datePattern = /^(\d{4}-\d{2}-\d{2})$/;
```
dwhogan: perhaps use Date.parse instead of regex to better validate the value

dwhogan (changes requested):
first-class work!!

database/src/migrations/20260501090000_property_driven_predicate_anchor.ts
mauberti-bc: Probably worth keeping to ensure we're in the right schema, in case something upstream changes. I'll put it back, thanks for noticing!

api/src/services/expression-predicate-semantic-validator.ts
mauberti-bc: Thanks, fixed. Added new `ApiValidationError` error class that returns 400 (distinct from the `IngestionValidationError` that we use for some pgboss jobs)

dwhogan (dismissed):
👍

dwhogan (approved):
👍

PR #431, opened 2026-05-04
SIMSBIOHUB-992: Fix all references to `submission_feature_property_timestamp.value`

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-992](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-992)

## Description of Changes

The download flow is broken because there are some more `submission_feature_property_timestamp.value` references, but that value got split into `date_value` and `time_value`. 

**Acceptance Criteria**
1. Change references to explicit date/time

Javascript Date/dayjs objects must be split into separate date and time values when queried against the submission_feature_property_timestamp table
All references to the singular `values` should be updated to use date_value/time_value
Queries for "time_value" with a null date_value should ignore date
eg. Find all records between 1pm and 5pm, irrespective of the date
Queries for "date_value" with a null time_value should ignore time

## Testing Notes

Automated coverage:

```bash
make test                     # unit tests
make test-db                  # DB integration
make test-sys                 # system integration
```

Manual smoke (download flow):

1. `make web && make minio && make queue` — start the stack.
2. With default seed data, search **telemetry** (`timestamp`), add to cart, check out, wait for **Ready**.
3. **Export CSV** from the download card; wait for `download_export` to reach `ready`.
4. Open the resulting CSV — datetime columns should render as separate date / time values (no empty cells, no `[object Object]`, no `undefined`).

### DIFF
--- a/api/src/utils/parquet-utils.ts
+++ b/api/src/utils/parquet-utils.ts
@@ -9,7 +9,11 @@
  */
 import { ParquetSchema, type FieldDefinition, type SchemaDefinition } from '@dsnp/parquetjs';
 
-import { DATETIME_DATE_SUFFIX, DATETIME_TIME_SUFFIX } from '../models/datetime-column';
+import {
+  assertNoDatetimeColumnCollisions,
+  DATETIME_DATE_SUFFIX,
+  DATETIME_TIME_SUFFIX
+} from '../models/datetime-column';
 import { ParquetFeatureData } from '../models/download';
 import type { CsvPropertyDefinition } from './csv-utils';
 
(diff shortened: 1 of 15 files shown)


api/src/utils/parquet-utils.ts
```diff
@@ -34,8 +139,10 @@ export function propertyTypeToParquetType(typeName: string): FieldDefinition['ty
       return 'DOUBLE';
     case 'boolean':
       return 'BOOLEAN';
-    case 'datetime':
-      return 'TIMESTAMP_MILLIS';
+    // `datetime` is intentionally absent: it is handled upstream by
```
mauberti-bc: Consider throwing an error for the 'datetime' case if it is wrong.

api/src/utils/parquet-utils.ts
```diff
@@ -147,14 +271,6 @@ export function featureToRow(
       case 'object':
         row[prop.feature_property_name] = JSON.stringify(value);
         break;
-      case 'datetime':
```
mauberti-bc, line 150: Consider error for  'datetime' case

mauberti-bc (changes requested):
Nice work!

mauberti-bc (changes requested):
Nice work!

mauberti-bc (approved):
👍

PR #432, opened 2026-05-05
SIMSBIOHUB-996: Prevent features from referencing themselves as their parent

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-996

## Description of Changes

- Prevents a feature from being its own parent.
- Rejects active parent chains like `A -> B -> C -> A`.
- Uses a deferred trigger so multi-row hierarchy updates can commit if the final state is valid.

## Testing Notes

- Import a sample tarball and ensure the ingestion job succeeded


mauberti-bc:
Closing because the per-row trigger does not scale with the single UPDATE used when setting the parent.

PR #434, opened 2026-05-05
SIMSBIOHUB-988: Create expression-based downloads

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-988](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-988)

## Description of Changes

This ticket adds expression-based download creation from search results.

When a user creates a download from a search, the backend should create a policy that represents the requested feature set. The download should store `download.policy_id`, and download generation should evaluate that policy to find the features written to the Parquet output.

The policy defines what features belong in the download. The download defines how that feature set is exported.

### 3. Create Policy For Download Feature Set

- Create a policy that represents the download feature set
- Create one policy statement for each selected feature type
- Scope each policy statement to the selected feature type
- Use a feature-type URN scope for each statement
- Link the current search expression to each generated policy statement

Example statement scopes:

```
urn:*:telemetry:*
urn:*:species_observation:*
urn:*:dataset:*
```

### 4. Link Download To Policy

- Create the download record after creating the policy
- Store the generated policy ID on `download.policy_id`
- Treat `download.policy_id` as the source of truth for the feature set to export
- Do not duplicate the expression tree directly on the download record
- Preserve the policy and expression links for auditability and reruns

### 5. Keep Download Policies Separate From Access Grants

- Do not create `team_policy` rows for download policies
- Do not create `security_scope` rows for download policies
- Do not create security anchors for download policies
- Do not grant access through download creation
- Evaluate download policies directly when generating the export

### 6. Generate Parquet From Download Policy

- Load `download.policy_id` during download generation
- Evaluate the linked policy to determine which features belong in the export
- Apply each policy statement's feature-type scope before evaluating its expression
- Apply the linked expression as the fine-grained feature condition
- Write only policy-matching features to the Parquet output
- Respect the selected export feature types when deciding which Parquet files to produce

### 7. Apply Authorization Rules During Export

- Ensure the user can only create downloads for data they are allowed to export
- Ensure secured features are not exported unless the user has access through active team access
- Do not rely on download policy creation to grant access
- Apply authorization separately from expression matching
- Prevent downloads from leaking secured feature data

### 8. Support Empty Or Broad Download Expressions

- If broad downloads are allowed, support policy statements without expression links
- If broad downloads are not allowed, require a valid search expression before enabling download creation
- Do not create an empty expression tree with zero clauses
- Treat omitted expression and `null` expression consistently
- Make broad-download behavior explicit in the UI and backend validation

### 10. Preserve Service Boundaries

- Keep the service layer responsible for creating the policy, policy statements, expression links, and download record
- Keep policy evaluation logic separate from download creation logic
- Keep Parquet generation logic separate from the create-download request
- Reuse existing expression persistence and policy-statement expression-linking services

## Testing Notes


api/src/paths/download/index.ts
```diff
@@ -217,70 +217,71 @@ POST.apiDoc = {
...
+ *
+ * @return {RequestHandler}
  */
 export function createDownload(): RequestHandler {
   return async (req, res) => {
-    const isAuthenticated = !!req.keycloak_token;
-    const connection = isAuthenticated
-      ? downloadPathDependencies.getDBConnection(req.keycloak_token)
-      : downloadPathDependencies.getAPIUserDBConnection();
+    const parseResult = CreateDownloadRequestBody.safeParse(req.body);
```
mauberti-bc, line 228: nit: add comment explaining why we need to Zod parse, since we usually rely on openapi schemas to validate parts of the request. Just to make it clear so it's not unintentionally deleted. I think Zod is used here because it's more useful for validating expressions than openapi.

api/src/paths/download/index.ts
```diff
@@ -217,70 +217,71 @@ POST.apiDoc = {
...
-      // Count-only validation — no ID materialization.
-      // Security filtering happens in SQL via buildSecurityFilter:
-      // - systemUserId === null → anonymous, exclude secured features
-      // - systemUserId === number → authenticated, include unsecured + scope-granted
-      const systemUserId = isAuthenticated ? connection.systemUserId() : null;
-      const matchCount = await searchFeatureService.getSearchFeaturesCount(filters, systemUserId);
-
-      if (matchCount === 0) {
-        throw new HTTP400('No features match the filter criteria');
+      let expressionId: string | null = null;
```
mauberti-bc: Suggestion: move the business orchestration of creating policy -> download -> job, etc. into the service layer rather than the endpoint file.

api/src/repositories/expression-evaluation.ts
```diff
@@ -0,0 +1,713 @@
...
+import { InternalTypedPredicate } from '../models/expression-predicate';
+import {
+  NormalizedExpressionTreeClause,
+  NormalizedExpressionTreeExpression,
+  NormalizedExpressionTreePredicate
+} from '../models/expression-tree-internal';
+import { parseTimestamp } from '../utils/timestamp';
+import { buildSecurityFilter } from './sql-fragments';
+
+/**
```
mauberti-bc, line 14: Nice extraction for reuse

api/src/repositories/expression-evaluation.ts
```diff
@@ -0,0 +1,713 @@
...
+    )
+    SELECT 1
+    FROM ancestors
+    WHERE taxon_id = ?`,
+    [targetTaxonId]
+  );
+}
+
+/**
+ * Applies a geometry expression operator.
```
mauberti-bc, line 692: nit: use complete jsdocs with `@param` and `@returns`

api/src/services/download/download-pipeline-service.ts
```diff
@@ -110,37 +116,36 @@ export class DownloadPipelineService extends DBService {
...
+   * produces a stable, alphabetic sequence of Parquet files.
+   *
+   * @param {string} _downloadId - The download ID (kept for future diagnostics; unused today).
+   * @param {DownloadSource} source - The download source (policy_id + create_user).
+   * @return {Promise<{ schemaLookup: Map<string, CsvPropertyDefinition[]>; featureTypes: string[]; statements: ActivePolicyStatementWithExpression[] }>}
    * @memberof DownloadPipelineService
    */
   async resolveParquetSchema(
-    downloadId: string,
+    _downloadId: string,
```
mauberti-bc: nit: remove `_downloadId` prop if unused, unless there is expected downstream usage?

api/src/services/download/download-policy-service.ts
```diff
@@ -0,0 +1,91 @@
...
+   * @memberof DownloadPolicyService
+   */
+  async createDownloadPolicy(payload: CreateDownloadPolicyPayload): Promise<{ policy_id: string }> {
+    // Reject unknown feature type names up front so a typo or stale FE id doesn't silently
+    // produce a download with empty Parquet files. The check is per-create rather than at
+    // app start because feature_type rows can be added/retired between deploys; one extra
+    // SELECT per request is cheap (~10 rows in seed; bounded by domain size).
+    const activeFeatureTypes = await this.featureIngestionRepository.getActiveFeatureTypeMap();
+    const knownNames = new Set(activeFeatureTypes.map((row) => row.name));
+    const unknown = payload.featureTypes.filter((name) => !knownNames.has(name));
```
mauberti-bc, line 61: Handle case sensitivity by lowercasing before `.has`?

mauberti-bc (changes requested):
Very nice!

api/src/services/download/download-policy-service.ts
dwhogan: this is programmatic data, so to keep symmetry, perhaps keep as is (even though the convention and some enforcement (urn builder) is lowercase.

api/src/services/download/download-pipeline-service.ts
dwhogan: fixed

api/src/repositories/expression-evaluation.ts
dwhogan: jsdocs important

api/src/repositories/expression-evaluation.ts
dwhogan: thanks

api/src/paths/download/index.ts
dwhogan: good suggestion, i added to the service

api/src/paths/download/index.ts
dwhogan: comment added

mauberti-bc (approved):
👍

PR #441, opened 2026-05-08
SIMSBIOHUB-1002: Create Downloads from Search Expressions (Frontend)

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1002](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1002)

## Description of Changes

Replace the search-results "Bulk Download" one-click button with a create-download dialog that collects download metadata, decodes the current search `ExpressionTree`, and submits the new download payload to the backend introduced by [SIMSBIOHUB-988](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-988).

The existing *Checkout* (cart) flow on the same page is *out of scope* — it stays cart_id-based and is not affected by this ticket.
 
**Acceptance Criteria**

1. Create Downloads From Search Expressions

Replace `handleDownloadAll` in `app/src/features/search/result/SearchResultPage.tsx` so the Bulk Download button opens the create-download dialog instead of immediately POSTing.
Use the current search `ExpressionTree` (decoded from URL state by [SIMSBIOHUB-979](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-979) plumbing) as the download criteria.
Strip UI-only fields (e.g. `ui_id`) before submit.
Do *not* submit the Base64URL-encoded URL value — submit the decoded `ExpressionTree` object.
Submit `expression: null` when the search has no expression (broad download); do not block submission on an empty expression.

2. Collect Download Metadata

Open a create-download dialog (or page) from the search-results Bulk Download entry point.
Required field: download `name`.
Optional field: download `description`.
Multi-select: `featureTypes[]` — which feature types to include in the Parquet output.
Disable submit while required fields are missing.
Disable submit when no export feature types are selected.
Prevent duplicate submissions while the request is in flight.

3. Show Download Creation Outcomes

Show a success snackbar (or equivalent) on success.
Show a clear error snackbar/dialog on failure, including the API error message.
Route the user to the download detail page or refresh the downloads sidebar per product flow.
Refresh download list state after successful creation where appropriate.

## Testing Notes

```bash
make test
```

Manual smoke (create-download flow):

1. On `/search/telemetry`, `Create Download` opens a dialog pre-filled with `Telemetry download` and `[telemetry]`; submit succeeds and the sidebar flips to Downloads with a new row.
2. Zero-result search → `Create Download` opens the "no features matching" OkDialog (no form).


app/src/features/search/result/SearchResultPage.tsx
```diff
@@ -218,37 +232,79 @@ export const SearchResultPage = () => {
...
+   *
+   * `expression` is forwarded as the literal page state. The expression-builder popover already
+   * strips builder-only `ui_id` fields at apply time, so the page state is wire-clean by
+   * construction; a second sanitizer here would duplicate the contract guarantee. The key must
+   * be present (the backend uses `.nullable()`, not `.optional()`); when no expression is
+   * applied, the page state is `null` and is sent as `null`.
+   */
+  const handleCreateDownload = useCallback(
+    (values: ICreateDownloadFormValues) =>
+      runSerialized(async () => {
```
mauberti-bc, line 275: Nice, I forgot about that hook!

app/src/features/search/result/SearchResultPage.tsx
```diff
@@ -218,37 +232,79 @@ export const SearchResultPage = () => {
...
-          open: true,
-          onClose: () => dialogContext.setOkDialog({ open: false })
-        });
-      }
-    } catch (error) {
-      dialogContext.setSnackbar({ snackbarMessage: (error as APIError).message, open: true });
-    } finally {
-      setIsDownloading(false);
+  const handleOpenCreateDownload = useCallback(() => {
+    if (pagination === undefined) {
```
mauberti-bc: Create Download should also be guarded by isLoading from `useSearchResults`. During a search refresh, the old pagination.total remains set until the new request finishes, so the dialog can open based on a stale result count.

```
if (isLoading || pagination === undefined) {
  return;
}
```

mauberti-bc (changes requested):

app/src/features/search/result/SearchResultPage.tsx
dwhogan: nice add

mauberti-bc (approved):
👍

PR #445, opened 2026-05-13
SIMSBIOHUB-981: Base64 encode search expression

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-981](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-981)

## Description of Changes

- Made expression-based search results bookmarkable by storing the applied search expression in the URL as `expr=<base64url>` on `/search/:featureType`.
- Added centralized expression URL utilities:
  - deterministic serialization (`stableStringify`, object-key sorting, array order preserved)
  - base64url encode/decode (URL-safe alphabet, no padding)
  - decode validation guard for `ExpressionTree`.
- Refactored search result expression state to be URL-driven:
  - decode `expr` on load/navigation
  - update `expr` only on Apply
  - remove `expr` when clearing filters
  - preserve unrelated query params
  - show non-blocking snackbar and fall back to no expression when `expr` is invalid.
- Kept backend contract unchanged:
  - frontend still calls `POST /api/search/feature/:feature_type`
  - decoded `ExpressionTree` is sent in request payload.
- Added `URL_PARAMS.EXPR` and `TypedURLSearchParams.getRaw()` for case-sensitive query values.
- Fixed hydration edge cases for shared/bookmarked URLs:
  - preserve predicate values even when property metadata is not immediately available
  - infer temporary predicate type from operator/value as fallback
  - improve property option hydration by paging through property search results (bounded) so used properties from URL expressions can be resolved without invalid request parameters.

## Testing Notes

- Added/expanded tests for expression URL behavior:
  - round-trip encode/decode
  - stable serialization determinism
  - base64url alphabet/no-padding
  - invalid decode/validation fallback paths.
- Updated expression-builder/search-result tests to reflect URL-driven behavior and hydration timing.
- Verified:
  - app typecheck passes
  - full app test suite passes
  - manual scenario: `Name contains Horse OR Name contains Snake` persists correctly when URL is copied/opened in a new tab, including populated property + value in the filter UI.


app/src/utils/expression-url.ts
```diff
@@ -0,0 +1,251 @@
...
+ * - Rejects `ui_id` on any node (defense-in-depth; the serializer already strips it).
+ *
+ * @param value - The value to check.
+ * @returns `true` when `value` is a valid `ExpressionTreeExpression`.
+ */
+export function isValidExpressionTree(value: unknown): value is ExpressionTreeExpression {
+  return isValidExpression(value);
+}
+
+function isValidExpression(value: unknown): value is ExpressionTreeExpression {
```
mauberti-bc, line 84: nit: add jsdocs to every util function, even if its usage is explained by a wrapper (eg. isValidExpressionTree in this context). Helps delineate functions in the file

app/src/utils/expression-url.ts
```diff
@@ -0,0 +1,251 @@
+import {
+  ExpressionLogicalOperator,
+  ExpressionPredicateOperator,
+  ExpressionTreeExpression
+} from 'interfaces/expression.interface';
+
+const LOGICAL_OPERATORS: ReadonlySet<string> = new Set<ExpressionLogicalOperator>(['AND', 'OR']);
+
+const PREDICATE_OPERATORS: ReadonlySet<string> = new Set<ExpressionPredicateOperator>([
```
mauberti-bc: `PREDICATE_OPERATORS` repeats the same operator literals already defined a few times (eg. by `ExpressionPredicateOperator`). 

Could we export a single `EXPRESSION_PREDICATE_OPERATORS` const array, derive the union type from it, and reuse that array here for the runtime `Set`? Or some similar way of not duplicating the operators.

That would keep the frontend type and URL validator from drifting if operators are added or renamed later.

mauberti-bc (changes requested):
Works very well!

mauberti-bc (approved):
👍

PR #446, opened 2026-05-13
SIMSBIOHUB-985: Restrict security scopes to access-orientied policies

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-985](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-985)

## Description of Changes

This ticket updates `SecurityScopeService` so security scopes are only created and maintained for access-oriented policies.

`security_scope` should not be treated as the generic scope mechanism for all policies. It is an access cache used to make search authorization fast for teams with active `team_policy` grants.

Policies used only for data requests, downloads, security reasons, or other non-access workflows should not create security scopes, anchors, or team scope mappings.

Acceptance Criteria

1. Restrict Security Scope Creation to Access Grants

Create `security_scope` records only when a policy is linked through `team_policy`  
Do not create `security_scope` records for generic policy creation  
Do not create `security_scope` records when policy statements are created outside an access-grant flow  
Do not create `security_scope` records for policies linked only to `data_request`, `download`, or `security_reason`  
Treat `security_scope` as an access cache, not as the generic policy scope model  

2. Update SecurityScopeService Responsibilities

Update `SecurityScopeService` so its public methods clearly operate on team access grants  
Ensure `createScopeForPolicyStatement` is only called from access-oriented flows  
Prefer naming or comments that make the access-only purpose clear  
Keep anchor computation and team scope derivation inside this service  
Remove assumptions that every policy statement should resolve to a security scope  

3. Create Scopes When Team Access Is Granted

When a `team_policy` is created, resolve the active policy statements for that policy  
For each relevant policy statement URN, compute or reuse the matching `security_scope`  
Create or reuse the policy-statement-to-scope mapping needed for access checks  
Publish anchor computation jobs for newly created or reused scopes where required by current idempotent behavior  
Insert or rebuild `team_security_scope` rows for the granted team access  

4. Keep Non-Access Policies Scope-Free

Creating a data request policy should not create security scopes  
Creating a download policy should not create security scopes  
Creating a security reason policy should not create security scopes  
Updating policy statements for non-access policies should not enqueue anchor computation jobs  
Non-access policy consumers should evaluate their policies directly instead of relying on `security_scope`  

5. Preserve Team Access Behavior

Existing search authorization should continue to use active team access mappings  
Authenticated users should see secured features only when their team has access through `team_policy`  
Anonymous users should continue to see only unsecured features  
Existing `team_security_scope` behavior should remain valid for active team policies  
Search authorization should not consider policies that are not linked through `team_policy`  

6. Update Team Policy Lifecycle Handling

When a `team_policy` is created, create/reuse required scopes and grant team access to them  
When a `team_policy` is removed or end-dated, rebuild or remove the affected team access mappings  
Ensure teams can still reach shared scopes through other active team policies  
Avoid deleting shared scopes or anchors while they are still used by another active team access grant  
Keep rebuild behavior idempotent and safe to retry  

7. Update Anchor Computation Triggers

Publish anchor computation jobs only for scopes used by team access  
Do not recompute anchors for policies that are not linked to `team_policy`  
Keep anchor computation idempotent for retry and partial completion  
Continue to support anchor recomputation when secured features change within a submission  
Ensure orphaned access scopes can still be cleaned up safely  

8. Update Policy Statement Cleanup

When policy statements are removed from an access policy, remove or update their access scope mappings  
Rebuild affected team security scopes after access policy statement changes  
Do not run security scope cleanup for policy statements that were never part of a team access grant  
Preserve shared scope rows when other active access policies still reference them  
Clean up orphaned access scope data only when no active access mappings remain  

9. Clarify Service Documentation

Update service comments to describe `security_scope` as a team access cache  
Remove documentation that describes security scopes as generic policy scopes  
Clarify that policy URNs are generic statement scopes, while security scopes are access materializations  
Clarify that data request, download, and security reason policies do not use `security_scope`  
Keep comments aligned with the actual call graph and lifecycle behavior  

10. Preserve Existing Database Model Where Possible

Reuse the existing `security_scope`, policy-statement mapping, anchor, and team scope tables where appropriate  
Avoid adding a new generic policy scope table in this ticket  
Avoid broad schema churn unless required to enforce access-only scope creation  
Keep existing unique indexes and idempotent insert behavior  
Preserve existing search authorization SQL shape where possible  

11. Prevent Incorrect Access Leakage

Ensure download policies cannot accidentally grant search access  
Ensure data request policies cannot grant access until approved into `team_policy`  
Ensure security reason policies cannot grant search access  
Ensure only active team access grants participate in authorization  
Ensure stale scope mappings do not cause users to see secured features they should not see  

## Testing Notes

API change only:

```bash
make test
make test-db
```

### DIFF
--- a/api/src/services/access-policy/policy-service.ts
+++ b/api/src/services/access-policy/policy-service.ts
@@ -168,8 +168,16 @@ export class PolicyService extends DBService {
     }
 
     if (to === 'approved') {
+      // Materialize the shared statement-scope rows once for this policy, then
+      // grant access per team. The statement-scope work depends only on the
+      // policy, so splitting these avoids re-running the policy-wide INSERTs
+      // and re-publishing anchor jobs for every team in the fan-out.
+      const materialized = await this.securityScopeService.materializePolicyStatementScopes(policyId);
+      if (!materialized) {
+        return;
+      }
       for (const teamPolicy of teamPolicies) {
-        await this.securityScopeService.materializeStatementScopesAndTeamAccess(teamPolicy.team_id, policyId);
+        await this.securityScopeService.grantTeamAccessForPolicy(teamPolicy.team_id, policyId);
       }
     } else if (from === 'approved') {
       for (const teamPolicy of teamPolicies) {
@@ -323,8 +331,9 @@ export class PolicyService extends DBService {
    * Statement creation no longer materializes security scopes. The access-cache
    * (`security_scope`, `policy_statement_scope`, `team_security_scope`) is only
    * populated when a `team_policy` link exists and the policy is approved — that
-   * fan-out lives in `SecurityScopeService.materializeStatementScopesAndTeamAccess`
-   * and fires from `TeamPolicyService.createTeamPolicy` and from policy approval.
+   * fan-out lives in `SecurityScopeService.materializePolicyStatementScopes` +
+   * `grantTeamAccessForPolicy` and fires from `TeamPolicyService.createTeamPolicy`
+   * and from policy approval.
    *
    * @param {CreatePolicy} policyData - Policy payload.
    * @param {CreatePolicyStatementPayload[]} statements - Statement payloads.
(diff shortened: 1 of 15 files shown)


api/src/services/access-policy/policy-service.ts
```diff
@@ -107,43 +119,84 @@ export class PolicyService extends DBService {
...
+    }
 
-    if (blockedStatuses.has(current)) {
-      throw new HTTP400(`Cannot approve request while policy is '${current}'`);
+    const teamPolicies = await this.teamPolicyRepository.getTeamPolicies({ policyIds: [policyId] });
+    if (teamPolicies.length === 0) {
+      return;
+    }
+
+    if (to === 'approved') {
```
mauberti-bc, line 170: Suggestion: is this doing unnecessary work by recalculating `security_scope_anchor` for each `team_policy` record? Since anchors are keyed by `security_scope_id` and are not team-specific, I think we can materialize/recompute anchors once for the updated policy, then rebuild `team_security_scope` per affected team.

mauberti-bc (changes requested):
Very sophisticated!

api/src/services/access-policy/policy-service.ts
dwhogan: Good catch, I refactored to materialize/recompute anchors once per policy

mauberti-bc (approved):

PR #447, opened 2026-05-15
SIMSBIOHUB-986: Create data request for search query

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-986](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-986)

## Description of Changes

This ticket adds the request-access flow from search results.

When a search response indicates that matching secured results exist but are not visible to the user, the UI should show a banner with a “Request Access” action. Clicking the action opens a request flow where the user selects which feature types they want access to.

The backend should create a data request backed by a policy. For each selected feature type, the backend creates a policy statement scoped to that feature type and links the current search expression to that statement.

Acceptance Criteria

1. Show Request Access Prompt From Search Results

Show a request-access banner when the search response indicates there are inaccessible matching results  
Do not show the banner when all matching results are visible  
Do not show the banner when there are no inaccessible matches  
Keep secured feature details hidden from users without access  
Make the banner action clear, such as “Request Access”  

2. Open Request Access Flow

Open a request-access dialog or page when the user clicks “Request Access”  
Show enough context for the user to understand what search they are requesting access for  
Do not expose raw expression JSON to the user  
Allow the user to cancel without creating a request  
Preserve the current search expression while the request flow is open  

3. Select Feature Types to Request

Show selectable feature types that can be requested from the current search context  
Allow the user to select one or more feature types  
Preselect the current search result feature type where appropriate  
Prevent submission when no feature types are selected  
Make it clear that selected feature types define what access is being requested  

4. Submit Request Using Current Expression

Submit the current search expression with the request-access payload  
Submit the selected feature types with the request-access payload  
Do not submit UI-only expression builder fields such as `ui_id`  
Do not submit the Base64URL-encoded URL value as the API expression  
Send the decoded and validated `ExpressionTree` object to the backend  

5. Create Data Request

Create a new `data_request` for the requesting user/team/workspace context  
Link the `data_request` to the generated policy  
Set the data request status to `requested`  
Do not create `team_policy` during request creation  
Do not grant access during request creation  
Do not create `security_scope` rows during request creation  

6. Create Policy From Selected Feature Types

Create a policy representing the requested feature set  
Create one policy statement for each selected feature type  
Scope each policy statement to the selected feature type  
Use a feature-type URN scope for each statement  
Use the current search expression as the fine-grained condition for each statement  

Example statement scopes:

urn:*:telemetry:*
urn:*:species_observation:*
urn:*:dataset:*

7. Link Expression to Each Policy Statement

Persist the current search expression through the existing expression persistence service  
Link the persisted root expression to each generated policy statement  
Reuse the same persisted expression for each statement when the expression is identical  
Do not duplicate expression persistence logic in the request-access service  
Ensure each statement has both a coarse feature-type scope and the linked expression condition  

8. Preserve Policy and Access Separation

Treat the generated policy as the requested feature set  
Treat the data request as the workflow record asking for access to that set  
Do not treat the policy itself as approved access  
Do not create access grants until an admin approves the request  
Do not use `security_scope` for the requested policy until it is linked through `team_policy` after approval  

9. Handle Empty or Missing Expressions

Allow request access from an unfiltered search only if the product flow supports broad access requests  
If broad requests are allowed, create feature-type-scoped statements without an expression link  
If broad requests are not allowed, require a valid search expression before enabling request access  
Do not create an empty expression tree with zero clauses  
Treat omitted expression and `null` expression consistently  

10. Show Success and Failure States

Show a success message after the request is created  
Provide a clear failure message if request creation fails  
Keep the user on the search page or route them to the request detail page according to the product flow  
Prevent duplicate submissions while the request is being created  
Refresh request/access prompt state after successful creation where appropriate  

## Testing Notes

Run all tests

Manual smoke (search-page request-access flow):

Prereq:

`make web && make minio && make queue` 

1. On `/search/result?featureType=telemetry`  A data request Access opens.
2. Submit with a non-empty reason and an empty collaborator picker.
3. In the admin ticket queue, ticket appears with subject `Data Request: telemetry — <first words of reason>`; open it and the new DR is listed in the sidebar with the same reason.


api/src/services/data-request-service.ts
```diff
@@ -128,39 +138,101 @@ export class DataRequestService extends DBService {
...
    * @return {Promise<DataRequest>} Created data request.
    * @memberof DataRequestService
    */
   async createDataRequestForTicket(ticketId: string, payload: CreateDataRequestPayload): Promise<DataRequest> {
-    // Administrative ticket-owned flow: do not create a ticket here.
-    // Create only request-scoped teams and artifacts linked to the provided ticket identifier.
+    // Always include the requester in both teams. The picker captures additional collaborators,
+    // not the canonical access list — the requester is the principal subject of the request and
+    // must receive access if the policy is later approved.
+    const members = Array.from(new Set([payload.requested_by, ...payload.system_user_ids]));
```
mauberti-bc: Your comment makes sense, but we should fix the `POST tickets/{ticketId}/data-request` endpoint to align. That POST endpoint is used by system admins when they manually create a data request inside a ticket. With this PR, we would then be adding the system admin as a data request recipient.

We should update the `requested_by: systemUserId` line to pass the req.body's `requested_by` instead of using the `connection.systemUserId()`, and then keep the `createDataRequestForTicket` unchanged:

```
const dataRequest = await dataRequestService.createDataRequestForTicket(ticketId, {
        requested_by: systemUserId, // change to req.body.requested_by
        reason,
        system_user_ids
      });
```

mauberti-bc (changes requested):
Very nice!

There's a SQL error that I introduced preventing expressions from being saved to the DB. I'll make a new PR for that fix, so after this merges, we can validate the fix in my branch.

api/src/services/data-request-service.ts
dwhogan: Good catch, thanks.

mauberti-bc (approved):

PR #450, opened 2026-05-15
SIMSBIOHUB-1005: Add policy_id to Security Rules

dylanrogowsky-oxd:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1005

## Description of Changes

- Adds nullable `security_rule.policy_id` (FK to `policy`, indexed). Rules without property-based policies may remain unlinked.
- Migration adds the column, FK, and index only; no policy or expression backfill.
- `replaceSecurityRuleExpression` dual-writes security rule and policy statement expression links when `policy_id` is set; policy sync runs on every call and is skipped when `policy_id` is null.
- Exposes nullable `policy_id` on active security rule retrieval (repository, Zod models, OpenAPI).
- Moves security Zod schemas into model files; seeds no longer require a policy per rule.

## Testing Notes

- Ran targeted API tests:
  - `npm test -- src/services/security-rule-expression-service.test.ts src/services/access-policy/policy-statement-expression-service.test.ts`
- Confirmed lint diagnostics are clean for modified files.

### DIFF
--- a/api/src/repositories/security-repository.ts
+++ b/api/src/repositories/security-repository.ts
@@ -23,6 +23,7 @@ export type PersecutionAndHarmSecurity = z.infer<typeof PersecutionAndHarmSecuri
 
 export const SecurityRuleRecord = z.object({
   security_rule_id: z.number(),
+  policy_id: z.string().uuid(),
   name: z.string(),
   description: z.string(),
   record_effective_date: z.string(),
@@ -51,6 +52,7 @@ export type SecurityCategoryRecord = z.infer<typeof SecurityCategoryRecord>;
 
 export const SecurityRuleAndCategory = z.object({
   security_rule_id: z.number(),
+  policy_id: z.string().uuid(),
   name: z.string(),
   description: z.string(),
   record_effective_date: z.string(),
@@ -330,6 +332,7 @@ export class SecurityRepository extends BaseRepository {
     const sql = SQL`
       SELECT 
         sr.security_rule_id,
+        sr.policy_id,
         sr.name,
         sr.description,
         sr.record_effective_date,
@@ -357,7 +360,20 @@ export class SecurityRepository extends BaseRepository {
   async getActiveSecurityRules(): Promise<SecurityRuleRecord[]> {
     defaultLog.debug({ label: 'getActiveSecurityRules' });
     const sql = SQL`
-      SELECT * FROM security_rule WHERE record_end_date IS NULL;
+      SELECT
+        security_rule_id,
+        policy_id,
+        name,
+        description,
+        record_effective_date,
+        record_end_date,
+        create_date,
+        create_user,
+        update_date,
+        update_user,
+        revision_count
+      FROM security_rule
+      WHERE record_end_date IS NULL;
     `;
     const response = await this.connection.sql(sql, SecurityRuleRecord);
     return response.rows;
--- a/api/src/repositories/security-repository.ts
+++ b/api/src/repositories/security-repository.ts
@@ -3,106 +3,23 @@ import { z } from 'zod';
 import { getKnex } from '../database/db';
 import { ApiExecuteSQLError } from '../errors/api-error';
 import { CountResult } from '../models/count';
-import { SecurityCategoryWithRuleCount } from '../models/security-category';
-import { SecurityRuleWithFeatureCount, SecuritySearchFilters } from '../models/security-rule';
+import { ArtifactPersecution, PersecutionAndHarmSecurity } from '../models/persecution-and-harm';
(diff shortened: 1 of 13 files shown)


database/src/migrations/20260515123000_security_rule_policy_id.ts
```diff
@@ -0,0 +1,180 @@
...
+    ALTER TABLE security_rule
+      ADD COLUMN policy_id uuid;
+
+    UPDATE security_rule sr
+    SET policy_id = map.policy_id
+    FROM tmp_security_rule_policy_map map
+    WHERE map.security_rule_id = sr.security_rule_id;
+
+    ALTER TABLE security_rule
+      ALTER COLUMN policy_id SET NOT NULL,
```
mauberti-bc: We should actually let `policy_id` be nullable, because some security rules might not map cleanly to properties (eg. proprietary data). In that case, no need to backfill, so you can remove that from the migration.

api/src/repositories/security-repository.ts
```diff
@@ -23,6 +23,7 @@ export type PersecutionAndHarmSecurity = z.infer<typeof PersecutionAndHarmSecuri
 
 export const SecurityRuleRecord = z.object({
```
mauberti-bc: Suggestion: Move these zod objects into `models/security-rule.ts` for consistency.

api/src/services/security-rule-expression-service.ts
```diff
@@ -46,5 +57,39 @@ export class SecurityRuleExpressionService extends DBService {
...
+      throw new ApiExecuteSQLError('No mapped policy statement found for security rule policy');
+    }
+
+    await this.policyStatementExpressionService.replacePolicyStatementExpression(
+      mappedStatement.policy_statement_id,
+      expressionId
+    );
+  }
+
+  private async getActiveSecurityRuleById(securityRuleId: number): Promise<{ policy_id: string }> {
```
mauberti-bc: nit: add jsdocs to all service/repository methods

mauberti-bc (changes requested):
Nice! A couple changes:

- `replaceSecurityRuleExpression()`: this returns early when `security_rule_expression` is already correct, so it never checks or repairs the matching `policy_statement_expression`. Since this change keeps both links in sync, the policy-side sync should still run.

Consider something like:
```
async replaceSecurityRuleExpression(securityRuleId: number, expressionId: string): Promise<void> {
    const existingLinks = await this.securityRuleExpressionRepository.getSecurityRuleExpressionsBySecurityRuleId(
      securityRuleId
    );
    const alreadyLinked = existingLinks.length === 1 && existingLinks[0].expression_id === expressionId;

    if (!alreadyLinked && existingLinks.length > 0) {
      await this.securityRuleExpressionRepository.deleteSecurityRuleExpressionsBySecurityRuleId(securityRuleId);
    }

    if (!alreadyLinked) {
      await this.securityRuleExpressionRepository.insertSecurityRuleExpression({
        security_rule_id: securityRuleId,
        expression_id: expressionId
      });
    }

    const securityRule = await this.getActiveSecurityRuleById(securityRuleId);
    const policyStatements = await this.policyStatementRepository.getPolicyStatements(securityRule.policy_id);
    const mappedStatement = this.getMappedPolicyStatement(policyStatements);

    if (!mappedStatement) {
      throw new ApiExecuteSQLError('No mapped policy statement found for security rule policy');
    }

    await this.policyStatementExpressionService.replacePolicyStatementExpression(
      mappedStatement.policy_statement_id,
      expressionId
    );
  }
```

- `getMappedPolicyStatement()`: this falls back to “any allow statement” or the first statement if `urn:*:*:*` is missing. That could update the wrong policy statement if the linked policy later has multiple statements. It would be safer to require the exact mapped statement.

```
return policyStatements.find(
      (statement) => statement.effect === PolicyEffect.ALLOW && statement.submission_feature_urn === 'urn:*:*:*'
    );
```

api/src/services/security-rule-expression-service.ts
dylanrogowsky-oxd: Added

api/src/repositories/security-repository.ts
dylanrogowsky-oxd: Moved

database/src/migrations/20260515123000_security_rule_policy_id.ts
dylanrogowsky-oxd: Revised to be nullable and removed the backfill

dylanrogowsky-oxd:
I have also refactored `replaceSecurityRuleExpression()`

mauberti-bc (approved):
Looks great

PR #451, opened 2026-05-15
SIMSBIOHUB-1016: Bug Fix Trigger Validating Predicate Property Types

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1016

## Description of Changes

- There was an existing trigger ensuring every active `predicate` stores a `feature_property_type_id` that matches the property it points at.
    - This prevents invalid expression predicates, like a string predicate being attached to a numeric property.
- Problem: the trigger reads `feature_property_type_id` from `feature_type_property`, but that column only exists on `feature_property`.
- Solution: validate through `feature_property` and separately check that optional `feature_type_property_id` belongs to the same property.

## Testing Notes

- Test by creating a data request with an expression after https://github.com/bcgov/biohubbc-platform/pull/447 merges


dwhogan (approved):
👍

PR #455, opened 2026-05-21
SIMSBIOHUB-1011: BioHub API Key Access

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1011](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1011)

## Description of Changes

Adds API key management to the user portal, allowing scripts and external systems to authenticate against dedicated BioHub routes without an interactive login session.

**Database**

- New `api_key` table with BioHub-standard audit columns (`create_date`, `create_user`, `update_date`, `update_user`, `revision_count`), plus `key_prefix`, `key_hash`, `expires_at`, `revoked_at`, `last_used_at`, and `record_end_date`. Includes audit/journal triggers and a FK to `system_user`.

**API key format and security**

- Keys follow the format `biohub_<8-char prefix>_<base64url secret>`. Only `key_prefix` and a `scrypt` hash (salted with `key_prefix`) are stored — the plaintext is shown once and never persisted. Verification uses positional parsing (immune to `_` in base64url) and timing-safe comparison.

**Authentication**

- New `ApiKey` OpenAPI security scheme. Requests to API-key-gated routes supply the key via `X-API-Key`. On success, a synthetic Keycloak-like token is injected so downstream user-context resolution works without touching existing JWT routes.
- `last_used_at` is updated on every successful authentication.

**API endpoints**

| Method | Route | Purpose |
|--------|-------|---------|
| `GET` | `/api/api-key` | List keys for the current user |
| `POST` | `/api/api-key` | Create a new key (returns plaintext once) |
| `PATCH` | `/api/api-key/{apiKeyId}` | Revoke — sets `revoked_at` and `expires_at` to now |
| `DELETE` | `/api/api-key/{apiKeyId}` | Soft-delete — also sets `record_end_date` so the row disappears from the list |

**Portal UI (`/portal/api-key`)**

- New *API Keys* tab under the user portal, following the same header/table layout as the Tickets and Submissions tabs.
- Table columns: Name, Key Prefix (monospace), Status chip (Active / Revoked / Expired), Created, Expires, Last Used.
- **Active keys** — ellipsis menu has *Revoke* and *Delete*.
- **Revoked / Expired keys** — ellipsis menu has *Delete* only.
- **Create** — "New API Key" button opens a modal. On success, the plaintext key is shown in a one-time copy view with a clipboard button.
- Each action has a confirmation dialog before making any changes.

**Key lifecycle rules**

- *Revoke*: sets `revoked_at = now()` and `expires_at = now()` — key immediately stops authenticating but stays visible in the list.
- *Delete*: sets `record_end_date = now()`, `expires_at = now()`, and `revoked_at = COALESCE(revoked_at, now())` — soft-deletes the row so it no longer appears in the list.
- New keys expire 6 months from creation.

## Testing Notes
- Verify a revoked key is rejected with 401 immediately after revoking.
- Verify an expired key (set `expires_at` in the past directly in the DB) is also rejected with 401.
- Confirm the plaintext key is not returned anywhere after the initial create response (check the list endpoint and the revoke endpoint).
- Confirm existing `GET /api/download-export/{exportId}` (Bearer-secured) is unaffected.
- On the portal, confirm the API Keys tab appears alongside Tickets and Submissions with consistent header, search, and table styling.


api/src/paths/script/download-export/{exportId}/index.ts
```diff
@@ -0,0 +1,89 @@
...
+import { Operation } from 'express-openapi';
+import { getDBConnection } from '../../../../database/db';
+import { DownloadStatusEnum } from '../../../../models/download-status';
+import { DownloadExportDetailResponseSchema } from '../../../../openapi/schemas/download-export';
+import { defaultErrorResponses } from '../../../../openapi/schemas/http-responses';
+import { authorizeRequestHandler } from '../../../../request-handlers/security/authorization';
+import { DownloadExportService } from '../../../../services/download/download-export-service';
+import { getLogger } from '../../../../utils/logger';
+
+const defaultLog = getLogger('paths/script/download-export/{exportId}');
```
mauberti-bc: I don't think we want this endpoint. The flow will be:
- User generates an API key
- User can request `GET download/{downloadId}/presigned` using the API key
    - This responds with a presigned url for that specific download's artifact

The API key is only relevant to the `download` record, not the exports of a download.

You could refactor this file to be the `GET download/{downloadId}/presigned` endpoint. (should this be GET?)

api/src/repositories/api-key-repository.ts
```diff
@@ -0,0 +1,227 @@
...
+   *
+   * Returns `null` if no active key matches the prefix.
+   * Used during API-key authentication — the full `key_hash` is included so the caller
+   * can verify the supplied plaintext key.
+   *
+   * @param {string} keyPrefix - The `key_prefix` value (e.g. `biohub_AbCdEfGh`).
+   * @return {Promise<ApiKey | null>}
+   * @memberof ApiKeyRepository
+   */
+  async getApiKeyByPrefix(keyPrefix: string): Promise<ApiKey | null> {
```
mauberti-bc: use `find*` if the method returns null.
- `get` prefix implies that the method returns the type or throws an error if the rowCount isn't 1.
- `find` prefix implies response might be null

app/src/features/portal/PortalApiKeysPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+        isLoading={keysLoader.isLoading}
+        searchTerm={searchTerm}
+        onSearch={setSearchTerm}
+        onAdd={handleOpenCreate}
+        onRevoke={setRevokeTarget}
+        onDelete={setDeleteTarget}
+      />
+
+      {/* New API Key dialog */}
+      <Dialog open={createDialogOpen} onClose={handleCloseCreate} maxWidth="sm" fullWidth>
```
mauberti-bc: Suggestion: Use existing dialog components, like `OkDialog` if there's no data collection, or `EditDialog` if there is a form (which there is in this case).

Also, break this big `PortalApiKeysPage` component down - create a dedicate PortalApiKeyDialog component

app/src/features/portal/PortalApiKeysPage.tsx
```diff
@@ -0,0 +1,258 @@
...
+    setCopied(false);
+    keysLoader.refresh();
+  };
+
+  /**
+   * Submit a new API key creation request.
+   *
+   * On success, `createResult` is set so the dialog switches to the one-time plaintext view.
+   */
+  const handleCreate = async () => {
```
mauberti-bc: Use formik and the `EditDialog` for consistency with existing code. The `CreateTicketDialog` and `TicketsPage` is an example to follow.

mauberti-bc (changes requested):
Very nice!

api/src/paths/script/download-export/{exportId}/index.ts
dylanrogowsky-oxd: This endpoint was for testing and ends up getting removed as part of PR-458, but I'll remove it from this PR.

api/src/repositories/api-key-repository.ts
dylanrogowsky-oxd: Updated the naming

app/src/features/portal/PortalApiKeysPage.tsx
dylanrogowsky-oxd: Adjusted to use the existing dialog components, and a new `PortalApiKeyDialog` component

app/src/features/portal/PortalApiKeysPage.tsx
dylanrogowsky-oxd: Updated to use Formik with `EditDialog`

mauberti-bc (approved):
👍

PR #458, opened 2026-05-22
SIMSBIOHUB-1012: API access to parquet downloads

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1012](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1012)

## Description of Changes

Adds a presigned-URL endpoint for Parquet downloads, accessible via API key.

**New endpoint: `GET /api/download/{downloadId}/presigned-url`**

- Authenticated via `X-API-Key` header (API key security scheme).
- Runs the same team-membership authorization check used by JWT download routes — the API key does not bypass access control.
- Returns presigned S3 URLs (30-minute expiry) for each Parquet artifact belonging to the download, one per feature type.
- Only returns URLs when the download status is `ready` or `downloaded`; all other statuses produce a 409 so scripts can distinguish "not ready yet" from access errors.
- Response shape per part: `{ feature_type, url, expires_at }`.

**Removed**

- `GET /api/script/download-export/{exportId}` — script route for CSV exports. The JWT-protected `GET /api/download-export/{exportId}` is unchanged.

## Testing Notes

- Create an API key via the portal API Keys tab, copy the plaintext key, and call `GET /api/download/{downloadId}/presigned-url` with `X-API-Key: <key>`. Confirm parts are returned with valid presigned URLs and a 30-minute expiry.
- Confirm a 409 is returned for a download in `pending` or `processing` status.
- Confirm a 403 is returned when the API key owner is not on the download's team.
- Confirm a 404 is returned for a non-existent download ID.
- Confirm the existing `GET /api/download-export/{exportId}` (Bearer-secured) is unaffected.

### DIFF
--- a/api/src/constants/download.ts
+++ b/api/src/constants/download.ts
@@ -3,6 +3,12 @@ import { getNumberEnv } from '../utils/env-utils';
 export const SIGNED_URL_EXPIRY_DOWNLOAD = 432000; // 5 days
 
 /**
+ * Presigned URL expiry for the API-key presigned-URL endpoint (30 minutes).
+ * Short-lived by design — scripts should re-request a fresh URL per download attempt.
+ */
+export const SIGNED_URL_EXPIRY_PARQUET_DOWNLOAD = 60 * 30; // 30 minutes
+
+/**
  * Default `download_export.max_part_size_bytes` (500 MB). Applied by the service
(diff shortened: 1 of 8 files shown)


mauberti-bc (approved):
👍

PR #459, opened 2026-05-25
SIMSBIOHUB-1017: Allow Anonymous Downloads

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1017](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1017)

## Description of Changes

- Anonymous users can create a download package from public search results without signing in.
- The user can name and describe the download before creating it.
- The created download is identified by a UUID and can be accessed by anyone who has the link.
- Public-by-link downloads must never include secured features, secured child features, or data protected by security rules.
- Anonymous download generation must evaluate data visibility as an anonymous user, so only non-secured data is packaged.
- Anonymous downloads should be stored as `download` records without `download_team` associations, making the UUID the access credential.
- `POST /api/download` should support anonymous creation for public-only packages.
- `GET /api/download/:downloadId` should continue to allow UUID-based access to anonymous downloads.
- Authenticated download behavior should remain unchanged, including team-linked downloads through `download_team`.
- Users with the link can check download status and retrieve the generated package without authentication.

## Testing Notes

**Combined FE + BE — restores anonymous download create

run tests

Manual smoke (anon-download flow):

1. Signed-out on `/search/dataset`, Create Download opens the form; submit renders a persistent dialog with a `/api/download-export/{exportId}` status URL and a copy button.
2. Anon API path — no bearer end to end:

```bash
# create — no Authorization header
curl -sX POST http://localhost:6100/api/download \
  -H 'Content-Type: application/json' \
  -d '{"name":"anon test","description":null,"featureTypes":["dataset"],"expression":null}'
# → 201; export_url non-null (the anon credential)

# poll the returned export_url — still no bearer
curl -s http://localhost:6100/api/download-export/<export_id>
# → 200; parts[] empty until status:"ready", then signed part URLs
```


api/src/paths/download/index.ts
```diff
@@ -195,6 +192,16 @@ POST.apiDoc = {
               download_url: {
                 type: 'string',
                 description: 'Fully-qualified API URL the caller can poll for status'
+              },
+              export_id: {
```
mauberti-bc: A download can theoretically be exported multiple times and to various file formats, so I don't think a single `export_id` and `export_url` is desired. See later comments about giving anonymous users access to the download page.

api/src/paths/download/index.ts
```diff
@@ -206,16 +213,24 @@ POST.apiDoc = {
...
- * the transaction boundary, and response shaping.
+ * Without a bearer token the request runs on the shared API-user connection and `requestedBy`
+ * is null. A null `requested_by` is the security identity "anonymous" — the parquet packaging
+ * filters to only unsecured data, so an anonymous caller can never pull secured records. An
+ * authenticated request runs on that user's connection and scopes the package to data visible
+ * to them; authorization is re-evaluated at export time against `requested_by`, so queuing
+ * early grants no extra access later.
+ *
+ * The two callers also differ in how they retrieve their package. An anonymous caller has no
+ * Downloads UI and no user identity to look the download up by, so the response hands back an
```
mauberti-bc: Anonymous users should be able to view the specific download in the browser, in which case we can remove the `export_*` properties:
- User clicks download
- Browser navigates to `/download/<uuid>`
- User can export the download from there

We can cache the downloads for the user, and eventually send emails. But until then, if the user closes the download's page, they would lose knowledge of that download - this is fine for now.

api/src/queue/jobs/process-download-job.ts
```diff
@@ -126,6 +128,16 @@ export const processDownloadJobHandler: PgBoss.WorkHandler<IProcessDownloadJobDa
         await new DownloadPipelineService(connection).transitionDownloadStatus(downloadId, DownloadStatusEnum.READY, [
           DownloadStatusEnum.PROCESSING
         ]);
+
+        // Anonymous downloads have no UI to request an export, so the default export created at
+        // request time is enqueued automatically once the parquet lands. Authenticated downloads
+        // keep the explicit two-call export flow.
+        if (source.requested_by === null) {
+          const [defaultExport] = await new DownloadExportService(connection).listExportsByDownloadId(downloadId);
```
mauberti-bc: No need to start the export during the download job

api/src/services/download/download-export-service.ts
```diff
@@ -72,6 +78,15 @@ export class DownloadExportService extends DBService {
     // Throws HTTP403 / HTTP404 as appropriate.
     const download = await this.downloadService.getAuthorizedDownload(downloadId, systemUserId);
 
+    // Anonymous (teamless) downloads get exactly one auto-chained default export and have no UI
+    // to request more — and the parquet→export auto-chain assumes that single export. Refuse
+    // explicit creation on a teamless download: an authenticated holder of the anon UUID must
+    // claim it first (which links a team) before requesting additional exports.
+    const isClaimed = await this.downloadService.isDownloadClaimedByTeam(downloadId);
+    if (!isClaimed) {
+      throw new HTTP409('Anonymous downloads cannot create additional exports — claim the download first');
```
mauberti-bc: We should allow multiple exports, even for anonymous downloads. 

This is out of scope for this ticket, but useful context:
- User clicks Export on download
- We create a new `download_export` record
- We join the new `download_export` to the existing `download_export_artifact`?

We don't want to needlessly recreate the CSV export, but if a user clicks Export, they should get their own export record with an accurate create_date. 

We might need another layer to reduce file storage, so we're not joining numerous csv fragment records for each export.

api/src/services/download/download-service.ts
```diff
@@ -17,7 +18,19 @@ export interface CreateDownloadRequestPayload {
...
+
+/**
+ * Result of creating a download request.
+ *
+ * `export_id` is non-null only for anonymous requests, where the default export is
+ * created up front so the single returned status URL is stable; it is null for
+ * authenticated requests, which create their export later through the UI.
+ */
+export interface CreateDownloadRequestResult {
+  download_id: string;
```
mauberti-bc, line 27: Clients only needs the `download_id` to navigate to

app/src/features/search/result/hooks/useSearchResultDownload.tsx
```diff
@@ -97,11 +102,32 @@ export const useSearchResultDownload = ({
...
-            open: true,
-            snackbarMessage: 'Download created. Track its progress in the Downloads sidebar.'
-          });
+          if (auth.isAuthenticated) {
+            setDownloadView(DOWNLOAD_SIDEBAR_VIEW.DOWNLOADS);
+            dialogContext.setSnackbar({
+              open: true,
+              snackbarMessage: 'Download created. Track its progress in the Downloads sidebar.'
+            });
+          } else if (response.export_url) {
```
mauberti-bc: Remove `response.export_url` and just navigate to `download/<download uuid>` upon success.

mauberti-bc (changes requested):
Nice! Main change is removing mentions of `export`. Anonymous downloads will be accessible to anybody with the UUID, and we should navigate to `download/<uuid>` after receiving the response.

app/src/features/download/PublicDownloadPage.tsx
```diff
@@ -0,0 +1,133 @@
...
+export const PublicDownloadPage = () => {
+  const { downloadId } = useParams<{ downloadId: string }>();
+  const api = useApi();
+  const downloadDataLoader = useDataLoader((id: string) => api.download.getDownload(id));
+
+  useEffect(() => {
+    if (downloadId) {
+      downloadDataLoader.load(downloadId);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mauberti-bc: nit: don't need the eslint ignore line for `.load`, only `refresh`.

mauberti-bc (dismissed):

api/src/paths/download/index.ts
dwhogan: fixed

api/src/paths/download/index.ts
dwhogan: fixed, thanks

api/src/services/download/download-export-service.ts
dwhogan: fixed

app/src/features/search/result/hooks/useSearchResultDownload.tsx
dwhogan: fixed

mauberti-bc (approved):

api/src/queue/jobs/process-download-job.ts
dwhogan: fixed

api/src/services/download/download-service.ts
dwhogan: fixed

app/src/features/download/PublicDownloadPage.tsx
dwhogan: thanks, fixed

PR #460, opened 2026-05-25
SIMSBIOHUB-1018: CRUD APIs for Feature Types and Properties

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1018](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1018)
- [SIMSBIOHUB-1019](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1019)
- [SIMSBIOHUB-1020](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1020)

## Description of Changes

Adds system-admin CRUD APIs for feature types, feature properties, and feature-type-property assignments. Backend-only — no admin UI in this PR.

**New admin endpoints** (all require System Admin Bearer auth):

| Resource | Routes |
|----------|--------|
| Feature types | `GET/POST /api/administrative/feature-types`, `GET/PUT/DELETE /api/administrative/feature-types/{featureTypeId}` |
| Feature properties | `GET/POST /api/administrative/feature-properties`, `GET/PUT/DELETE /api/administrative/feature-properties/{featurePropertyId}` |
| Assignments | `GET/POST /api/administrative/feature-types/{featureTypeId}/properties`, `GET/PUT/DELETE .../properties/{featureTypePropertyId}` |

List endpoints support pagination and search (feature types/properties). Assignments are scoped to the parent feature type on nested routes. Deletes are soft deletes via `record_end_date`.

**Supporting changes**

- New repositories, services, models, and OpenAPI schemas for the above resources.
- `/api/codes` now includes `description` on feature types and nullable `description` on properties.
- `feature_type.sort` and admin CRUD `allow_multiple` are **not** exposed in the API (DB columns unchanged; inserts rely on DB defaults). List ordering still uses `sort` internally.

## Testing Notes

- Authenticate as a system admin and exercise full CRUD for each resource (create → get → list → update → delete).
- For feature property create, use a valid `feature_property_type_id` from the database.
- For assignment create, verify parent feature type and feature property must exist; duplicate assignments return 409.
- Confirm nested assignment routes reject cross–feature-type access (wrong `featureTypeId` → 404).
- Run API tests: `npm test -- --grep "FeatureType|FeatureProperty|createFeatureType|CodeRepository"`.
- Confirm `/api/codes` still returns expected feature type/property metadata.

### DIFF
--- a/api/src/repositories/feature-type-property-repository.ts
+++ b/api/src/repositories/feature-type-property-repository.ts
@@ -1,8 +1,33 @@
 import SQL from 'sql-template-strings';
-import { ApiExecuteSQLError, ApiNotFoundError } from '../errors/api-error';
-import { ExpressionPredicatePropertyMetadata, FeatureTypeProperty } from '../models/feature-type-property';
+import { getKnex } from '../database/db';
+import { ApiConflictError, ApiExecuteSQLError, ApiNotFoundError } from '../errors/api-error';
+import { CountResult } from '../models/count';
+import {
+  AdminFeatureTypeProperty,
+  CreateFeatureTypePropertyRecord,
+  ExpressionPredicatePropertyMetadata,
+  FeatureTypeProperty,
+  UpdateFeatureTypePropertyRecord
+} from '../models/feature-type-property';
+import { ApiPaginationOptions } from '../zod-schema/pagination';
 import { BaseRepository } from './base-repository';
 
+/** Columns selected for admin CRUD responses (raw feature_type_property row). */
+const ADMIN_FEATURE_TYPE_PROPERTY_COLUMNS = [
+  'feature_type_property_id',
+  'feature_type_id',
+  'feature_property_id',
+  'required_value',
+  'sort'
+] as const;
+
+/**
+ * A repository class for accessing feature type property data.
+ *
+ * @export
+ * @class FeatureTypePropertyRepository
+ * @extends {BaseRepository}
+ */
 export class FeatureTypePropertyRepository extends BaseRepository {
   /**
    * Get a feature property record by canonical feature-property name.
@@ -11,6 +36,7 @@ export class FeatureTypePropertyRepository extends BaseRepository {
    * @return {Promise<FeatureTypeProperty>} Matching feature-type-property row.
    * @throws {ApiNotFoundError} If no active feature property matches the name.
    * @throws {ApiExecuteSQLError} If more than one active row is returned.
+   * @memberof FeatureTypePropertyRepository
    */
   async getFeaturePropertyByName(featurePropertyName: string): Promise<FeatureTypeProperty> {
     const sqlStatement = SQL`
@@ -66,6 +92,7 @@ export class FeatureTypePropertyRepository extends BaseRepository {
    * @return {Promise<FeatureTypeProperty>} Matching feature-type-property row.
    * @throws {ApiNotFoundError} If no active row exists for the id.
    * @throws {ApiExecuteSQLError} If more than one active row is returned.
+   * @memberof FeatureTypePropertyRepository
    */
   async getFeaturePropertyByFeatureTypePropertyId(featureTypePropertyId: number): Promise<FeatureTypeProperty> {
     const sqlStatement = SQL`
@@ -122,6 +149,9 @@ export class FeatureTypePropertyRepository extends BaseRepository {
    * @param {number} featurePropertyId - Shared feature property id.
    * @param {number | null} featureTypePropertyId - Optional concrete feature type property id.
    * @return {Promise<ExpressionPredicatePropertyMetadata>} Resolved metadata.
+   * @throws {ApiNotFoundError} If no matching active property metadata exists.
+   * @throws {ApiExecuteSQLError} If an unexpected row count is returned.
+   * @memberof FeatureTypePropertyRepository
    */
   async getExpressionPredicatePropertyMetadata(
     featurePropertyId: number,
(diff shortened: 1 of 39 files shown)


api/src/repositories/feature-type-property-repository.ts
```diff
@@ -180,4 +210,221 @@ export class FeatureTypePropertyRepository extends BaseRepository {
...
+  /**
+   * Insert a new feature type property record.
+   *
+   * @param {CreateFeatureTypePropertyRecord} data - Data for the record to insert.
+   * @return {Promise<number>} The new feature_type_property_id.
+   * @throws {ApiConflictError} If the feature property is already assigned to the feature type.
+   * @throws {ApiExecuteSQLError} If the insert does not affect exactly one row.
+   * @memberof FeatureTypePropertyRepository
+   */
+  async insertFeatureTypeProperty(data: CreateFeatureTypePropertyRecord): Promise<number> {
```
mauberti-bc: Repository methods should call `await this.connection.knex` once. The service layer should orchestrate if we need multiple operations.

api/src/repositories/feature-type-property-repository.ts
```diff
@@ -180,4 +210,221 @@ export class FeatureTypePropertyRepository extends BaseRepository {
...
+      .where({
+        feature_type_id: data.feature_type_id,
+        feature_property_id: data.feature_property_id
+      })
+      .first();
+
+    const existingResult = await this.connection.knex(existingQuery);
+
+    if ((existingResult.rowCount ?? 0) > 0) {
+      throw new ApiConflictError('Feature property is already assigned to this feature type', [
```
mauberti-bc: This sort of check should be the service layer's responsibility

api/src/services/feature-type-property-service.ts
```diff
@@ -31,8 +50,109 @@ export class FeatureTypePropertyService extends DBService {
...
+   *
+   * @param {CreateFeatureTypePropertyRecord} data - Fields to insert.
+   * @return {Promise<AdminFeatureTypeProperty>} The created record.
+   * @throws {ApiNotFoundError} If the parent feature type or feature property does not exist.
+   * @throws {ApiConflictError} If the feature property is already assigned to the feature type.
+   * @throws {ApiExecuteSQLError} If the insert does not affect exactly one row.
+   * @memberof FeatureTypePropertyService
+   */
+  async createFeatureTypeProperty(data: CreateFeatureTypePropertyRecord): Promise<AdminFeatureTypeProperty> {
+    const featureTypeRepository = new FeatureTypeRepository(this.connection);
```
mauberti-bc: Prefer instantiating classes in the constructor, not within methods, for consistency.

api/src/services/feature-type-property-service.ts
```diff
@@ -31,8 +50,109 @@ export class FeatureTypePropertyService extends DBService {
...
+    const featurePropertyRepository = new FeaturePropertyRepository(this.connection);
+
+    await featureTypeRepository.getFeatureType(data.feature_type_id);
```
mauberti-bc: nit: `Promise.all` for independent calls like this.

api/src/services/feature-property-service.ts
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Create a feature property record.
+   *
+   * @param {CreateFeatureProperty} data - Feature property fields required to create the record.
+   * @return {Promise<FeatureProperty>} The created feature property (with resolved type_name).
+   * @throws {ApiExecuteSQLError} If the insert does not affect exactly one row.
+   * @memberof FeaturePropertyService
+   */
+  async createFeatureProperty(data: CreateFeatureProperty): Promise<FeatureProperty> {
+    const featurePropertyId = await this.featurePropertyRepository.insertFeatureProperty(data);
```
mauberti-bc, line 40: Consider validating that `feature_property_type_id` exists and is valid first, and return useful error code.

api/src/services/feature-property-service.ts
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @param {number} featurePropertyId - Feature property identifier.
+   * @param {UpdateFeatureProperty} data - Partial feature property fields to update.
+   * @return {Promise<FeatureProperty>} Updated feature property (with resolved type_name).
+   * @throws {ApiExecuteSQLError} If the update does not affect exactly one row.
+   * @throws {ApiNotFoundError} If no active feature property exists for the id.
+   * @memberof FeaturePropertyService
+   */
+  async updateFeatureProperty(featurePropertyId: number, data: UpdateFeatureProperty): Promise<FeatureProperty> {
+    await this.featurePropertyRepository.updateFeatureProperty(featurePropertyId, data);
```
mauberti-bc, line 94: We can support updating the `feature_property_type_id` for a `feature_property`, but in case the old record has real data, we want to re-index that property from all the features that use it. Otherwise a `feature_property` could be both `geometry` and `string` (which might be ok, but it seems risky).

Suggestion: Remove `feature_property_type_id` from the UpdateFeatureProperty type, and stop updating `feature_property_type_id`.

mauberti-bc (changes requested):
Looks good, just a few minor comments!

api/src/services/feature-type-property-service.ts
dylanrogowsky-oxd: Fixed. The two FK validation calls are now wrapped in Promise.all since they are independent.

api/src/repositories/feature-type-property-repository.ts
dylanrogowsky-oxd: A new repository method, `findActiveFeatureTypePropertyByFeatureTypeAndProperty`, was added to perform the lookup as a single, self-contained knex call.

api/src/repositories/feature-type-property-repository.ts
dylanrogowsky-oxd: The `ApiConflictError` throw that guarded against duplicate assignments was removed from `insertFeatureTypeProperty`. It is now in `FeatureTypePropertyService.createFeatureTypeProperty`, which calls `findActiveFeatureTypePropertyByFeatureTypeAndProperty` and throws `ApiConflictError` if a row is returned.

api/src/services/feature-property-service.ts
dylanrogowsky-oxd: A new repository method, `getFeaturePropertyTypeById`, was added to `FeaturePropertyRepository`. It queries the `feature_property_type` table (active rows only) and throws `ApiNotFoundError` if the ID does not exist.

mauberti-bc (approved):
👍

PR #461, opened 2026-05-27
SIMSBIOHUB-948: CRUD operations for Security Categories and Reasons

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-948](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-948)

## Description of Changes

Adds full CRUD for security categories and reasons on the Manage Security admin page (`/admin/security`). The page was read-only before; admins can now add, edit, and delete both resource types from the UI.

**API** (System Admin or Data Administrator):

| Resource | Routes |
|----------|--------|
| Categories | `POST /api/administrative/security/categories`, `GET/PUT/DELETE .../categories/{securityCategoryId}` |
| Reasons | `POST /api/administrative/security/reasons`, `GET/PUT/DELETE .../reasons/{securityRuleId}` |

Deletes are soft deletes. Category delete returns 409 if the category still has active reasons. Reason delete returns 409 if the reason is still applied to one or more submission features. New reasons are created with `policy_id` null. The reasons list response now includes `security_category_id` and `category_name`, and the feature count only reflects active (non-soft-deleted) applications.

**API architecture**

The security layer has been split from a single monolithic service/repository pair into dedicated, per-table classes:

| Class | Responsibility |
|-------|----------------|
| `SecurityCategoryRepository` | `security_category` table CRUD |
| `SecurityCategoryService` | Category business logic and delete guard |
| `SecurityRuleRepository` | `security_rule` table CRUD + `getActiveAppliedFeatureCount` |
| `SecurityRuleService` | Rule business logic, delete guard (`assertSecurityRuleIsUnused`) |
| `SecurityRepository` | Cross-cutting methods (persecution/harm, submission-feature-security apply/remove) |
| `SecurityService` | Cross-cutting service logic (artifact security, scope recomputation) |

Internally, all types and method names use `Rule` to match the `security_rule` database table (`SecurityRule`, `CreateSecurityRule`, `UpdateSecurityRule`). The external REST routes (`/reasons`), OpenAPI schema titles, and UI remain unchanged.

The `getSecurityRulesWithFeatureCount` query now uses a join callback with `andOnNull('sfs.record_end_date')` on the `submission_feature_security` join, so soft-deleted applications are excluded from the count while rules with zero applications still return `feature_count = 0`.

**UI**

- Page title updated to **Manage Security**
- Categories and Reasons sections follow the same pattern as Policies/Teams: search, Add button, Actions column, `EditDialog` + Formik forms
- Reason create/edit includes a required Category dropdown
- Categories table refreshes when reasons are created, updated, or deleted so the Reasons count column stays accurate

## Testing Notes

- Log in as a system admin and open `/admin/security`
- Categories: add, edit, and delete a category with no reasons; confirm delete is blocked (409) when reasons exist
- Reasons: add, edit, and delete a reason; confirm category counts update on the categories table; confirm delete is blocked (409) when the reason is applied to features
- Confirm the reasons list feature count does not include soft-deleted applications
- Run `make test` (or targeted: `npm test -- --grep "administrative/security"` in `api/`, security component tests in `app/`)


api/src/services/security-service.ts
```diff
@@ -540,4 +550,112 @@ export class SecurityService extends DBService {
...
+  }
+
+  /**
+   * Soft-delete a security reason by ID.
+   *
+   * @param {number} securityRuleId
+   * @return {*}  {Promise<void>}
+   * @memberof SecurityService
+   */
+  async deleteSecurityReason(securityRuleId: number): Promise<void> {
```
mauberti-bc: The service method is `deleteSecurityReason` but it calls `deleteSecurityRule`. The database table's name is the authoritative name, which I think is **security rule** -- we should replace "reason" with "rule".

Also, we should split the `security-service.ts` into multiple services/repositories, one for each table, which is our normal convention:
* `security-category-service.ts`
* `security-rule-service.ts`
* `security-category-repository.ts`
* `security-rule-repository`

and each service/repository has its relevant CRUD methods.

api/src/repositories/security-repository.ts
```diff
@@ -593,14 +602,19 @@ export class SecurityRepository extends BaseRepository {
...
       .select(
         'sr.security_rule_id',
+        'sr.security_category_id',
+        'sc.name as category_name',
         'sr.name',
         'sr.description',
         knex.raw('COUNT(sfs.submission_feature_security_id)::integer AS feature_count')
       )
       .from('security_rule as sr')
+      .innerJoin('security_category as sc', function () {
```
mauberti-bc: This query counts soft-deleted `submission_feature_security` records. Add `whereNull('sfs.record_end_date')`?

api/src/services/security-service.ts
mauberti-bc: We should prevent deleting security reasons that are still applied to features. The existing pattern is to use `assert*` service methods for these checks, for example `this.assertNoMetadataConflicts(existingBySlug, payloadBySlug)`.

Suggested changes:
* Add `SecurityRuleService.assertSecurityRuleIsUnused(securityRuleId)`.
* In that method, count active applications:

```
SELECT COUNT(*)
FROM submission_feature_security
WHERE security_rule_id = $1
  AND record_end_date IS NULL;
```

* If count > 0, throw a conflict error.
* Call this before soft-deleting the security rule.
* In a separate PR, we'll have to add for unapplying a rule from features

mauberti-bc (changes requested):

api/src/services/security-service.ts
dylanrogowsky-oxd: - Created `security-category-repository.ts `with all security_category CRUD methods
- Created `security-rule-repository.ts` with all security_rule CRUD methods
- security-repository.ts now only contains the cross-cutting methods and continues to export `SECURITY_APPLIED_STATUS`.
- Created `SecurityCategoryService` wrapping `SecurityCategoryRepository` with the same category methods, including the delete guard (`assertSecurityRuleIsUnused`-style pattern)
- Created `SecurityRuleService` wrapping `SecurityRuleRepository`, with `assertSecurityRuleIsUnused` calling `getActiveAppliedFeatureCount` before `deleteSecurityRule`
- `SecurityService` retains only cross-cutting methods.

mauberti-bc (approved):
👍

PR #462, opened 2026-05-28
SIMSBIOHUB-1021: Export `feature` properties

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1021](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1021)

## Description of Changes

As a BioHub user, I want feature-backed properties to export as referenced feature URNs, so that downloaded CSV and Parquet files contain durable feature references.

Acceptance Criteria

1. Feature Property Export

WHEN a feature-backed property is included in a download/export,
THEN BioHub should export the referenced feature’s URN as the property value.

2. Feature Property Hydration

WHEN export hydration fetches typed property rows,
THEN BioHub should include `feature` properties from `submission_feature_property_feature`.
WHEN a feature-backed property row is hydrated,
THEN BioHub should join to the referenced `submission_feature` to resolve its URN.

3. CSV and Parquet Output

WHEN a CSV export includes a feature-backed property,
THEN the exported value should be the referenced feature’s URN.
WHEN a Parquet export includes a feature-backed property,
THEN the exported value should be the referenced feature’s URN.

4. Null Handling

WHEN the referenced feature does not exist or is inactive,
THEN the exported property value should be `null`.
Note: This should never happen

## Testing Notes

**Backend only — `feature`-typed property columns become populated in dataset CSV + Parquet exports.**

```bash
make test       # unit (both layers); parquet round-trip suite catches the array/object double-encode regression
make test-db    # ← the one that matters: download-parquet-pipeline integration proves the new `feature` SELECT end-to-end
make test-sys   # broader regression (slower; skip if test-db is green)
```

### DIFF
--- a/api/src/utils/csv-utils.ts
+++ b/api/src/utils/csv-utils.ts
@@ -125,6 +125,7 @@ export function flattenFeatureWithParent(
  * - array → delegate to flattenArray()
  * - artifact_key → files/{submissionFeatureId}_{filename}
  * - object → JSON.stringify(value)
+ * - feature → JSON.stringify(value), '' for null/undefined
  * - null/undefined → empty string
  *
  * @param data - The feature's JSONB data.
@@ -171,6 +172,9 @@ export function flattenFeatureBySchema(
         result[prop.feature_property_name] = Array.isArray(value) ? flattenArray(value) : toStringOrEmpty(value);
         break;
       case 'object':
+        result[prop.feature_property_name] = value == null ? '' : JSON.stringify(value);
+        break;
+      case 'feature':
         result[prop.feature_property_name] = value == null ? '' : JSON.stringify(value);
         break;
       default:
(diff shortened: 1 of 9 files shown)


api/src/utils/csv-utils.ts
```diff
@@ -171,6 +172,7 @@ export function flattenFeatureBySchema(
         result[prop.feature_property_name] = Array.isArray(value) ? flattenArray(value) : toStringOrEmpty(value);
         break;
       case 'object':
+      case 'feature':
         result[prop.feature_property_name] = value == null ? '' : JSON.stringify(value);
```
mauberti-bc, line 176: Consider something like this, so that scalar strings aren't wrapped in quotes in the csv cell:
```
case 'feature':
  if (value == null) {
    result[prop.feature_property_name] = '';
    break;
  }

  if (Array.isArray(value)) {
    result[prop.feature_property_name] = JSON.stringify(value);
    break;
  }

  result[prop.feature_property_name] = toStringOrEmpty(value);
  break;
```

mauberti-bc (changes requested):
Nice!

mauberti-bc (approved):
👍

api/src/utils/csv-utils.ts
dwhogan: fixed

PR #466, opened 2026-06-04
SIMSBIOHUB-1025: Improve Search Performance

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1025](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1025)

## Description of Changes

As a system administrator, I want expression search to use `submission_feature_closure` to determine whether two features are related, so that search no longer performs recursive graph traversal at query time.

### Background

Expression search currently resolves related features by recursively traversing feature relationships.

The new closure table precomputes those relationships. Search should use the closure table as a lookup:

> Is evidence feature A related to target feature B?

Instead of recursively walking from A to B during the search request, the query should check for an existing row:

- `submission_feature_closure.source_submission_feature_id` = evidence feature id
- `submission_feature_closure.target_submission_feature_id` = target feature id

### Scope

- Replace recursive relationship traversal in expression search with closure-table lookup
- Replace the recursive ancestor walk in the shared security fragments (`isEffectivelySecured` / `isAccessibleToUser` in `sql-fragments.ts`) with a `submission_feature_closure` lookup over the `is_ancestor = true` subset — in place, so all callers switch (see AC #3)
- Preserve existing expression search behaviour
- Preserve evidence and target security filtering
- Preserve pagination and count behaviour
- Keep the public search API unchanged

### 1. Use Closure Lookup for Relationship Checks

WHEN expression search needs to determine whether an evidence feature is related to a target feature
THEN it should check `submission_feature_closure`
AND it should not recursively traverse feature relationships at query time

### 2. Remove Runtime Recursive Traversal

WHEN expression search runs
THEN it should not dynamically recurse through:
- `submission_feature_feature`
- `parent_submission_feature_id`
- `submission_feature_property_feature`

AND it should rely on `submission_feature_closure` to determine feature relatedness

### 3. Use Closure Lookup for the Security Ancestor Walk

WHEN a security check needs the ancestor chain of a feature (to find an active security rule or a reachable scope anchor)
THEN the shared fragments `isEffectivelySecured` and `isAccessibleToUser` should resolve ancestors via `submission_feature_closure` (`source_submission_feature_id` = the feature, `WHERE is_ancestor = true`)
AND they should not use `WITH RECURSIVE ancestor_chain` to walk `parent_submission_feature_id` at query time

### 4. Preserve Security Semantics Across All Callers

WHEN the rewritten fragments are used by any caller — expression search, filter search, cart, and the authorization scope-anchor recompute job
THEN each caller returns the same visibility/security result as the recursive implementation
AND anonymous-vs-authenticated, effective-date, and scope-grant semantics are unchanged

## Implementation Notes

Current conceptual shape:

```
predicate
  -> evidence feature ids
  -> recursive traversal
  -> target feature ids
```

Target conceptual shape:

```
predicate
  -> evidence feature ids
  -> closure table lookup
  -> target feature ids
```

The core lookup should be equivalent to:

```sql
JOIN submission_feature_closure c
  ON c.source_submission_feature_id = evidence.submission_feature_id
JOIN submission_feature target_sf
  ON target_sf.submission_feature_id = c.target_submission_feature_id
```

The public expression tree API should not change.

## Out of Scope

- This ticket assumes `submission_feature_closure` is already populated by the pg-boss closure rebuild job.
- This ticket does not update download expansion.

## Testing Notes

**Backend only**

```bash
make test
make test-db
make test-sys 
```

### DIFF
--- a/api/src/repositories/sql-fragments.ts
+++ b/api/src/repositories/sql-fragments.ts
@@ -22,14 +22,15 @@ import { Knex } from 'knex';
  * `target_submission_feature_id`. This matters because it runs on every search, cart,
  * and download request.
  *
- * Requires the closure to be populated for the feature's upload. This holds on every
- * caller: features only become searchable after indexing rebuilds the closure, and the
- * anchor recompute write path is only ever triggered for already-indexed submissions
- * (applying a security rule or approving a policy never changes a feature's ancestry,
- * so the closure built at index time is still current). Any feature reaching this check
- * therefore already has its closure rows (at minimum the self-loop). Callers that
- * operate on un-indexed or possibly-inactive ids (e.g. raw id arrays) must filter to
- * active features first, since the closure has no row for soft-deleted features.
+ * Fails closed on missing closure rows. The closure is built by an async recompute job
+ * that runs *after* an upload is flipped to `indexed` (and therefore searchable), in a
+ * separate transaction; a recompute that has not yet run, or that failed, does not revert
+ * that status. So an active, searchable, secured feature can transiently — or, on a failed
+ * recompute, indefinitely — have no closure rows. Rather than read "no rows" as "unsecured"
+ * (which would leak the feature), this check treats the absence of any closure ancestry row
+ * as secured: if we cannot prove a feature is unsecured, we hide it. Under normal operation
+ * every active feature carries at least its self-loop `(F, F)`, so the fail-closed branch is
+ * inert on the happy path.
  *
  * Returns an `EXISTS (...)` SQL expression (returns boolean) with zero `?` placeholders.
  *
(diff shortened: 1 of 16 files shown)


api/src/repositories/sql-fragments.ts
```diff
@@ -9,105 +9,108 @@
...
+ * operate on un-indexed or possibly-inactive ids (e.g. raw id arrays) must filter to
+ * active features first, since the closure has no row for soft-deleted features.
+ *
+ * Returns an `EXISTS (...)` SQL expression (returns boolean) with zero `?` placeholders.
  *
  * @param featureIdExpr SQL expression for the starting submission_feature_id
- *   (e.g. 'wf.submission_feature_id', 'candidate.submission_feature_id')
+ *   (e.g. 'wf.submission_feature_id', 'expression_results.submission_feature_id')
  */
 export function isEffectivelySecured(featureIdExpr: string): string {
```
mauberti-bc, line 40: Possible security gap:

I believe the new security check asks: “Does this feature reach itself or a secured ancestor through `submission_feature_closure`?”

- If yes, the `submission_feature` is treated as secured.
- If no, the `submission_feature` is treated as unsecured.

That relies on every active searchable feature having at least its closure self-loop. If a closure recompute has not run yet or failed, an active secured feature may have no closure rows. In that case, the query returns “unsecured” by default.

We should instead default to the record being secured.

mauberti-bc (changes requested):
Nice! Search is very fast. Consider adding seed data to the `submission_feature_closure` table in this branch since they are related. I added the below seed for testing.

```
const computeSubmissionFeatureClosureForUpload = async (
  knex: Knex,
  submission_upload_id: string
): Promise<void> => {
  await knex.raw(
    `
      DELETE FROM submission_feature_closure c
      USING submission_feature sf
      WHERE sf.submission_upload_id = ?::uuid
        AND c.source_submission_feature_id = sf.submission_feature_id;
    `,
    [submission_upload_id]
  );

  await knex.raw(
    `
      WITH RECURSIVE
      active_features AS (
        SELECT submission_feature_id
        FROM submission_feature
        WHERE submission_upload_id = ?::uuid
          AND record_end_date IS NULL
      ),
      self_loops AS (
        SELECT submission_feature_id AS source, submission_feature_id AS target
        FROM active_features
      ),
      parent_edges AS (
        SELECT child.submission_feature_id AS source, child.parent_submission_feature_id AS target
        FROM submission_feature child
        JOIN active_features af_src ON af_src.submission_feature_id = child.submission_feature_id
        JOIN active_features af_tgt ON af_tgt.submission_feature_id = child.parent_submission_feature_id
        WHERE child.parent_submission_feature_id IS NOT NULL
      ),
      property_edges AS (
        SELECT pf.submission_feature_id AS source, pf.referenced_submission_feature_id AS target
        FROM submission_feature_property_feature pf
        JOIN active_features af_src ON af_src.submission_feature_id = pf.submission_feature_id
        JOIN active_features af_tgt ON af_tgt.submission_feature_id = pf.referenced_submission_feature_id
      ),
      ancestry_edges AS (
        SELECT source, target FROM self_loops
        UNION ALL SELECT source, target FROM parent_edges
      ),
      direct_edges AS (
        SELECT source, target FROM self_loops
        UNION ALL SELECT source, target FROM parent_edges
        UNION ALL SELECT source, target FROM property_edges
      ),
      ancestry_closure AS (
        SELECT source, target FROM ancestry_edges
        UNION
        SELECT c.source, d.target
        FROM ancestry_closure c
        JOIN ancestry_edges d ON d.source = c.target
      ),
      closure AS (
        SELECT source, target FROM direct_edges
        UNION
        SELECT c.source, d.target
        FROM closure c
        JOIN direct_edges d ON d.source = c.target
      )
      INSERT INTO submission_feature_closure (source_submission_feature_id, target_submission_feature_id, is_ancestor)
      SELECT cl.source, cl.target, (anc.source IS NOT NULL) AS is_ancestor
      FROM closure cl
      LEFT JOIN ancestry_closure anc
        ON anc.source = cl.source AND anc.target = cl.target;
    `,
    [submission_upload_id]
  );
};
```

api/src/repositories/sql-fragments.ts
dwhogan: good catch, thanks

mauberti-bc (approved):
👍

PR #471, opened 2026-06-08
SIMSBIOHUB-1040: Bug fix to prevent duplicate search request

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1040

## Description of Changes

- Deduplicates search-expression update handling so only one endpoint request is triggered per change.
- Prevents redundant loading-state transitions caused by overlapping duplicate requests.
- Stabilizes skeleton loader rendering so it does not flicker during search expression updates.

## Testing Notes

1) Submit a search expression 
2) Change tabs
3) Update search expression
4) Confirm the skeleton loader does not flicker
5) Confirm only one request per expression change

### DIFF
--- a/app/src/features/search/result/hooks/useSearchResults.tsx
+++ b/app/src/features/search/result/hooks/useSearchResults.tsx
@@ -75,6 +75,7 @@ export const useSearchResults = (
   const { searchParams, setSearchParams: setRawSearchParams } = useSearchQueryParams();
   const [data, setData] = useState<SearchFeatureResponse>();
   const [isLoading, setIsLoading] = useState(false);
+  const [hasSettled, setHasSettled] = useState(false);
   const searchApiRef = useRef(api.search);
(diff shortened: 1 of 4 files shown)


app/src/features/search/result/hooks/useSearchResults.tsx
```diff
@@ -243,7 +267,7 @@ export const useSearchResults = (
 
   return {
     rows: data?.features ?? [],
-    isLoading,
+    isLoading: isLoading || data === undefined,
```
dwhogan: Consider: the initial api load could returns 'undefined', handle this super edge case  with some like like

    isLoading: isLoading || !hasSettled

Set hasSettled on first api success

dwhogan (dismissed):
👍

added a none blocking edge case consideration

dwhogan (approved):
👍

PR #472, opened 2026-06-08
SIMSBIOHUB-1028: Reuse Export Packages for Downloads

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1028](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1028)

## Description

**Reuse Export Packages for Downloads**

As a user, I want identical exports for the same download version to reuse generated artifacts, so that BioHub does not regenerate the same export files for every user request.

### Background

The current `download_export` flow creates a new export row and queues a new CSV/ZIP generation job for each request.

Generated object keys include `exportId`:

```
downloads/{downloadId}/exports/{exportId}/biohub-{exportId}-part-{N}.zip
```

Because the physical artifacts are tied to the user-facing export row, identical exports cannot share files.

### Goal

Separate:

- **download** — logical user/team download request
- **download_version** — materialized snapshot of a download at a point in time
- **download_version_export** — user-facing export request for a specific download version
- **download_version_export_artifact_group** — reusable physical export artifact set
- **download_version_export_artifact** — generated artifact parts for the reusable group

### Proposed Data Model

**1. Download Version**

```
download_version
  download_version_id
  download_id
  status
  started_at
  completed_at
  error_message
  materialized_at
```

~~`download.current_download_version_id` should point to the current version where needed.~~

> **As-built:** there is **no** `current_download_version_id` pointer. Per the PR #472 review, `download` is the stable parent (policy, ownership, teams, metadata) and reads resolve a single `download_version` — defaulting to the **most recent** (a specific version may be targeted when provided). Reruns are additive: a new snapshot is a new `download_version`, never a mutation of the parent. See AC #1 and AC #14–#19.

> **As-built:** `created_from_policy_id` was dropped — policy is a *download* invariant (a new policy is a new download; `download.policy_id` is UNIQUE), so it can never diverge per-version. In its place, `materialized_at` is the data watermark, set when the version reaches `ready`. `download` itself drops `download_status`/`started_at`/`completed_at` — the materialization lifecycle moved onto the version, and read paths alias `download_version.status AS download_status`, so the API surface and the frontend are unchanged (only the column location moved).

**2. Download Version Artifacts**

```
download_version_artifact
  download_version_artifact_id
  download_version_id
  artifact_id
  feature_type_name nullable
  record_end_date
```

**3. User-Facing Export Request**

```
download_version_export
  download_version_export_id
  download_version_id
  format
  mode
  max_part_size_bytes
  download_version_export_artifact_group_id nullable
  create_user
  create_date
  record_end_date
```

> **As-built:** the group FK is named `download_version_export_artifact_group_id` (matching the referenced PK), not the shorter `export_artifact_group_id`. It is nullable only transiently within the create transaction; it is set on attach.

**4. Reusable Export Artifact Group**

```
download_version_export_artifact_group
  download_version_export_artifact_group_id
  download_version_id
  format
  mode
  max_part_size_bytes
  exporter_version (NOT NULL, default 1)
  status
  started_at
  completed_at
  error_message
  create_date
  record_end_date
```

> **As-built:** `exporter_version` is `NOT NULL DEFAULT 1` (not nullable) so it can be a non-null member of the active-uniqueness key — bump it on any packer-logic change to invalidate stale `ready` groups.

**5. Export Artifact Parts**

```
download_version_export_artifact
  download_version_export_artifact_id
  download_version_export_artifact_group_id
  artifact_id
  chunk_id
  record_end_date
```

### Uniqueness Rule

WHEN active artifact groups are created
THEN identical export options for the same download version should resolve to one active group

Suggested uniqueness constraint:

```
UNIQUE(download_version_id, format, mode, max_part_size_bytes, exporter_version)
WHERE record_end_date IS NULL
```

## Acceptance Criteria

**1. Add Download Versions**

WHEN the migration runs
THEN `download_version` should exist
AND a `download` should be able to have multiple versions
AND `download` should NOT carry a `current_download_version_id` pointer — the API resolves a version per request (see AC #14)

**2. Add Version Artifacts**

WHEN the migration runs
THEN `download_version_artifact` should exist
AND it should link a `download_version` to materialized artifacts

**3. Add User-Facing Export Records**

WHEN the migration runs
THEN `download_version_export` should exist
AND a download version should support multiple export records

**4. Add Reusable Export Artifact Groups**

WHEN the migration runs
THEN `download_version_export_artifact_group` should exist
AND groups should be unique by:
`download_version_id`, `format`, `mode`, `max_part_size_bytes`, `exporter_version`

**5. Add Export Artifact Parts**

WHEN the migration runs
THEN `download_version_export_artifact` should exist
AND artifact parts should link to `download_version_export_artifact_group`

**6. Reuse Ready Artifact Groups**

WHEN a user requests an export for a `download_version`
AND a matching artifact group already has `status = 'ready'`
THEN BioHub should create a new `download_version_export` row
AND attach it to the existing artifact group
AND return the export as ready without queuing a new job

**7. Reuse In-Progress Artifact Groups**

WHEN a user requests an export for a `download_version`
AND a matching artifact group has `status IN ('pending', 'processing')`
THEN BioHub should create a new `download_version_export` row
AND attach it to the existing artifact group
AND not queue another export job

**8. Create New Artifact Group When Needed**

WHEN a user requests an export
AND no matching active artifact group exists
THEN BioHub should create a new artifact group
AND create a new `download_version_export` row attached to that group
AND enqueue one background export job for the group

**9. Prevent Concurrent Duplicate Groups**

WHEN concurrent identical export requests are made
THEN only one active artifact group should be created
AND all matching export rows should attach to that group

**10. Update Export Worker**

WHEN the export worker runs
THEN it should process `download_version_export_artifact_group_id`
AND it should write artifact parts for the group
AND it should not write artifacts against an individual user-facing export row

**11. Use Group-Based Object Keys**

WHEN export artifacts are written
THEN object keys should use the reusable artifact group ID

Expected key shape:

```
downloads/{downloadId}/versions/{downloadVersionId}/exports/{artifactGroupId}/biohub-{artifactGroupId}-part-{N}.zip
```

**12. Resolve Export URLs Through Group**

WHEN export detail or download URL endpoints are called
THEN BioHub should resolve artifact parts through `download_version_export.download_version_export_artifact_group_id`
AND return URLs for the group artifacts linked to that export

**13. Replace Direct Export-to-Artifact Behaviour**

WHEN this model is implemented
THEN existing direct `download_export` to artifact behaviour should be replaced or migrated to group-based artifact behaviour

### Version-Centric Model (PR #472 review — as-built)

**14. Resolve a Single Version on Read (No Stored Pointer)**

WHEN a download is read (detail or list)
THEN BioHub should resolve a single `download_version`, defaulting to the most recent active version
AND a specific `download_version` may be targeted when provided
AND there should be no stored `download.current_download_version_id` pointer
AND `download` remains the stable parent (policy, ownership, teams, metadata) while `download_version` owns the processing state (status, timing, materialized snapshot, artifacts)

**15. Create and Enqueue a Version on Download Request**

WHEN a download request is created
THEN BioHub should create a `download_version`
AND enqueue that version for processing

**16. Process Queue Jobs by Version**

WHEN the parquet materialization job is enqueued and processed
THEN the job payload should carry `{ downloadVersionId }` (not `{ downloadId }`)
AND the job should resolve the parent `download` and its status from the version

**17. Require an Explicit Version on Export Creation**

WHEN a user requests an export
THEN the request should require an explicit `download_version_id`
AND BioHub should not default the export to a version
AND the download read responses should expose the resolved `download_version_id` so the client can supply it

**18. Version-Scope Parquet Object Keys**

WHEN raw parquet artifacts are written
THEN object keys should include `downloadVersionId`

Expected key shape:

```
downloads/{downloadId}/versions/{downloadVersionId}/{featureType}/data.parquet
```

AND re-running a download should write a new version's parquet without overwriting an older version's files

**19. Additive Reruns**

WHEN a download is re-run
THEN a new `download_version` (a new snapshot) should be created
AND the parent `download` should not be mutated

## Implementation Notes

- User-facing exports remain per request/user.
- Physical export artifacts are shared by artifact group.
- **As-built:** the user-facing `Content-Disposition` filename is `<downloadId>_<downloadVersionId>_<timestamp>_part<N>.zip` (download + version lineage, the group's `started_at` timestamp, and the 1-based part index — so two distinct export runs of the same download+version can't collide). The object key still uses the reusable artifact group ID.
- Failed artifact groups should be handled explicitly. A retry may either reuse and reset the failed group or create a new active group after ending the failed one, but it must not violate the active uniqueness constraint.

## Test Coverage

Add tests for:

- ready artifact group reuse
- pending/processing artifact group reuse
- concurrent identical export request race
- failed group retry behaviour
- single background job per unique export group
- part URL resolution through `download_version_export_artifact_group_id`
- migration or replacement of direct export-to-artifact behaviour

**As-built coverage** — every item is covered. Suites: `[unit]` `make test` · `[db]` `make test-db` · `[sys]` `make test-sys`.

| Ticket item | Covered by |
|---|---|
| ready group reuse | `[unit]` `download-export-service.test.ts` resolver matrix → "ready → attaches to the existing group, does not create, does not publish"; `[db]` `download-version-export.integration.ts` → "reuse of a ready group" |
| pending/processing reuse | `[unit]` resolver matrix → `[PENDING, PROCESSING]` "attaches to the in-flight group, does not create, does not publish"; `[db]` → "reuse of an in-flight group" |
| concurrent identical race | `[unit]` `download-version-export-repository.test.ts` ("`ON CONFLICT … DO NOTHING`", no throw on rowCount 0) + `download-export-service.test.ts` ("feeds the re-selected group id, not the probe"); `[db]` → "resolver conflict / re-select contract" + "migration cutover (schema)" partial-unique-index assertion. **Caveat:** single-connection *deterministic* collision — true two-connection simultaneity is guaranteed structurally (partial-unique index + `ON CONFLICT` + queue singleton), not driven by a concurrent test. |
| failed group retry | `[unit]` resolver matrix → "failed → ends the dead group before creating a fresh one, then publishes once"; `[db]` → "recovery from a failed group" |
| single job per group | `[unit]` `publisher.test.ts` → singletonKey `export-group-<id>`; resolver matrix "publishes once" (none/failed publish exactly once; ready/pending/processing never publish) |
| URL resolution through group | `[unit]` `download-export-service.test.ts` → `listExportPartUrls` + `getAuthorizedExportWithParts` (ready→`parts[]`, pending→`[]`, auth short-circuit); `[db]` → "part-URL resolution"; `[sys]` `download-export-pipeline-media.integration.ts` |
| replace direct export-to-artifact | `[db]` `download-version-export.integration.ts` → "migration cutover (schema)" (legacy `download_export`/`download_artifact`/`download_export_artifact` dropped; five new tables exist); `[sys]` `ingest-to-export.integration.ts` → cart → download → export round-trips to CSV against the migrated schema |

## Testing Notes

**Combined (DB drop-and-replace migration + reuse-engine queue worker + thin FE) — the reuse is invisible to the UI by design; the integration suites are the verification.**

```bash
make test       # unit, both layers (models, services, repos, paths, FE card + hook)
make test-db    # ← the one that matters: the reuse engine — resolve-or-create group, ON CONFLICT concurrency, failed-group end-and-recreate, group-sourced status
make test-sys   # full ingest → materialize → export → part-zip packaging through the queue + minio
```

`make test-db` is the verification — it proves Outcomes 1–4 (N identical export requests collapse to one group + one job) against real SQL; `make test-sys` proves the worker packages part-zips end-to-end (`ingest-to-export`, `download-export-pipeline-media`). The sidebar lists and downloads exports exactly as before, no manual app driving needed.


database/src/migrations/20260604130248_download_version_tables.ts
```diff
@@ -0,0 +1,485 @@
...
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    SET SEARCH_PATH = biohub, public;
+
+    --------------------------------------------------------------------------------
+    -- DROP LEGACY VERSION-LESS TABLES (reverse dependency order)
+    --------------------------------------------------------------------------------
+
+    -- Triggers drop automatically with their tables.
+    DROP TABLE IF EXISTS download_export_artifact;
```
mauberti-bc, line 23: Just acknowledging that this is ok because there are no production records in these tables. 👍

database/src/migrations/20260604130248_download_version_tables.ts
```diff
@@ -0,0 +1,485 @@
...
+    COMMENT ON COLUMN download_version.create_user IS 'The id of the user who created the record.';
+    COMMENT ON COLUMN download_version.update_date IS 'The datetime the record was last updated.';
+    COMMENT ON COLUMN download_version.update_user IS 'The id of the user who last updated the record.';
+    COMMENT ON COLUMN download_version.revision_count IS 'Revision count used for concurrency control.';
+
+    --------------------------------------------------------------------------------
+    -- DOWNLOAD.CURRENT_DOWNLOAD_VERSION_ID
+    --------------------------------------------------------------------------------
+
+    ALTER TABLE download ADD COLUMN current_download_version_id UUID;
```
mauberti-bc: I don't know that we want `download.current_download_version_id`. Instead, the query should resolve a single version, defaulting to the most recent. When requesting downloads, users can request a specific version. 

Summary of desired state:
- Treat `download` as the stable parent: policy, ownership, teams, metadata.
- Treat `download_version` as the actual processing unit: status, timing, materialized snapshot, artifacts.
- Creating a download request should create a `download_version` and enqueue that version.
- Queue jobs should process `{ downloadVersionId }`, not `{ downloadId }`.
- Queries should target a specific `download_version_id` when provided.
- Export creation should require an explicit `download_version_id`; exports should not default to a version.
- Parquet S3 keys should include `downloadVersionId` to avoid overwriting older versions.
- This makes reruns additive: new snapshot = new `download_version`, not mutation of the parent `download`.

mauberti-bc (changes requested):
Nice, very comprehensive!

1) The parquet files are still written per download, instead of per download_version:
`const s3Key = `downloads/${downloadId}/${featureTypeName}/data.parquet`;`. Should that be updated in this PR? 

We don't want to overwrite previous parquet files when writing a new version. Users can still access old versions if they know the ID.

2) Do we intentionally create the queue twice, in both `worker.ts` and `publisher.ts`: `boss.createQueue(JobQueues.PROCESS_DOWNLOAD_VERSION_EXPORT)`?

api/src/repositories/download/download-version-repository.ts
```diff
@@ -0,0 +1,244 @@
...
+   * @return {Promise<DownloadVersionRecord | null>}
+   * @memberof DownloadVersionRepository
+   */
+  async findDownloadVersionById(downloadVersionId: string): Promise<DownloadVersionRecord | null> {
+    const sql = SQL`
+      SELECT
+        download_version_id,
+        download_id
+      FROM download_version
+      WHERE download_version_id = ${downloadVersionId};
```
mauberti-bc: Consider adding `record_end_date IS NULL` to the query, since we typically treat soft-deleted records as if they were deleted?

mauberti-bc (changes requested):
Great work! A few other changes to fix types. Probably worth fixing in this PR:

- `useDownloadApi.interface.ts` - drop `feature_count` because the API doesn’t return it
- `useDownloadExportApi.interface.ts` - clean up old `download_export` wording and remove `downloaded` from export statuses
- `download-version-export.ts` - remove `downloaded` from the export status enum
- `download/index.ts` - remove `downloaded` from the export status enum in the list schema
- `download.ts` - update the `DownloadArtifactInfo` comment to reflect the new `download_version_artifact` table
- `download-status.ts` - add separate export-group status constants instead of sharing the download job constants
- `download-helpers.ts` - drop `feature_count` from the mock download data and keep mock export statuses to real export states

api/src/repositories/download/download-version-repository.ts
dwhogan: Good catch, thanks!

database/src/migrations/20260604130248_download_version_tables.ts
dwhogan: fixed

database/src/migrations/20260604130248_download_version_tables.ts
dwhogan: yup

mauberti-bc (approved):

PR #473, opened 2026-06-08
SIMSBIOHUB-1042: Show feature properties in search result table

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1042

## Description of Changes

- Added property schema metadata to search results.
- Rendered feature properties as table columns.
- Removed row actions from results table.
- Removed relevance, feature type, and description columns.
- Removed feature descriptions from search payload.
- Stabilized table columns across pagination.
- Improved cell ellipsis and lock-column spacing.
- Updated types and tests for the new payload.

## Testing Notes

- Search results table should dynamically render columns


api/src/openapi/schemas/search/search-feature.ts
```diff
@@ -72,14 +74,41 @@ export const featureSearchResultSchema: OpenAPIV3.SchemaObject = {
...
+    feature_type_property_id: { type: 'integer' },
+    feature_property_id: { type: 'integer' },
+    feature_property_type_id: { type: 'integer' },
+    name: { type: 'string' },
+    display_name: { type: 'string' },
+    description: { type: 'string', nullable: true },
+    type_name: { type: 'string' },
+    required_value: { type: 'boolean' },
+    calculated_value: { type: 'boolean' },
+    allow_multiple: { type: 'boolean' }
```
dwhogan: consider: is allow_multiple a forward add, it is not in use in the code

dwhogan (changes requested):
👍

found a sql fix when smoke testing the search

api/src/repositories/search-feature-repository.ts
```diff
@@ -101,7 +191,7 @@ export class SearchFeatureRepository extends BaseRepository {
         'sf.feature_type_id',
         'ft.name as feature_type_name',
         knex.raw(`sf.data->>'name' as feature_name`),
-        knex.raw(`sf.data->>'description' as feature_description`),
+        knex.raw(`COALESCE(sf.data->'properties', '{}'::jsonb) as properties`),
```
dwhogan: properties key not required knex.raw(COALESCE(sf.data, '{}'::jsonb) as properties)

app/src/utils/search-result-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Falsy values are displayed as an empty string, arrays are flattened into a comma-separated list, and object values
+ * use the shared safe JSON stringifier so structured values remain inspectable in grid cells.
+ *
+ * @param {JsonValue | undefined} value - Raw submission feature property value from the search result row.
+ * @returns {string} Display-ready property value.
+ */
+export const formatSubmissionPropertyValue = (value: JsonValue | undefined): string => {
+  // Treat absent, empty, and falsey property values as no display value in the compact grid.
+  if (!value) {
```
dwhogan: consider a nullish guard 

 if (value === null || value === undefined) {
      return '';
    }

api/src/openapi/schemas/search/search-feature.ts
mauberti-bc: It should be used now. Dylan's recent migration backfilled data, so properties that used to be `array` type are now a scalar (string, number, etc.) with `allow_multiple = true`.

Instead of properties like `{ focal_species: [{taxon_id: 1}, {taxon_id: 2}]}`, we use `{taxon_id: [1, 2]}`.

dwhogan: good to know, thanks

dwhogan (dismissed):
👍

api/src/repositories/search-feature-repository.ts
mauberti-bc: FIxed

app/src/utils/search-result-utils.ts
mauberti-bc: FIxed, thanks

dwhogan (approved):
👍

PR #474, opened 2026-06-09
SIMSBIOHUB-1027: Automatic Security Rule Screening

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1027](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1027)

## Description of Changes

- Adds automatic security screening as an **independent background workflow** (SonarQube-style), not a stage of the upload lifecycle. `indexed` remains the terminal-success status; if screening never runs the upload is unaffected and security can still be applied manually.
- New `submission_upload_security` table — one row per screening event (status, started/ended timestamps, errors, metadata, pg-boss `job_id`), mirroring the `submission_validation` pattern.
- New `submission-upload-security` pg-boss queue (+ DLQ, singleton per upload). The closure job enqueues it in the same transaction as the closure write, so screening never starts before `submission_feature_closure` is populated. The job is single-flight per upload, records a `started → completed` event row (DLQ records `failed`), evaluates active rules, and inserts idempotent draft `submission_feature_security` rows for trigger and closure-related features. The policy evaluator is a stub (separate ticket) — screening runs end-to-end but inserts no rows yet.
- `submission_feature_security.status` is `draft | active` (default `active`), plus a nullable `submission_upload_security_id` FK linking a row to the scan event that produced it (NULL = manual/legacy).
- **Workflow:** draft rows never act as applied security — all readers filter `status = 'active'` (except the rule-delete guard, which still counts drafts). Manually applying a rule promotes a conflicting draft to `active` and leaves `submission_upload_security_id` NULL.
- `submission_upload_job_status` is unchanged.

## Testing Notes

- API unit suite passes; `tsc` clean for api/app/database; ESLint clean.
- Upload a tarball and confirm the upload reaches `indexed` (terminal success, green check) — screening does not appear in the upload status.
- Confirm a `submission_upload_security` row is written by the job (`started → completed`); force a failure and confirm a `failed` event row is recorded **without** changing `submission_upload.status`.
- Manual-apply promotion: seed a `draft` row, apply that rule via the admin endpoint, confirm it flips to `active`.
- Validate migration up/down against a disposable database (creates/drops `submission_upload_security`, adds/drops the FK column, round-trips the `screened → active` rename).
- Note: with the evaluator stubbed, the scan-event lifecycle is the observable behavior — no draft rows are inserted yet.


api/src/models/submission-feature-security.ts
```diff
@@ -4,6 +4,7 @@ export const SubmissionFeatureSecurityRecord = z.object({
   submission_feature_security_id: z.number(),
   submission_feature_id: z.number(),
   security_rule_id: z.number(),
+  status: z.enum(['draft', 'screened']),
```
mauberti-bc: Status should be `draft` or `active`. In this context, draft just means that it was applied by the pgboss job. When a human applies a security rule, it will go straight to `active`.

api/src/constants/submission-upload.ts
```diff
@@ -8,7 +8,7 @@ import { SubmissionUpload } from '../models/submission-upload';
  * `failed` is intentionally excluded so users can restart jobs that failed for
  * transient external reasons.
  */
-export const TERMINAL_UPLOAD_STATUSES: SubmissionUpload['status'][] = ['indexed', 'invalid'];
+export const TERMINAL_UPLOAD_STATUSES: SubmissionUpload['status'][] = ['security_screened', 'invalid'];
```
mauberti-bc: The security screening should probably be independent of the `uploading` -> `ingesting` -> `indexing`. We shouldn't have a `security_screened` status here.

Revert the changes to `submission_upload.status` and remove the 'security_screening' and
  'security_screened'. 

Similar to SonarQube issues that run in the background of a PR, if the security screening does not run, it shouldn't affect the `submission_upload` because a human could just do it manually.

database/src/migrations/20260604000000_submission_feature_security_status.ts
```diff
@@ -0,0 +1,41 @@
...
+ *              default so that existing manually-applied rows and the manual-apply path
+ *              continue to behave as before without any data migration.
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    SET SEARCH_PATH = biohub, public;
+
+    CREATE TYPE submission_feature_security_status AS ENUM (
+      'draft',
+      'screened'
```
mauberti-bc: Functionally correct but rename to 'active' for clarity

database/src/migrations/20260604000001_submission_upload_status_screening.ts
```diff
@@ -0,0 +1,104 @@
...
+ * security_screened  — automatic screening completed; upload has reached terminal success.
+ *
+ * 'indexed' was previously the terminal-success state. After this migration it is an
+ * intermediate state; 'security_screened' becomes the new terminal success.
+ */
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`--sql
+    SET SEARCH_PATH = biohub, public;
+
+    CREATE TYPE submission_upload_job_status_v3 AS ENUM (
```
mauberti-bc: I think this migration file can be removed because we do not want to change the `submission_upload_job_status` as part of this PR.

mauberti-bc (changes requested):
- Keep `indexed` as the terminal upload success status.
  - Upload processing ends at indexing; security screening is a separate workflow.

- Do not replace `indexed` with `security_screened` in backend terminal status constants.
  - Revert `TERMINAL_UPLOAD_STATUSES` to include `indexed`.

- Revert frontend upload status changes that make `security_screened` the success state.
  - Keep `indexed` in terminal status lists, success colors, and success icons.

- Do not model screening as a linear `submission_upload.status`.
  - Avoid `indexed -> security_screening -> security_screened`.

- Add `submission_upload_security` table.
  - One row represents one security scanning event for a submission upload.

- Rename `automatic-security-screening-service/repository`.
  - Use `SubmissionUploadSecurityService` and `SubmissionUploadSecurityRepository`, named after the `submission_upload_security` table.
  - Our convention is for each service/repository to represent a database table, with methods primarily for modifying records in that table.

- Add nullable `submission_upload_security_id` to `submission_feature_security`.
  - Foreign key to the `submission_upload_security` record that yielded that rule 
  - `NULL` = manual or legacy security.
  - Non-null = produced by a specific scan event.

- Use `submission_upload_security` for scan lifecycle.
  - Track status, started/ended timestamps, errors, and metadata there.

- Rename `submission_feature_security.status` values.
  - Use `draft | active`, not `draft | screened`.

api/src/models/submission-feature-security.ts
dylanrogowsky-oxd: Done — status is now `draft | active` (default `active`). The pgboss job inserts `draft`; manual apply inserts/promotes straight to `active`.

api/src/constants/submission-upload.ts
dylanrogowsky-oxd: Done — reverted all `submission_upload.status` changes and removed both statuses. Screening is now an independent background workflow whose lifecycle lives in the new `submission_upload_security` table; `indexed` stays terminal, and security can still be applied manually if screening never runs.

database/src/migrations/20260604000000_submission_feature_security_status.ts
dylanrogowsky-oxd: Done — renamed `screened → active` across the enum, all readers, and manual-apply.

database/src/migrations/20260604000001_submission_upload_status_screening.ts
dylanrogowsky-oxd: Done — migration deleted; `submission_upload_job_status` is unchanged.

dylanrogowsky-oxd:
All points addressed (inline replies on the four threads above). Covering the remaining review items:

- **Add `submission_upload_security` table** — done; one row per scan event (status, started/ended timestamps, errors, metadata, pg-boss `job_id`), mirroring `submission_validation`.
- **Rename service/repository to the table** — done; `SubmissionUploadSecurityService` / `SubmissionUploadSecurityRepository` (+ queue/job/publisher). The repo owns only `submission_upload_security`; the `submission_feature_security` draft-insert moved into `SecurityRepository` per the one-repo-per-table convention.
- **Nullable `submission_upload_security_id` FK on `submission_feature_security`** — done; set by scans, NULL for manual/legacy.
- **Scan lifecycle in `submission_upload_security`** — done; job writes `started → completed` (DLQ → `failed`), with no `submission_upload.status` changes.

Note: the policy evaluator remains a stub (separate ticket), so screening runs end-to-end but inserts no draft rows yet.

mauberti-bc (changes requested):
Consider adding a `status = 'active'` filter to `getActiveAppliedFeatureCount` so draft screening rows do not block deleting an otherwise unused security rule.

dylanrogowsky-oxd:
> Consider adding a `status = 'active'` filter to `getActiveAppliedFeatureCount` so draft screening rows do not block deleting an otherwise unused security rule.

Updated the PR for this

mauberti-bc (approved):
👍

PR #476, opened 2026-06-12
SIMSBIOHUB-1056: Fix Expression Predicate Label Bug

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1056

## Description of Changes

- Preserves selected predicate property labels when shared autocomplete options refresh.
- Caches selected property metadata at builder and predicate-row level.
- Keeps selected properties available in predicate autocomplete options.
- Shows a skeleton fallback while selected property metadata is unresolved.
- Allows unresolved predicate rows to be deleted from the skeleton state.

## Testing Notes

1. Enter a search expression
2. Refresh the page
3. Confirm that the label correctly displays
4. Confirm initial skeleton loaders display


dwhogan (approved):
👍

PR #478, opened 2026-06-16
SIMSBIOHUB-1029: Update Submission Upload Status from ClamAV job

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1029](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1029)

## Description of Changes

Previously, when a malware scan job failed after exhausting all retries, the `submission_upload` was left stuck at status `uploaded`. The dead letter handler only marked `artifact_security` as `ERROR` and never propagated the failure to the owning upload.

- Added `ArtifactSecurityService.failSubmissionUploadByArtifactSecurityId`, which resolves the owning `submission_upload` via the artifact's `upload_archive` bridge and transitions it to `failed`. It is tolerant by design (runs in a terminal failure handler): no-op when the artifact has no `upload_archive`, and swallows the `ApiConflictError` raised when the upload is already in a terminal state.
- Wired this into `malwareScanFailedHandler` so a scan failure marks both `artifact_security` (`ERROR`) and `submission_upload` (`failed`) in the same transaction.

## Testing Notes

- Unit tests added for `failSubmissionUploadByArtifactSecurityId` (happy path, no-archive no-op, conflict swallowed, error propagation) and updated for `malwareScanFailedHandler` to assert the upload is marked `failed`.
- To verify manually: upload a submission with ClamAV unreachable so the scan job exhausts its retries. Confirm the job lands in the `MALWARE_SCAN_FAILED` dead letter queue and the `submission_upload` status moves from `uploaded` to `failed` (rather than staying stuck at `uploaded`).
- Confirm the happy path is unaffected: with ClamAV reachable, a clean scan still promotes the tarball to the main bucket and the upload progresses to `ingesting`.

### DIFF
--- a/api/src/queue/jobs/malware-scan-job.ts
+++ b/api/src/queue/jobs/malware-scan-job.ts
@@ -1,8 +1,8 @@
 import PgBoss from 'pg-boss';
-import { getAPIUserDBConnection } from '../../database/db';
 import { SecurityStatusEnum } from '../../models/security-status';
 import { ArtifactSecurityService } from '../../services/upload/artifact-security-service';
 import { getLogger } from '../../utils/logger';
+import { withConnection } from '../with-connection';
 
 const defaultLog = getLogger('queue/jobs/malware-scan-job');
 
@@ -64,16 +64,12 @@ export const malwareScanJobHandler: PgBoss.WorkHandler<IMalwareScanJobData> = as
       artifactSecurityId
     });
 
-    const connection = getAPIUserDBConnection();
-
     try {
-      await connection.open();
+      await withConnection(async (connection) => {
+        const artifactSecurityService = new ArtifactSecurityService(connection);
 
-      const artifactSecurityService = new ArtifactSecurityService(connection);
-
-      await artifactSecurityService.executeScan(artifactSecurityId);
-
-      await connection.commit();
+        await artifactSecurityService.executeScan(artifactSecurityId);
+      });
 
       defaultLog.info({
         label: 'malwareScanJobHandler',
@@ -93,8 +89,6 @@ export const malwareScanJobHandler: PgBoss.WorkHandler<IMalwareScanJobData> = as
       });
 
       throw error; // pg-boss will handle retry based on configuration
-    } finally {
-      connection.release();
     }
   }
 };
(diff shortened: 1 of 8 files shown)


api/src/queue/jobs/malware-scan-job.ts
```diff
@@ -124,6 +124,9 @@ export const malwareScanFailedHandler: PgBoss.WorkHandler<IMalwareScanJobData> =
         security: SecurityStatusEnum.ERROR
       });
 
+      // Propagate the scan failure to the owning submission upload so it does not stay stuck at `uploaded`.
+      await artifactSecurityService.failSubmissionUploadByArtifactSecurityId(artifactSecurityId);
```
mauberti-bc, line 118: Unrelated to your changes but worth fixing now because it blocks your code from working properly:

The `malwareScanJobHandler` methods are missing `connection.rollback`. Can you refactor to use `withConnection`, following the pattern from `indexSubmissionFeaturesFailedHandler`, which nicely handles the transaction lifecycle.

api/src/services/upload/artifact-security-service.ts
```diff
@@ -159,6 +159,51 @@ export class ArtifactSecurityService extends DBService {
...
+   * `upload_archive`, and swallows the {@link ApiConflictError} raised when the upload is already terminal.
+   *
+   * @param {string} artifactSecurityId - Artifact security record identifier.
+   * @returns {Promise<void>}
+   */
+  async failSubmissionUploadByArtifactSecurityId(artifactSecurityId: string): Promise<void> {
+    const securityRecord = await this.getArtifactSecurity(artifactSecurityId);
+
+    let uploadArchive: UploadArchive | null = null;
+    try {
```
mauberti-bc: We should not have try-catches within the service layer, unless calling a separate web service. The endpoint/entrypoint (eg. the pgboss job) for the flow owns the single try-catch . 

Example:
```
async failSubmissionUploadByArtifactSecurityId(artifactSecurityId: string): Promise<void> {
    const securityRecord = await this.getArtifactSecurity(artifactSecurityId);

    const uploadArchive = await this.uploadArchiveService.findUploadArchiveByArtifactId(securityRecord.artifact_id);

    if (!uploadArchive) {
      return;
    }

    const submissionUpload = await this.submissionUploadService.getSubmissionUploadByUploadId(uploadArchive.upload_id);

    if (TERMINAL_UPLOAD_STATUSES.includes(submissionUpload.status)) {
      return;
    }

    await this.submissionUploadService.transitionSubmissionUploadStatus(
      submissionUpload.submission_upload_id,
      'failed',
      ['uploaded', 'ingesting', 'ingested', 'indexing', 'failed']
    );
  }
```

mauberti-bc (changes requested):

api/src/queue/jobs/malware-scan-job.ts
dylanrogowsky-oxd: Updated `malware-scan-job.ts` to use `withConnection`.

mauberti-bc (approved):
👍

PR #480, opened 2026-06-16
SIMSBIOHUB-1044: Add blueprint database seeds

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1044](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1044)
- [SIMSBIOHUB-1043](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1043) - Dependency

## Description of Changes

- Adds a migration (`20260616130000_seed_initial_blueprint.ts`) that seeds the initial default Blueprint from the current `feature_type_property` configuration, so indexing behavior is preserved once property assignment is driven by the Blueprint tables.
  - Creates one initial Blueprint: `version_number = 1`, `is_default = true`, `record_effective_date = now()`, `record_end_date` null, `parent_blueprint_id` null (effective + not ended ⇒ available for indexing).
  - Seeds one `blueprint_feature_type` row per active `feature_type` (carrying its `sort`).
  - Seeds one `blueprint_feature_type_property` row per active `feature_type_property`, copying `feature_type_id`, `feature_property_id`, `required_value`, `allow_multiple`, and `sort`.
- A single chained statement performs all three inserts atomically so they share the one generated `blueprint_id`. The seed is idempotent (`WHERE NOT EXISTS` an active Blueprint), and `create_user` is left to the audit trigger, matching existing seed migrations.
- `required_value` is `COALESCE`d to `false` since the source column is nullable while the Blueprint column is `NOT NULL`; the copy is otherwise 1:1.

## Testing Notes

- Run the migrations against a populated database, then confirm the seed against the source tables:
  - One `blueprint` row with the flags above.
  - `blueprint_feature_type` count == active `feature_type` count.
  - `blueprint_feature_type_property` count == active `feature_type_property` count, with zero rows differing on `feature_type_id` / `feature_property_id` / `required_value` / `allow_multiple` / `sort`.
- Verified locally: 23/23 feature types and 107/107 property assignments seeded, 0 mismatches, `create_user` populated on all rows. `tsc` clean for the `database` package.
- `down()` removes the seeded assignments (children first) and the initial Blueprint.


database/src/migrations/20260616130000_seed_initial_blueprint.ts
```diff
@@ -0,0 +1,88 @@
...
+    seed_feature_types AS (
+      INSERT INTO blueprint_feature_type (blueprint_id, feature_type_id, sort)
+      SELECT nb.blueprint_id, ft.feature_type_id, ft.sort
+      FROM new_blueprint nb
+      CROSS JOIN feature_type ft
+      WHERE ft.record_end_date IS NULL
+      RETURNING 1
+    )
+    INSERT INTO blueprint_feature_type_property (
+      blueprint_id, feature_type_id, feature_property_id, required_value, allow_multiple, sort
```
mauberti-bc: After the changes to SIMSBIOHUB-1043, we need to insert the `blueprint_feature_type_id` rather than `feature_type_id`.

mauberti-bc (changes requested):

mauberti-bc (dismissed):
👍

dylanrogowsky-oxd:
Was approved prior to the merge/rebase. Test failing is the search timeout, which is a pre-existing intermittent issue.

PR #481, opened 2026-06-16
SIMSBIOHUB-1045: Apply blueprint during property indexing

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1045](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1045)

## Description of Changes

- Property indexing now resolves staged properties through the **selected Blueprint** instead of `feature_type_property`. A property is parsed only when the Blueprint includes its feature type (`blueprint_feature_type`) and assigns the property (`blueprint_feature_type_property`); `required_value` / `allow_multiple` come from the assignment. Primitive type lookup stays on `feature_property_type`.
- Each upload is **pinned to a Blueprint**: new `submission_upload.blueprint_id` (migration, NOT NULL + FK). On creation the id is resolved as: provided `blueprint_id` (validated `record_end_date IS NULL`, else 400) → most recent prior upload for the submission → active default. Optional `blueprint_id` is accepted on the upload request schemas and threaded through `_startArchiveUploadForSubmission`.
- Indexing uses `submission_upload.blueprint_id` as the source of truth — the default Blueprint is no longer selected inside the indexing SQL. Uploads are grandfathered in (no re-validation).
- Review fix: the required-property check constrains the pool-entry join with `ftp.feature_type_id = bft.feature_type_id` so a property assigned under another feature type can't satisfy requiredness.

## Testing Notes

- Re-submissions stay stable when the system default Blueprint changes (they reuse the prior upload's Blueprint).
- Behavior is unchanged for the seeded default Blueprint (1:1 with `feature_type_property`); the migration backfills existing uploads and the mock-data seed pins the default.
- Unit tests added (Blueprint resolver branches, repo lookups, indexing SQL); full suite + `tsc` + ESLint clean.

### DIFF
--- a/api/src/repositories/blueprint-repository.ts
+++ b/api/src/repositories/blueprint-repository.ts
@@ -5,7 +5,9 @@ export class BlueprintRepository extends BaseRepository {
   /**
    * Find a Blueprint's id only if it is currently available for new uploads.
    *
-   * A Blueprint is available when `record_end_date IS NULL` (not soft-deleted). This is the only
+   * A Blueprint is available when `record_end_date IS NULL` (not soft-deleted) and
+   * `record_effective_date <= now()` (live, not a draft or future-dated). A null
+   * `record_effective_date` represents a draft Blueprint that is not yet available. This is the only
    * availability check applied to a caller-provided `blueprint_id`; once stored on an upload the
    * Blueprint is grandfathered in and is not re-validated during indexing.
    *
@@ -20,7 +22,8 @@ export class BlueprintRepository extends BaseRepository {
         blueprint
       WHERE
         blueprint_id = ${blueprintId}
-        AND record_end_date IS NULL;
+        AND record_end_date IS NULL
+        AND record_effective_date <= now();
     `;
 
     const response = await this.connection.sql<{ blueprint_id: number }>(sqlStatement);
--- a/api/src/repositories/blueprint-repository.ts
+++ b/api/src/repositories/blueprint-repository.ts
@@ -1,0 +1,57 @@
+import { SQL } from 'sql-template-strings';
+import { BaseRepository } from './base-repository';
+
+export class BlueprintRepository extends BaseRepository {
+  /**
+   * Find a Blueprint's id only if it is currently available for new uploads.
+   *
+   * A Blueprint is available when `record_end_date IS NULL` (not soft-deleted). This is the only
+   * availability check applied to a caller-provided `blueprint_id`; once stored on an upload the
+   * Blueprint is grandfathered in and is not re-validated during indexing.
+   *
+   * @param {number} blueprintId - The requested Blueprint id.
+   * @returns {Promise<number | null>} - The `blueprint_id` if available, otherwise null.
+   */
+  async findActiveBlueprintById(blueprintId: number): Promise<number | null> {
+    const sqlStatement = SQL`
+      SELECT
+        blueprint_id
+      FROM
+        blueprint
+      WHERE
+        blueprint_id = ${blueprintId}
+        AND record_end_date IS NULL;
+    `;
+
+    const response = await this.connection.sql<{ blueprint_id: number }>(sqlStatement);
+
+    return response.rows[0]?.blueprint_id ?? null;
+  }
+
+  /**
+   * Find the id of the active default Blueprint (`is_default = true AND record_end_date IS NULL`).
+   *
+   * Used as the fallback for a brand-new submission that has no prior upload to inherit a Blueprint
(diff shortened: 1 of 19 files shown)


api/src/repositories/submission-feature-property-ingestion-repository.ts
```diff
@@ -980,14 +1015,25 @@ export class SubmissionFeaturePropertyIngestionRepository extends BaseRepository
...
+          ON bft.blueprint_id = sb.blueprint_id
+         AND bft.record_end_date IS NULL
+        -- Required properties assigned by the Blueprint.
+        JOIN blueprint_feature_type_property bftp
+          ON bftp.blueprint_feature_type_id = bft.blueprint_feature_type_id
+         AND bftp.record_end_date IS NULL
+         AND bftp.required_value = TRUE
+        -- Canonical pool entry; bridges the assignment to its feature type / property and surrogate id.
+        JOIN feature_type_property ftp
+          ON ftp.feature_type_property_id = bftp.feature_type_property_id
```
mauberti-bc, line 1029: Consider also checking that the assigned property belongs to the Blueprint feature type. The FK from `blueprint_feature_type_property.feature_type_property_id` only proves the property exists in the global pool; it does not prove it belongs to `bft.feature_type_id`.

```sql
JOIN feature_type_property ftp
  ON ftp.feature_type_property_id = bftp.feature_type_property_id
 AND ftp.feature_type_id = bft.feature_type_id
 AND ftp.record_end_date IS NULL
```

api/src/repositories/submission-feature-property-ingestion-repository.ts
```diff
@@ -961,15 +985,26 @@ export class SubmissionFeaturePropertyIngestionRepository extends BaseRepository
...
+   * is considered present only when staging has a non-null value and, for arrays, at least one
+   * element.
    *
    * @param {string} submissionUploadId Upload scope.
    * @returns {Promise<void>}
    */
   async recordMissingRequiredPropertyErrorsBySubmissionUploadId(submissionUploadId: string): Promise<void> {
     const sql = SQL`
-      WITH upload_feature_types AS (
+      WITH selected_blueprint AS (
```
mauberti-bc, line 1001: `selected_blueprint` will come from `submission_upload.blueprint_id` once available.

mauberti-bc (changes requested):
A few changes are needed to make Blueprint-based indexing backwards compatible.
The goal is that re-submitting (creating a new `submission_upload`) should not start failing because the system default Blueprint changed after the original upload.

- Add `submission_upload.blueprint_id`.

- On `submission_upload` creation:
  - Use the provided `blueprint_id` if one is supplied in the payload.
      - The `SubmissionUploadRequestSchema` should accept `blueprint_id` as an optional property, and passes it through to `this.submissionUploadService.insertSubmissionUpload` in `_startArchiveUploadForSubmission`
  - Otherwise default to the most recent prior `submission_upload.blueprint_id` for the same `submission_id`.
  - If there is no prior upload, default to the active Blueprint where `is_default = true`.

- During indexing:
  - Use `submission_upload.blueprint_id` as the source of truth.
  - Do not select the current default Blueprint inside the indexing SQL.

This keeps each upload tied to the Blueprint it was created with, so existing uploads remain stable when the default Blueprint changes later.

api/src/repositories/blueprint-repository.ts
```diff
@@ -0,0 +1,57 @@
...
+   * Find a Blueprint's id only if it is currently available for new uploads.
+   *
+   * A Blueprint is available when `record_end_date IS NULL` (not soft-deleted). This is the only
+   * availability check applied to a caller-provided `blueprint_id`; once stored on an upload the
+   * Blueprint is grandfathered in and is not re-validated during indexing.
+   *
+   * @param {number} blueprintId - The requested Blueprint id.
+   * @returns {Promise<number | null>} - The `blueprint_id` if available, otherwise null.
+   */
+  async findActiveBlueprintById(blueprintId: number): Promise<number | null> {
```
mauberti-bc, line 17: Add `AND record_effective_date <= now()` alongside the existing record_end_date IS NULL check so submitters can only pin uploads to currently available Blueprints. A null record_effective_date represents a draft Blueprint that is not yet available.

mauberti-bc (changes requested):

api/src/repositories/blueprint-repository.ts
dylanrogowsky-oxd: Updated the PR for this

mauberti-bc (approved):
👍

PR #482, opened 2026-06-17
SIMSBIOHUB-1030: Denormalized CSV export from parquet

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1030](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1030)

## Description of Changes

As a BioHub user, I want to create CSV files from a download package, so that I can use the exported feature data in spreadsheet tools.

```
download-package/
  tables/
    telemetry/
      part-00000.parquet
    deployment/
      part-00000.parquet
    animal/
      part-00000.parquet
```

This ticket covers CSV creation from those Parquet tables.

The CSV export flow should support two options:

- create one CSV per selected feature type
- create one denormalized CSV by merging selected feature types

For a denormalized CSV, the user should be able to choose:

- which feature types to merge
- which columns are used to merge rows
- which output columns to keep

Example:

```
telemetry.deployment_id left joins deployment.deployment_id
deployment.animal_id left joins animal.animal_id
```

This would allow a telemetry CSV to include selected deployment and animal fields.

### 1. Create CSVs from selected feature types

WHEN a user starts CSV export from a download package
THEN BioHub should show the feature types available in the package
AND the user should be able to select which feature types to export

WHEN the user exports selected feature types without merging
THEN BioHub should create one CSV per selected feature type

Example:

```
telemetry   -> telemetry.csv
deployment  -> deployment.csv
animal      -> animal.csv
```

### 2. Create one denormalized CSV

WHEN the user chooses to create a denormalized CSV
THEN BioHub should allow the user to select the feature types to merge
AND choose a root feature type that defines the row grain of the output CSV

Example:

```
Root feature type: telemetry
Output: one CSV row per telemetry row
```

### 3. Configure merge steps

WHEN multiple feature types are included in a denormalized CSV
THEN BioHub should allow the user to define merge steps

Each merge step should include:

- left feature type
- left column
- right feature type
- right column
- merge type

Initial merge type should be left join.

Example:

```
telemetry.deployment_id left joins deployment.deployment_id
deployment.animal_id left joins animal.animal_id
```

### 4. Select output columns

WHEN creating a CSV export
THEN BioHub should show available columns grouped by feature type
AND the user should be able to choose which columns to include in the output

For denormalized CSVs, selected columns may come from any feature type included in the merge.

### 5. Generate valid output columns

WHEN the CSV is generated
THEN the output should include only the selected columns
AND output column names should be unique

Example unique column names:

```
telemetry_recorded_at
deployment_device_number
animal_animal_name
```

### 6. Generate and store CSV files

WHEN the user submits the export
THEN BioHub should generate the requested CSV file or files
AND store them as export files associated with the download package
AND make them available for download in the UI

### 7. Preserve rows during left joins

WHEN a denormalized CSV uses a left join
THEN BioHub should preserve rows from the left side of the merge
AND leave right-side columns empty when no matching row is found

### 8. Validate export configuration

WHEN the user submits the CSV export
THEN BioHub should validate that selected feature types and columns exist in the package
AND validate that merge steps reference valid columns
AND prevent CSV generation when the configuration is invalid

## Implementation Notes

The CSV export should be generated from the selected Parquet feature tables in the download package.

The export configuration should be based on selected feature types, merge steps, and output columns.

The original Parquet files should remain unchanged.

## Testing Notes

**Combined FE + BE — new combined-CSV export dialog + denormalized join engine; includes a migration.**

```bash
make test      # unit (both layers — api + app)
make test-db   # drives the real Parquet→CSV join engine through MinIO (left-join keep, null-key NULL≠NULL, fan-out, chained merge, config-hash dedup, over-budget preflight)
make test-sys  # broader regression incl. ingest→export (slower; skip if test-db is green)
```

Manual smoke (combined-CSV export):

1. On a ready download's Export menu, "CSV — single combined file" opens a dialog listing the download's materialized feature types (defaulted to all); root = telemetry + a merge step (Start with telemetry → Join dataset on `parent_uuid`→`uuid`) + a column subset → submits, the export row goes Ready, and the downloaded zip is one CSV at one-row-per-telemetry grain with only the chosen `{type}_{column}` headers, right-side values populated.
2. "CSV — per feature type" opens no dialog and creates immediately (954 behaviour preserved as `mode: per_feature_type`); the zip holds one lossless CSV per materialized feature type.


database/src/migrations/20260610175136_download_export_config.ts
```diff
@@ -0,0 +1,45 @@
...
+ * The recipe is a self-contained description of how to read, join, and write an export. Hashing
+ * it canonically lets identical recipes against one version dedupe onto a single artifact group,
+ * and folds the old format/mode/exporter_version cache key into the hash itself.
+ */
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    SET SEARCH_PATH = biohub, public;
+
+    ALTER TABLE download_version_export_artifact_group
```
mauberti-bc, line 19: * This migration adds `config` and `config_hash` as `NOT NULL` columns without backfilling existing `download_version_export_artifact_group` rows. 
* That will fail on any DB that already has export groups. We should either add nullable columns, backfill old groups with the canonical per_feature_type config/hash (or some placeholder value), then set NOT NULL.

mauberti-bc (changes requested):
Note the sonarqube issues.

database/src/migrations/20260610175136_download_export_config.ts
dwhogan: Good catch, thanks.

mauberti-bc (changes requested):
Does it make sense to validate the merge steps?

Example:
```
if (config.root_feature_type) {
    try {
      const orderedSteps = orderMergeSteps(config.root_feature_type, config.merge_steps);

      if (config.mode === 'denormalized') {
        const reachableFeatureTypes = new Set<string>([config.root_feature_type]);

        for (const step of orderedSteps) {
          reachableFeatureTypes.add(step.right_feature_type);
        }

        for (const featureType of config.feature_types) {
          if (!reachableFeatureTypes.has(featureType)) {
            errors.push(
              `feature type '${featureType}' is not reachable from root_feature_type '${config.root_feature_type}' through merge_steps`
            );
          }
        }
      }
    } catch (error) {
      errors.push(error instanceof Error ? error.message : String(error));
    }
  }
```

We are validating that the requested merge_steps form an executable join plan:
- Start with the root table.
- Each merge step must join from a table already in the result.
- The step adds one new table to the result.
- Any requested output table must be either the root or added by one of those steps.
- The join columns must exist.

So this is valid if the columns exist, even if it ignores original parent/child relationships:
```
root: observation
step: observation.device_id = deployment.device_id
```

And this is invalid:
```
root: observation
include: deployment
steps: none
```
because `deployment` never gets joined into the result.

mauberti-bc (approved):
👍

PR #483, opened 2026-06-17
SIMSBIOHUB-1050: Download Rebuild Schedule

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1050](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1050)[SIMSBIOHUB-1050]

## Description of Changes

As a BioHub user, I want BioHub to automatically re-run a download job on a schedule, so that new `download_version` records can be created from the same download configuration over time.

### Background

A `download` represents the reusable download configuration.

A `download_version` represents one materialized result of running that download configuration at a point in time.

BioHub should support scheduled re-runs of a download. Each scheduled run creates a new `download_version`.

The schedule should be stored separately from `download_version`, because the schedule belongs to the parent `download`, not to an individual version.

### 1. Add `download_schedule`

WHEN the migration runs
THEN BioHub should add a `download_schedule` table
AND the table should store the schedule used to create future `download_version` records for a `download`.

```sql
CREATE TABLE download_schedule (
  download_schedule_id  integer         GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1),
  download_id           uuid            NOT NULL,
  cron_expression       text            NOT NULL,
  timezone              text            NOT NULL,
  next_run_date         timestamptz(6),
  last_run_date         timestamptz(6),
  record_end_date       timestamptz(6),
  create_date           timestamptz(6)  DEFAULT now() NOT NULL,
  create_user           integer         NOT NULL,
  update_date           timestamptz(6),
  update_user           integer,
  revision_count        integer         DEFAULT 0 NOT NULL,
  CONSTRAINT download_schedule_pk PRIMARY KEY (download_schedule_id),
  CONSTRAINT download_schedule_download_fk FOREIGN KEY (download_id) REFERENCES download(download_id)
);

CREATE INDEX download_schedule_download_id_idx ON download_schedule(download_id);
CREATE INDEX download_schedule_next_run_date_idx ON download_schedule(next_run_date);

-- A download can only have one active schedule
CREATE UNIQUE INDEX download_schedule_nuk1 ON download_schedule(download_id, (record_end_date is NULL)) WHERE record_end_date is null;
```

### 2. Store Cron-Based Schedule

WHEN a schedule is created for a `download`
THEN BioHub should store the schedule as a cron expression
AND store the timezone used to interpret the cron expression.

Example:

```
cron_expression = '0 2 * * *'
timezone = 'America/Vancouver'
```

### 3. Compute Next Run Date

WHEN a `download_schedule` is created or updated
THEN BioHub should compute and store `next_run_date`
AND `next_run_date` should be based on `cron_expression` and `timezone`.

### 4. Find Due Schedules

WHEN the scheduler job runs
THEN BioHub should find active `download_schedule` rows where `next_run_date <= now()`
AND `record_end_date IS NULL`.

### 5. Create New Download Version

WHEN a schedule is due
THEN BioHub should enqueue or run the existing download-version materialization workflow
AND create a new `download_version` for the schedule's `download_id`.

### 6. Update Schedule After Run

WHEN a scheduled run successfully creates a new `download_version`
THEN BioHub should update `last_run_date`
AND compute the next `next_run_date`.

### 7. Prevent Duplicate Runs

WHEN multiple workers check due schedules concurrently
THEN BioHub should prevent the same schedule occurrence from creating duplicate `download_version` records.

Implementation may use row locking, job deduplication, or a run table.

### 8. Preserve Existing Manual Runs

WHEN a user manually creates a new `download_version`
THEN the manual run should continue to work
AND it should not require a `download_schedule`.

### 9. Disable Schedule

WHEN a schedule is disabled or removed
THEN BioHub should set `record_end_date`
AND the schedule should no longer create new `download_version` records.

## Implementation Notes

The schedule belongs to `download`, not `download_version`.

Each scheduled execution creates a new `download_version`.

The existing download-version creation workflow should be reused. The scheduled job should not duplicate materialization logic.

A future ticket may add schedule management UI.

## Out of Scope

This ticket does not add the schedule management UI.

This ticket does not change the existing download-version materialization logic beyond allowing it to be triggered by a schedule.

This ticket does not implement notifications for completed scheduled downloads.

## Testing Notes

**Backend only — admin-gated schedule API (`POST`/`GET`/`DELETE /download/{id}/schedule`) + hourly pg-boss poll job; no UI (deferred).**

```bash
make test       # unit (both layers): service, route, repo, cron-util, poll-job, worker
make test-db    # ← the one that matters: partial-index upsert + full poll→rerun (SKIP LOCKED claim, mints one download_version)
make test-sys   # broader regression (slower; skip if test-db is green)
```

Manual sanity (admin-only API) — needs a `download_id`. Deployed envs have real ones; on a fresh local DB mint one (a bare download 404s — `findDownloadById` inner-joins the latest version, so it needs a version too):

```bash
# 1. mint a download_id (local only — there is no download seed data yet)
docker exec platform-db-all-container psql -U postgres -d biohubbc -tAc \
  "SET search_path TO biohub; WITH p AS (INSERT INTO policy(name) VALUES('manual test') RETURNING policy_id), d AS (INSERT INTO download(format,policy_id) SELECT 'csv',policy_id FROM p RETURNING download_id) INSERT INTO download_version(download_id,status) SELECT download_id,'ready' FROM d RETURNING download_id;"
# 2. schedule it (admin token)
curl -X POST http://localhost:6100/api/download/<downloadId>/schedule \
  -H "Authorization: Bearer <idir-admin-token>" -H 'Content-Type: application/json' \
  -d '{"cron_expression":"0 2 * * *","timezone":"America/Vancouver"}'
```


mauberti-bc (changes requested):
Nice! 

Consider adding an idempotency key that does not depend on download_version_id. 
* The scheduled job creates the download_version record, so a key based on download_version_id is too late to prevent duplicate version creation for the same scheduled run. 
* Derive the key from the scheduled run instead, e.g. `download_schedule:${download_schedule_id}:${next_run_date}` using the due next_run_date before it is advanced.

mauberti-bc (approved):
👍

PR #484, opened 2026-06-18
SIMSBIOHUB-1051: Curated Download Gallery

dwhogan:
## Links to Jira Tickets

[SIMSBIOHUB-1051](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1051)

## Description of Changes

Backend-only support for **galleries** — curated, reusable collections of downloads for display in curated areas (first use case: the home screen; the model is not home-screen specific).

Adding a download to a gallery only surfaces it on the board — it does **not** grant access to open or export the download. Download access remains governed by `download_team`.

## Data Model (additive)

- **`gallery`** — `name`, `slug` (short, stable, URL-safe key; **unique** among active galleries), `visibility` (`public` | `private`, default `public`), `description`; soft-delete via `record_end_date`.
- **`gallery_download`** — join table with `sort`; idempotent active-uniqueness on `(gallery_id, download_id)`; soft-delete.

Consumers reference a gallery by its stable `slug` (not its mutable `name`).

## API (8 endpoints)

- **Admin (SYSTEM_ADMIN):** create / list / update / delete gallery; add / remove gallery download.
- **Public (anonymous):** get gallery; get gallery downloads (each with its `exports[]`). Both scoped to `visibility = public` — a private gallery returns **404**.

## Testing Notes

```bash
make test      # unit: models, repositories, services, route handlers
make test-db   # gallery SQL: active-only filtering, slug uniqueness, visibility scoping,
               #   ON CONFLICT dedupe, sort ordering, getGalleryDownloads version-resolution join
```


api/src/services/gallery/gallery-service.ts
```diff
@@ -0,0 +1,213 @@
...
+   * @param {number | null} [sort] - Optional display order; an absent value sorts last.
+   * @return {Promise<void>}
+   * @throws {ApiNotFoundError} when no active gallery matches the given ID.
+   * @throws {HTTP400} when no download matches the given download ID.
+   * @memberof GalleryService
+   */
+  async addDownloadToGallery(galleryId: number, downloadId: string, sort?: number | null): Promise<void> {
+    await this.galleryRepository.getGalleryById(galleryId);
+
+    const download = await this.downloadRepository.findDownloadById(downloadId);
```
mauberti-bc: Suggestion: Switch to `getDownloadById`, which throws internally if the download is not found, so you can remove the `throw new HTTP400('Download not found');` from the service layer

api/src/services/gallery/gallery-service.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * Remove a download from a gallery.
+   *
+   * @param {number} galleryId - The gallery ID.
+   * @param {string} downloadId - The download ID to remove.
+   * @return {Promise<void>}
+   * @memberof GalleryService
+   */
+  async removeDownloadFromGallery(galleryId: number, downloadId: string): Promise<void> {
```
mauberti-bc: Suggestion: Move the methods affecting gallery-download joins into new `GalleryDownloadService` and `GalleryDownloadRepository`, so that the `GalleryService`/repository can focus on CRUD for the gallery table.

database/src/migrations/20260616120000_gallery.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.raw(`
+    SET SEARCH_PATH = biohub, public;
+
+    --------------------------------------------------------------------------------
+    -- GALLERY
+    --------------------------------------------------------------------------------
+
+    CREATE TABLE gallery (
```
mauberti-bc, line 11: Consider adding `slug` as a stable key we can use for the frontend URL. 
* Name might be long and grammatically correct (eg. BC Parks Biodiversity Monitoring Surveys)
* Slug is short (eg. /bc-parks-monitoring)
* Unique key on slug, just like for name

database/src/migrations/20260616120000_gallery.ts
```diff
@@ -0,0 +1,109 @@
...
+      gallery_id      INTEGER        GENERATED ALWAYS AS IDENTITY NOT NULL,
+      name            TEXT           NOT NULL,
+      description     TEXT,
```
mauberti-bc: Regarding access controls:
 
- Add `visibility` to `gallery`
  - Enum values: `public`, `private`
  - Default: `public`
- Include `visibility` in gallery API models/responses (all relevant interfaces)
- Allow `POST /gallery` to specify `visibility`
- Enforce visibility on public gallery reads
  - Public endpoints should only return `visibility = public`
  - Private galleries can remain admin-manageable for now
- Keep download access unchanged
  - Adding a download to a gallery only surfaces it on the board
  - It must not grant access to open/export the download
  - Download access remains controlled by `download_team`

We'll probably eventually need `gallery_team`, but we can leave galleries as either public or admin-only for now.

api/src/models/gallery.ts
```diff
@@ -0,0 +1,67 @@
...
+ * HTTP request body for adding a download to a gallery.
+ *
+ * `.strict()` rejects unknown keys so a frontend decoder bug fails fast at the
+ * boundary rather than leaking stray fields into the gallery membership.
+ *
+ * `downloadId` is the camelCase request field for the download's uuid. `sort` is
+ * nullable: NULL sorts last in the gallery ordering, so a member added without
+ * an explicit position lands at the end.
+ */
+export const AddGalleryDownloadRequestBody = z
```
mauberti-bc: Consider moving the gallery-download join-reated types/zod to `models/gallery-download.ts` to keep files small.

api/src/repositories/gallery/gallery-repository.ts
```diff
@@ -0,0 +1,305 @@
...
+   *
+   * @param {string} name - The gallery name to look up.
+   * @return {Promise<GalleryRecord | null>}
+   * @memberof GalleryRepository
+   */
+  async findActiveGalleryByName(name: string): Promise<GalleryRecord | null> {
+    const sql = SQL`
+      SELECT gallery_id, name, description, create_date
+      FROM gallery
+      WHERE name = ${name} AND record_end_date IS NULL;
```
mauberti-bc: Replace the lookup by `name` with lookup by `slug` (see comment on the migration file), to make name less significant (we might rename a gallery and don't want to break consumers).

mauberti-bc (changes requested):
Nice!

api/src/services/gallery/gallery-service.ts
dylanrogowsky-oxd: Done — `addDownloadToGallery` now calls `getDownloadById` (throws 404 internally); removed the manual `HTTP400`.

api/src/services/gallery/gallery-service.ts
dylanrogowsky-oxd: Done — added `GalleryDownloadService` + `GalleryDownloadRepository` for the join; `GalleryService`/`GalleryRepository` now own only gallery-table CRUD.

database/src/migrations/20260616120000_gallery.ts
dylanrogowsky-oxd: Done — added `slug` (required, kebab-case). It's now the unique key: `gallery_nuk1` moved off `name` → `slug`, so a gallery can be renamed without breaking consumers.

database/src/migrations/20260616120000_gallery.ts
dylanrogowsky-oxd: Done — added `visibility` (`public`/`private`, default `public`): in all models/responses, settable on POST/PUT, and enforced on both public reads (private → 404; admin endpoints still see all). Download access unchanged — still `download_team`. (`gallery_team` left for later.)

api/src/models/gallery.ts
dylanrogowsky-oxd: Done — moved `AddGalleryDownloadRequestBody` to `models/gallery-download.ts` (and the join OpenAPI schemas to `openapi/schemas/gallery-download.ts`).

api/src/repositories/gallery/gallery-repository.ts
dylanrogowsky-oxd: Done — `findActiveGalleryByName` → `findActiveGalleryBySlug`; create/update 409 pre-check now keys on `slug`. `name`'s unique constraint dropped — `slug` is the sole unique key.

mauberti-bc (approved):
👍

PR #485, opened 2026-06-23
SIMSBIOHUB-1065: Add `policy_expression` and remove dead `policy_statement_condition`

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1065

## Description of Changes

- Added `policy_expression` as the policy-owned expression identity table.
- Refactored `policy_statement_expression` to reference `policy_expression_id`.
- Enforces one `policy_statement_expression` per `policy_statement` (optional)
- Backfills `policy_expression` by merging expressions for statements that have multiple expressions
- Preserved expression reuse with service-level lookup before insert.
- Removed legacy policy statement condition tables and backend/frontend condition code.
- Kept runtime policy evaluation based on statement URNs plus expression trees.
- Added nullable `name` and `description` metadata to `policy_expression`.
- Updated API schemas, models, services, repositories, tests, and policy admin transforms accordingly.
- Removed old seed data that inserted policy statement conditions.

### Backfill Approach

- Create `policy_expression` as a policy-owned wrapper for existing `expression` rows.
- Add nullable `policy_statement_expression.policy_expression_id`.
- Stop if active legacy condition rows still exist, since dropping them would lose semantics.
- Soft-end stale active statement-expression links whose policy, statement, or expression is inactive.
- Create one active `policy_expression` per remaining active `(policy_id, expression_id)`.
- Backfill active `policy_statement_expression` rows to their matching `policy_expression`.
- Require active statement-expression rows to have `policy_expression_id`.
- For statements with multiple active expressions, synthesize/reuse one `AND` expression and wrap it in `policy_expression`.
- Repoint one active statement-expression row to the synthesized `AND`, then soft-end extra active rows.
- Enforce one active expression link per policy statement.
- Ensure active links cannot cross policies.
- Drop legacy condition tables after the guard passes.

## Testing Notes

- Submit a search with an expression filter and confirm the result set is correct.
- Create a download from that search and confirm export generation completes.
- Download the generated artifact and verify it contains only the expected feature set.
- Create a ticket from the same scoped context.
- Create a data request from the ticket using the selected feature scope/expression.
- Confirm the generated policy is created with statements linked through `policy_expression`.
- Confirm the policy grants access to the expected features and excludes out-of-scope features.
- Confirm team access from the generated policy works for the requested users.


api/src/services/access-policy/policy-expression-service.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (existing) {
+      return existing;
+    }
+
+    return this.policyExpressionRepository.insertPolicyExpression({
+      policy_id: policyId,
+      expression_id: expressionId,
+      name: options?.name ?? null,
+      description: options?.description ?? null
+    });
```
dylanrogowsky-oxd: Non-atomic read-then-write - Possible, but unlikely race condition?

dylanrogowsky-oxd (dismissed):

api/src/services/access-policy/policy-expression-service.ts
mauberti-bc: Fixed using advisory lock, thanks

PR #487, opened 2026-06-24
SIMSBIOHUB-1058: Fix search results pagination

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1058

## Description of Changes

- Fixes search result scrolling by adding minHeight: 0 to flex containers.
- Removes internal DataGrid pagination config from search result tables.
- Reverts to page 1 when limit changes

## Testing Notes

- Verify that changing the page or rows-per-page value in the search results footer updates the URL query params and sends the expected pagination values in the search request body.

### DIFF
--- a/app/src/features/search/result/SearchResultPage.test.tsx
+++ b/app/src/features/search/result/SearchResultPage.test.tsx
@@ -78,6 +78,7 @@ const mockCreateDataRequest = vi.fn();
 const mockGetAvailableUsers = vi.fn();
 const mockSetSnackbar = vi.fn();
 const mockSetOkDialog = vi.fn();
+const mockSetResultSearchParams = vi.fn();
 
 const codesPayload = {
   feature_type_with_properties: [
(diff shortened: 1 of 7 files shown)


dylanrogowsky-oxd:
The description mentions:

> Changes rows-per-page handling to update only limit, without forcing page=1﻿

If my page size is set to 10 items per page and there are 7000+ items.

1. Click to go to page 2
2. Change my page size to 25
3. It goes to page 1 and updates the query string params correctly

It forces page 1, which is appears to be inconsistent with the description. Is this expected?

mauberti-bc:
> The description mentions:
> 
> > Changes rows-per-page handling to update only limit, without forcing page=1
> 
> If my page size is set to 10 items per page and there are 7000+ items.
> 
>     1. Click to go to page 2
> 
>     2. Change my page size to 25
> 
>     3. It goes to page 1 and updates the query string params correctly
> 
> 
> It forces page 1, which is appears to be inconsistent with the description. Is this expected?

The comment was misleading but the code is correct: it should change to page 1 when limit changes. Thanks for noticing that - I changed the comment.

dylanrogowsky-oxd (approved):

PR #490, opened 2026-06-25
SIMSBIOHUB-1046: Update Property Staging and Indexed Property References

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1046](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1046)

## Description of Changes

- `feature_type_property_id` stays the primary reference on `submission_feature_property_*` (search and data access unchanged). This PR adds a **`NOT NULL`** `blueprint_feature_type_property_id` column recording which Blueprint assignment the indexing job used. Every indexed row provably has an assignment (the pipeline only indexes Blueprint-assigned properties, and assignments are immutable), so the column is never legitimately null.
- Migration adds the column, backfills existing rows via each feature's pinned Blueprint, then `SET NOT NULL` + FK + index.
- The indexing job populates the column on its 8 canonical inserts (resolved inline). Seeds and integration-test fixtures that insert property rows directly now also supply it inline (since `NOT NULL` rejects inserts that omit it).
- No reader/downstream changes — search, downloads, expression-evaluation, models, and per-type repos remain at dev.

## Testing Notes

- `npm run typecheck` / `npm run test` (2773 passing) / `eslint` clean; migration + seed lint/typecheck.
- Migration verified on a seeded dev DB: backfilled every existing row then `SET NOT NULL` succeeded; `down` drops cleanly. Full `seed:run` populates the column inline (100% non-null). DB integration suites pass (search, expression-evaluation, closure, feature-ingestion, download-parquet code/taxon/feature hydration); the one unrelated failure is an object-store bucket env not set in the local shell.
- Note: branch is behind `dev` and needs a dev merge before landing.


mauberti-bc (changes requested):
A few changes to make after looking at the search behaviour:
- The persisted `submission_feature_property_*` rows should continue to use `feature_type_property_id` as the main property reference.
- Blueprints should only be used during validation/resolution to decide whether a submitted property is allowed for the upload’s Blueprint.
- This branch currently renames/translates `feature_type_property_id` into `blueprint_feature_type_property_id` in the typed property tables, then makes search/download/expression evaluation join through `blueprint_feature_type_property` to recover the property.
  - That couples indexed data and search behavior to Blueprint assignment rows, which is not the intended model.
  - It also adds avoidable joins to search/download/expression queries.
- The desired model is to keep `feature_type_property_id` directly on `submission_feature_property_*` for search/data access performance.
- `blueprint_feature_type_property_id` can be added as an extra column to record which Blueprint assignment was used during validation.

**Suggested direction**:

- Keep `feature_type_property_id` on `submission_feature_property_*` as the field used by search and data access.
- Add `blueprint_feature_type_property_id` only as secondary validation/provenance metadata, so existing consumers can continue reading the property tables without changes.

dylanrogowsky-oxd:
@mauberti-bc - As discussed on the call, I have made a change so `blueprint_feature_type_property_id` is `NOT NULL`. 

Migration adds the column, backfills existing rows via each feature's pinned Blueprint, then `SET NOT NULL`

mauberti-bc (approved):
👍

PR #492, opened 2026-06-25
SIMSBIOHUB-1057: Conditionally Prompt Data Requests

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1057](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1057)
- [SIMSBIOHUB-1067](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1067)

## Description of Changes

Adds a `has_more_secured_features` signal to the feature search response, drives the "Request Data" banner from it, and prompts unauthenticated users to log in before requesting data.

**Backend**
- Feature search response now includes a top-level boolean `has_more_secured_features` (OpenAPI schema + service + handler), computed in parallel with the existing search/count/properties queries.
- New `SearchFeatureRepository.hasInaccessibleSecuredFeaturesByExpressionTree`: an `EXISTS`/`LIMIT 1` probe over the expression-matched candidate set **without** the caller access filter. True when a matched feature is effectively secured and not accessible to the caller — using `NOT isAccessibleToUser`, the same anchor-based definition as the visible-results filter, so the banner stays consistent with which rows are shown. Returns only the boolean; no hidden row data is exposed.
- New `buildUnfilteredExpressionTreeFeatureIdsSubquery` for the unfiltered candidate set.

**Frontend**
- "Request Data" banner is driven by `has_more_secured_features` (replaces deriving it from visible `is_secured` rows). Visible secured rows still render with `is_secured`.
- Unauthenticated "Request Access" redirects to login, preserving the search location (path + encoded `expr` query) as the return URL. The OIDC callback strips only the auth params (via the unit-tested `stripOidcParams` helper) so the search is restored after login.

**Seed**
- Mock seed (`04_mock_test_data.ts`) now builds `submission_feature_closure` per upload (shared `seed-utils.ts`), matching the indexing pipeline. Without closures, seeded features read as fail-closed-secured — producing false banners and hidden rows.

**Behavior note**
- The banner's accessibility check is anchor-based, identical to the visible-row filter. During the brief (~10s) anchor-recompute lag after a feature is secured, a caller with a covering policy grant may see a transient banner until anchors catch up. Accepted, in exchange for the banner always matching the rows that are shown.

## Testing Notes

- Requires `submission_feature_closure` to be built (the seed change handles this — **rebuild the `db_setup` image** so the seed fix is picked up). On a DB with unbuilt closures, every feature reads as secured.
- Verify: logged out / no-policy → banner shows when secured matches exist; `urn:*:*:*` user → no banner and secured rows visible with lock icons; delete the policy → banner returns.
- Tests: API unit (`search-feature-repository`, service, path, `expression-evaluation`) + DB integration (`security-scope-search.integration.ts`, incl. the anchor-lag transient-banner case). App: `SearchResultPage`, `useSearchResultDataRequest`, `stripOidcParams` (`Utils`), `useSearchApi`.
- Known local-only noise: two `findScopeIdsMatchingSubmission` integration tests can fail if a `urn:*:*:*` scope was created manually (count assertion); they pass on a clean DB.

### DIFF
--- a/api/src/paths/search/feature/{feature_type}/index.ts
+++ b/api/src/paths/search/feature/{feature_type}/index.ts
@@ -76,12 +76,8 @@ export function searchFeatures(): RequestHandler {
 
       const expressionTree = expressionTreeParseResult?.data;
 
-      const { features, properties, count } = await service.searchFeaturesByExpressionTreeWithCount(
-        featureType,
-        expressionTree,
-        pagination,
-        systemUserId
-      );
+      const { features, properties, count, has_more_secured_features } =
+        await service.searchFeaturesByExpressionTreeWithCount(featureType, expressionTree, pagination, systemUserId);
 
       await connection.commit();
 
@@ -90,7 +86,8 @@ export function searchFeatures(): RequestHandler {
       return res.status(200).json({
         features,
         properties,
-        pagination: makePaginationResponse(count, pagination)
+        pagination: makePaginationResponse(count, pagination),
+        has_more_secured_features
       });
     } catch (error) {
       defaultLog.error({ label: 'searchFeatures', message: 'error', error });
--- a/api/src/paths/search/feature/{feature_type}/index.ts
+++ b/api/src/paths/search/feature/{feature_type}/index.ts
@@ -81,8 +81,6 @@ export function searchFeatures(): RequestHandler {
 
       await connection.commit();
 
-      res.setHeader('Cache-Control', 'public, max-age=90');
-
       return res.status(200).json({
         features,
         properties,
--- a/api/src/repositories/search-feature-repository.ts
+++ b/api/src/repositories/search-feature-repository.ts
@@ -6,7 +6,12 @@ import { SearchFeatureResultWithRelevancy } from '../services/search-feature-ser
 import { ApiPaginationOptions } from '../zod-schema/pagination';
 import { BaseRepository } from './base-repository';
 import { dependencies as expressionEvaluation } from './expression-evaluation';
-import { buildSecurityFilter, isAccessibleToUser, isEffectivelySecured, isSubmissionFeatureActive } from './sql-fragments';
+import {
+  buildSecurityFilter,
+  isAccessibleToUser,
+  isEffectivelySecured,
+  isSubmissionFeatureActive
+} from './sql-fragments';
 
 /**
  * Repository for searching submission features by expression-tree criteria.
--- a/api/src/repositories/search-feature-repository.ts
+++ b/api/src/repositories/search-feature-repository.ts
@@ -6,7 +6,7 @@ import { SearchFeatureResultWithRelevancy } from '../services/search-feature-ser
 import { ApiPaginationOptions } from '../zod-schema/pagination';
 import { BaseRepository } from './base-repository';
 import { dependencies as expressionEvaluation } from './expression-evaluation';
-import { buildSecurityFilter, isEffectivelySecured, isSubmissionFeatureActive } from './sql-fragments';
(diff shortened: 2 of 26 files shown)


api/src/paths/search/feature/{feature_type}/index.ts
mauberti-bc: We should probably remove the `Cache-Control: public` header from this endpoint. The response is now user-dependent so caching can leave has_more_secured_records stale after access changes. That could confuse users/admins if it remains false even after a policy or permission update.

app/src/features/search/result/hooks/useSearchResultDataRequest.ts
```diff
@@ -34,11 +36,22 @@ export const useSearchResultDataRequest = ({ featureType, expressionTree }: UseS
...
+   * The data-request flow requires an authenticated BioHub user. When the caller
+   * is unauthenticated they are redirected to the Keycloak login page, preserving
+   * the current search location (path + query string, which includes the encoded
+   * `expr` search expression) as the return URL so the search results are
+   * restored after login.
    */
   const handleOpenCreateDataRequest = useCallback(() => {
+    if (!auth.isAuthenticated) {
+      const redirectUri = `${globalThis.location.origin}${globalThis.location.pathname}${globalThis.location.search}`;
+      auth.signinRedirect({ redirect_uri: redirectUri });
```
mauberti-bc: My suggestion to redirect with the `expr` might actually fail in prod: 
* *Note: wildcard* redirect URIs are not allowed in Prod for security reasons (https://sso-requests.apps.gold.devops.gov.bc.ca/)

Suggestion: Look into the `state` param and other options for preserving the search expression

api/src/repositories/search-feature-repository.ts
```diff
@@ -86,6 +86,63 @@ export class SearchFeatureRepository extends BaseRepository {
...
+   * - For anonymous users: true when any matched feature is effectively secured (none are accessible).
+   *
+   * No feature data is selected — only the boolean is returned, so no hidden secured rows are exposed.
+   *
+   * @param {string} anchorFeatureType - Target feature type returned by the search
+   * @param {NormalizedExpressionTreeExpression | undefined} expressionTree - Expression tree criteria
+   * @param {number | null} [systemUserId] - Security context (null = anonymous)
+   * @return {Promise<boolean>} True if matching secured features exist that the caller cannot access
+   */
+  async hasInaccessibleSecuredFeaturesByExpressionTree(
```
mauberti-bc, line 117: Very nice!

Ensure that `unfiltered` only means access/security filters are not applied. The query should still only consider valid features: `submission_feature.record_effective_date < now()` and `submission_feature.record_end_date IS NULL` (see PR #499).

mauberti-bc (changes requested):

api/src/paths/search/feature/{feature_type}/index.ts
dylanrogowsky-oxd: Good catch, removed it. The response is per-user now so caching would leave the banner stale after an access change.

app/src/features/search/result/hooks/useSearchResultDataRequest.ts
dylanrogowsky-oxd: Yeah, that would've broken in prod. Moved the search location into the `state` param instead of touching redirect_uri, and onSigninCallback restores it after login. Thanks for flagging.

api/src/repositories/search-feature-repository.ts
dylanrogowsky-oxd: It does, confirmed. isSubmissionFeatureActive (from #499) is applied unconditionally in both candidate paths, the unfiltered version only drops the access filter. Added a test asserting the candidate query still requires record_effective_date.

PR #496, opened 2026-06-26
SIMSBIOHUB-1075: Add 'survey' feature_type to replace 'dataset'

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1075

## Background

- We want to rename the `download` concept to `dataset`, since that is more accurate for users
- To do that, we must rename the `dataset` feature type to `survey`, to reduce confusion

## Description of Changes

- Requires https://github.com/bcgov/biohubbc/pull/1645
- Added `survey` as the new active feature type for future uploads.
- Soft-deleted the old `dataset` feature type without deleting historical data (kept for backwards compatibility for ingesting existing tarballs).
- Kept existing `dataset` records readable in search and expression queries.
- Avoided rewriting existing submissions, security scopes, policies, or feature history.
- Updated UI wording from `Survey Details` to `Submission Details` where the page is submission-level.

## Testing Notes

- Publishing a SIMS survey should succeed


dylanrogowsky-oxd (approved):
> Kept existing dataset records readable in search and expression queries.

Just wanted to highlight/confirm that legacy datasets are available from search but not from the Surveys tab, which aligns with the description.

PR #500, opened 2026-06-27
SIMSBIOHUB-1064: Allow access to route for secured features

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1064](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1064)

## Description of Changes

A user with access to a secured **parent** feature could not open the secured **child/descendant** features belonging to it — the feature-detail page returned `403` and redirected to `/forbidden`. This was inconsistent with search and cart, which already grant access to any secured descendant of an accessible ancestor.

The root cause was **two parallel authorization systems**. Search/cart/download use the closure-based `isAccessibleToUser` / `buildSecurityFilter` (`api/src/repositories/sql-fragments.ts`), which grants access when a feature is unsecured or the user's team has a scope grant via a `security_scope_anchor` on the feature **or any ancestor** (walked through `submission_feature_closure`). The feature-detail path instead used a narrow exact-feature URN check that only matched the requested feature, so an ancestor grant did not cover its descendants.

This PR aligns feature-detail authorization with the same closure-based semantics — fixed in backend authorization logic, not with frontend exceptions.

- **`team-authorization-repository.ts`** — Replaced the narrow `findTeamPolicyBySubmissionFeature` (exact-URN match) with `isSubmissionFeatureAccessibleToUser(systemUserId, submissionFeatureId, submissionId?)`. It runs a single `SELECT EXISTS(...)` that:
  - gates liveness — `record_effective_date <= now() AND record_end_date IS NULL` → soft-deleted / not-yet-effective features are forbidden;
  - reuses `buildSecurityFilter`, so unsecured features are open to everyone (incl. anonymous via `null`), and secured features are granted only when an ancestor in `submission_feature_closure` is a reachable `security_scope_anchor`;
  - keeps the optional `submission_id` path-integrity guard.
- **`team-authorization-service.ts`** — `isUserAuthorizedForTeamEntity` now accepts `number | null`; the `submission_feature` case delegates directly to the new repository method. `ticket` / `data_request` still require an authenticated user.
- **`authorization-service.ts`** — `authorizeByTeam` now lets anonymous requests reach the `submission_feature` check with a `null` system user id (so unsecured features are viewable by non-authenticated users), while other entities still short-circuit to `false` for anonymous.
- Removed the now-dead `findTeamPolicyBySubmissionFeature` query and `TeamPolicyRecord` model; added a small `SubmissionFeatureAccessRecord` zod model for response validation (consistent with the sibling repository methods).
- Both endpoints routing through the `Team` / `submission_feature` scheme — feature detail (`/submission/{submissionId}/features/{submissionFeatureId}`) and its `/properties` — are fixed consistently.

## Testing Notes

Unit tests added/updated for the repository, service, and authorization-service layers (full API suite green).

End-to-end, against a seeded DB — for a user granted a secured **parent** feature:

1. Open the parent feature detail → loads (200).
2. Open a secured **child** of that parent → loads (200), **not** redirected to `/forbidden`.
3. Open a secured feature under a **different** parent the user lacks → `403` → `/forbidden`.
4. As **anonymous**: open an unsecured feature → loads; open a secured feature → `403`.
5. Open a soft-deleted (`record_end_date` set) or not-yet-effective (`record_effective_date > now()`) feature → `403`.
6. Repeat 1–3 against the feature `/properties` endpoint to confirm consistency.

### DIFF
--- a/api/src/services/authorization/team-authorization-service.ts
+++ b/api/src/services/authorization/team-authorization-service.ts
@@ -24,16 +24,31 @@ export class TeamAuthorizationService extends DBService {
    * Entity semantics:
    * - `ticket`: membership in the ticket visibility team.
    * - `data_request`: membership in the data-request visibility team.
-   * - `submission_feature`: the feature is not effectively secured, or the user's team
-   *   holds a security scope anchored on the feature or one of its ancestors (the same
-   *   ancestry-aware check the search/download read paths apply).
+   * - `submission_feature`: ancestry-aware, closure-based access shared with the search/download
+   *   read paths — the feature is unsecured, or the user's team holds a security scope anchored
+   *   on the feature or one of its ancestors. `systemUserId` may be `null` for anonymous users
+   *   (who can still access unsecured features).
    *
-   * @param {number} systemUserId
+   * @param {number | null} systemUserId The authenticated user's id, or `null` for anonymous.
    * @param {TeamAuthorizationEntity} entity
    * @return {Promise<boolean>}
    * @memberof TeamAuthorizationService
    */
-  async isUserAuthorizedForTeamEntity(systemUserId: number, entity: TeamAuthorizationEntity): Promise<boolean> {
+  async isUserAuthorizedForTeamEntity(systemUserId: number | null, entity: TeamAuthorizationEntity): Promise<boolean> {
+    // submission_feature uses closure-based access and supports anonymous (unsecured) access.
+    if (entity.entity === 'submission_feature') {
+      return this.teamAuthorizationRepository.isSubmissionFeatureAccessibleToUser(
+        systemUserId,
+        entity.submissionFeatureId,
+        entity.submissionId
+      );
+    }
+
+    // ticket / data_request require an authenticated user.
+    if (systemUserId === null) {
+      return false;
+    }
+
     let record: { record_end_date: string | null } | null;
 
     switch (entity.entity) {
@@ -47,16 +62,6 @@ export class TeamAuthorizationService extends DBService {
           entity.dataRequestId
         );
         break;
-
-      case 'submission_feature':
-        // Ancestry-aware access check shared with the search/download read paths: unsecured
-        // features are open; secured features require the user's team to hold a scope anchored
-        // on the feature or one of its ancestors.
-        return this.teamAuthorizationRepository.isSubmissionFeatureAccessibleToUser(
-          systemUserId,
-          entity.submissionFeatureId,
-          entity.submissionId
-        );
 
       default:
         return false;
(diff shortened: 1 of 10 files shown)


api/src/services/authorization/team-authorization-service.ts
```diff
@@ -24,16 +24,31 @@ export class TeamAuthorizationService extends DBService {
...
+   *   (who can still access unsecured features).
    *
-   * @param {number} systemUserId
+   * @param {number | null} systemUserId The authenticated user's id, or `null` for anonymous.
    * @param {TeamAuthorizationEntity} entity
    * @return {Promise<boolean>}
    * @memberof TeamAuthorizationService
    */
-  async isUserAuthorizedForTeamEntity(systemUserId: number, entity: TeamAuthorizationEntity): Promise<boolean> {
+  async isUserAuthorizedForTeamEntity(systemUserId: number | null, entity: TeamAuthorizationEntity): Promise<boolean> {
```
mauberti-bc: It feels like a slippery slope to allow `systemUserId` to be nullable in `isUserAuthorizedForTeamEntity`, because feature access is no longer purely a team-membership check.

A submission feature is accessible when either:

* the feature is open; or
* the authenticated user is granted access through the `security_scope` concept and associated policy/team relationships.

Suggestion: keep `isUserAuthorizedForTeamEntity(systemUserId: number, ...)` limited to entities where authorization means “does this user belong to the relevant team?” Add a separate policy-based authorization rule, such as `authorizeByPolicy`, for submission features. That handler can evaluate public access first and, when the feature is secured, evaluate the user against the feature’s security policy.

Something akin to:
```
/**
 * Check whether the current request is authorized to access a submission feature.
 *
 * Access is granted when the feature is open, or when the authenticated user
 * has access through the feature's security policy.
 *
 * @param {AuthorizeByPolicy} authorizeRule
 * @returns {Promise<boolean>}
 */
async authorizeByPolicy(authorizeRule: AuthorizeByPolicy): Promise<boolean> {
  if (!authorizeRule) {
    return false;
  }

  const user = await this.getCachedSystemUser();

  const teamAuthorizationService = new TeamAuthorizationService(this.connection);

  return teamAuthorizationService.isSubmissionFeatureAccessibleToUser(
    user?.system_user_id ?? null,
    authorizeRule.submissionFeatureId,
    authorizeRule.submissionId
  );
}
```

Then we can revert the `systemUserId: number | null` types back to `systemUserId` in the modified method signatures, so methods downstream of `isUserAuthorizedForTeamEntity` are unchanged.

mauberti-bc (changes requested):

mauberti-bc (approved):
👍

PR #501, opened 2026-06-27
SIMSBIOHUB-1063: Grant Users Access to their Own Submissions

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1063](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1063)

## Description of Changes

As a SIMS submitter, I want access to the submissions I created so I can view my submitted data in BioHub after it is received.

Previously, the submission-upload flow never wrote a `submission_team` row, so a submitter could not see their own submission through `GET /submission` (which only returns submissions reachable via `team_member → team → submission_team → submission`). The ticket created during upload does mint a team containing the submitter, but that ticket team is intentionally **not** part of the `submission_team` access path. This PR grants the submitting user access to their submission through `submission_team`.

**Access grant**
- Added `submission_team` model, repository (`SubmissionTeamRepository.insertSubmissionTeam` — idempotent via `WHERE NOT EXISTS`, since the `submission_team_nuk1` unique index is partial and can't be targeted by `ON CONFLICT`), and `SubmissionTeamService.grantSubmissionAccessToUser`.
- Access is granted by linking the submission to a **per-user "owner" team** (find-or-create, one team reused per user), independent of the ticket team. Added `getTeamByName` to `TeamRepository`/`TeamService` to support the find-or-create.
- A transaction-scoped `pg_advisory_xact_lock` (keyed per user) serializes concurrent grants for the same user, preventing a duplicate-team / duplicate-link unique-constraint race. Follows the existing advisory-lock idiom (`policy-expression-repository`, `ticket-repository`).

**Submitter resolution + authentication**
- Both upload routes (`POST /submission/upload/archive` and `POST /submission/{submissionId}/upload`) now resolve the submitting user from the authenticated token via the existing `UserService.ensureSystemUser` (creates/reactivates the `system_user` as appropriate).
- If the token can't be resolved to a user, the request is rejected with `401` before any records are created (the resolution runs inside the request transaction, so a failure rolls back the `submission`, `submission_upload`, ticket, and `submission_team` records). Existing `Contributor` authorization is unchanged.

**Wiring**
- `UploadIngestionService` threads the resolved submitter id through `startArchiveUpload` / `startArchiveUploadForExistingSubmissionByUuid` / `_startArchiveUploadForSubmission` and performs the grant in the shared internal flow, so it covers **both** new submissions and uploads appended to existing submissions. For appends, the user performing the append is granted access (ticket-team behaviour is unchanged).

## Testing Notes

- New/updated unit tests: `submission-team-repository`, `submission-team-service` (find-or-create both branches + advisory lock), `team-repository.getTeamByName`, `upload-ingestion-service` (grant called for new and append flows; append grants the appender, not the original owner), and both upload route handlers (submitter resolved, submitter id forwarded, `401` + rollback when the token can't resolve). Full API suite green (`npm run test`), `npm run lint` and `npm run typecheck` clean.
- Manual end-to-end: authenticate as a submitter, `POST /submission/upload/archive`, then `GET /submission` and confirm the new submission appears. Repeat `POST /submission/{submissionId}/upload` (same or different user) and confirm that user now sees the submission.
- Negative path: send a request whose token has no resolvable user guid → expect `401` and verify no `submission` / `submission_upload` / ticket / `submission_team` rows were created.


api/src/paths/submission/{submissionId}/upload/index.ts
```diff
@@ -63,18 +66,33 @@ POST.apiDoc = {
  */
 export function createSubmissionUpload(): RequestHandler {
   return async (req, res) => {
-    const connection = getDBConnection(req.keycloak_token);
+    const token = req.keycloak_token!;
```
mauberti-bc, line 67: This token will be for the service account, rather than the user who triggered the service account's request. I think the request body will need to include the token or the original user's information to create the user record.

PR #502, opened 2026-06-27
SIMSBIOHUB-1061: Fail Authorization for Soft-Deleted BioHub Users

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1061](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1061)

## Description of Changes

Soft-deleted (revoked) users could regain access by logging in again. This PR closes that gap and makes `system_user.user_guid` resolution case-insensitive everywhere.

- **Soft-deleted users fail authorization.** Added a single `isSystemUserInactive` predicate (`models/user.ts`) — a user with `record_end_date` set to now or in the past — used as the soft-delete gate in `AuthorizationService.getSystemUserObject`, the chokepoint every authorization rule resolves through. Inactive users now resolve to `null`, so **all** rules reject them uniformly. Previously only `authorizeBySystemRole` checked `record_end_date`, leaving soft-deleted system admins and users with historical team membership able to read secured data.
- **Case-insensitive login resolution.** `getUserByGuid` now matches `LOWER(user_guid)` and `addSystemUser` stores GUIDs lower-cased. Login (`PUT /user/self` → `upsertSelf`) resolves the existing soft-deleted row regardless of token casing and throws `HTTP401` using the same `isSystemUserInactive` predicate — no new/duplicate active row, no implicit reactivation.
- **Duplicate GUID prevention.** Replaced the case-sensitive unique index `system_user_nuk2` with a case-insensitive one on `LOWER(user_guid)` (new migration `20260627000000_user_guid_case_insensitive_unique.ts`), with a pre-flight guard that raises a clear error if existing rows collide when lower-cased.
- **Consistent session context.** Updated the `api_set_context` DB function (in the same migration) to resolve the session user by `LOWER(user_guid)`, so a non-lower-case token GUID still resolves the lower-cased `system_user` row.
- Reactivation stays explicit (admin `/user/add` flow only); login never implicitly recreates or reactivates a revoked user.

## Testing Notes

- Unit tests added/updated for the shared `isSystemUserInactive` predicate, the `getSystemUserObject` soft-delete gate, `upsertSelf` rejecting expired users, case-insensitive `getUserByGuid`, and lower-cased `addSystemUser`. Full API suite passes.
- Manual end-to-end (running stack): soft-delete a user (set `record_end_date`), then confirm:
  - secured routes return `403`;
  - `PUT /user/self` with the same GUID in different casing returns `401`, creates no new `system_user` row, and leaves `record_end_date` set;
  - a secured route still succeeds when the stored GUID casing differs from the token (validates the `api_set_context` fix);
  - the admin reactivation flow still restores access.
- DB migration: run `npm run migrate`; confirm `system_user_nuk2` is now on `LOWER(user_guid)` and `api_set_context` resolves case-insensitively. `npm run migrate:rollback` restores the case-sensitive index and function.


api/src/services/authorization/authorization-service.ts
```diff
@@ -322,9 +323,25 @@ export class AuthorizationService extends DBService {
...
   }
 
+  /**
+   * Returns `true` if the system user is inactive (soft-deleted), i.e. their `record_end_date` is set to now or in the
+   * past.
+   *
+   * @param {SystemUserExtended} systemUser
+   * @return {*}  {boolean}
+   */
+  static isSystemUserInactive(systemUser: SystemUserExtended): boolean {
```
mauberti-bc: Is the wrapper necessary? Could callers use the shared `isSystemUserInactive` helper directly instead?

mauberti-bc (changes requested):
Nice! 

`getCachedSystemUser()` should also check `record_end_date` before returning a cached or injected user. Otherwise, a soft-deleted user could still be treated as active on authorization paths that use this helper.

```ts
async getCachedSystemUser(): Promise<SystemUserExtended | null> {
  if (this._systemUser) {
    return AuthorizationService.isSystemUserInactive(this._systemUser) ? null : this._systemUser;
  }

  const user = await this.getSystemUserObject();
  this._systemUser = user ?? undefined;

  return user;
}
```

mauberti-bc (approved):

PR #505, opened 2026-07-07
SIMSBIOHUB-1059: Store Taxonomic Hierarchy

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1059](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1059)

## Description of Changes

- Adds `parent_itis_tsn` and `parent_taxon_id` to `taxon`.
- Adds a self-referencing FK from `taxon.parent_taxon_id` to `taxon.taxon_id`.
- Adds indexes for `taxon.parent_itis_tsn` and `taxon.parent_taxon_id`.
- Derives parent-child links from the ordered ITIS Solr `hierarchyTSN` lineage.
- Treats a taxon’s parent as the previous TSN in the ordered ITIS lineage.
- Fetches the full ITIS hierarchy for uncached TSNs.
- Inserts or reuses every taxon in the returned lineage, including ancestors.
- Stores parent links for each lineage taxon where a parent exists.
- Ensures missing taxa and hierarchy exist during submission ingestion before resolving taxon properties.
- Allows valid-but-uncached TSNs to resolve during ingestion instead of failing with `UNRESOLVED_TAXON`.
- Updates taxon search filters (`ChildOf`, `ParentOf`, `AscendsFrom`, `DescendsFrom`) to use indexed `parent_taxon_id` relationships instead of parsing `parentTSN` from `itis_data`.
- *Note*: ITIS does not return hierarchy information for `invalid` taxa, so tarballs that reference invalid taxa _that have never been inserted before_ will intentionally fail. This is deliberate and ensures that invalid taxa can be added only if they previously imported when they were valid.

## Testing Notes

- Tested on a seeded database to simulate the DEV environment.
- Ran the migration and confirmed `taxon` has:
  - `parent_itis_tsn`
  - `parent_taxon_id`
  - `taxon_parent_itis_tsn_idx`
  - `taxon_parent_taxon_id_idx`
- Confirmed existing rows can be linked through `parent_itis_tsn` and `parent_taxon_id`.
- Called `GET /api/taxonomy/taxon?tsn=<tsn>` for a taxon whose ancestors were not cached.
- Confirmed the full lineage is stored locally with parent links set.
- Imported a submission referencing a TSN that was not cached locally.
- Confirmed ingestion resolved and stored the taxon property instead of failing with `UNRESOLVED_TAXON`.
- Checked taxon child, parent, ancestor, and descendant filtering in search.
- Confirmed hierarchy-based search filters still return expected results.


api/src/repositories/submission-feature-property-ingestion-repository.ts
```diff
@@ -863,6 +863,57 @@ export class SubmissionFeaturePropertyIngestionRepository extends BaseRepository
...
+    `;
+
+    const response = await this.connection.sql(
+      sql,
+      z.object({
+        tsn: z.number()
+      })
+    );
+
+    return response.rows.map((row) => row.tsn);
```
mauberti-bc, line 899: nit: avoid Javascript methods like `map` in repository methods. Instead, return `response.rows` or `response.rows[0]` and move data manipulation to the service layer.

mauberti-bc (changes requested):
Very nice using `submission_upload_staging_taxon_candidate`!

database/src/migrations/20260713120000_add_taxon_parent_hierarchy.ts
```diff
@@ -0,0 +1,63 @@
...
+    SET SEARCH_PATH = biohub, public;
+
+    ----------------------------------------------------------------------------------------
+    -- Add rank and parent hierarchy columns
+    ----------------------------------------------------------------------------------------
+    ALTER TABLE taxon ADD COLUMN IF NOT EXISTS rank varchar(50);
+    ALTER TABLE taxon ADD COLUMN IF NOT EXISTS parent_itis_tsn integer;
+    ALTER TABLE taxon ADD COLUMN IF NOT EXISTS parent_taxon_id integer;
+
+    ALTER TABLE taxon ADD CONSTRAINT taxon_parent_taxon_fk FOREIGN KEY (parent_taxon_id) REFERENCES taxon(taxon_id);
```
dylanrogowsky-oxd, line 30: @mauberti-bc Should there be a `DROP ... IF EXISTS` before this, just in case?

```SQL
ALTER TABLE taxon DROP CONSTRAINT IF EXISTS taxon_parent_taxon_fk;
```

mauberti-bc: Shouldn't be necessary since we never had a parent taxon concept before

mauberti-bc (approved):
👍

PR #507, opened 2026-07-07
SIMSBIOHUB-1068: Track Submission Feature Upserts

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1068](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1068)
- [SIMSBIOHUB-1060](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1060) - Dependency

## Description of Changes

- Adds `submission_feature_log`, an append-only table that records when an active feature is superseded during upload reconciliation. Each row links the old feature to its replacement and snapshots both content hashes.
- `removed` is reserved for a future removal workflow — the enum and CHECK support it, but nothing writes it yet.
- New and unchanged features are deliberately not logged; those are already covered by `submission_feature.submission_upload_id` and the reconciliation summary.
- Log rows are written in the reconciliation service right after features are published. A unique index on the previous feature id enforces one transition per version, and the approval aborts if the classified / soft-ended / logged superseded counts don't match.
- The table differs from the ticket's DDL sketch where it made sense to follow existing conventions: enum `action` + `details` jsonb (matching `submission_upload_feature_reconciliation`), varchar(64) hashes, uuid `submission_upload_id`, and the standard audit columns/triggers.

Worth calling out: the log captures transitions as they happen at approval, so the chain tip isn't guaranteed to be the live feature (e.g. denying an already-approved upload rewinds it without touching the log). Current state should still be read from active `submission_feature` rows — the table comment documents this.

## Testing Notes

- No API or response shape changes; the log is internal.
- Unit tests cover the new repository and the service wiring, including the count-divergence guard.
- Integration test (`submission-feature-reconciliation-service.integration.ts`) covers the end-to-end flow: a supersede writes one linked row, new/unchanged/re-approval write nothing, a multi-hop chain resolves to the current feature, and the uk1/ck1 constraints reject invalid rows.

### DIFF
--- a/api/src/__integration__/db/submission-feature-reconciliation-service.integration.ts
+++ b/api/src/__integration__/db/submission-feature-reconciliation-service.integration.ts
@@ -137,14 +137,14 @@ describe('SubmissionFeatureReconciliationService — activation and submission_f
     // Upload A publishes a brand new feature: not a transition, so nothing is logged.
     const versionA = await publishVersion(submissionId, HASH_1);
     expect(versionA.counts).to.eql({ new: 1, unchanged: 0, superseded: 0, conflict: 0 });
-    expect(await getLogRows(submissionId)).to.have.length(0);
+    expect(await getLogRows(submissionId)).to.be.empty;
 
     // Upload B re-submits the same source with changed content: A is superseded by B.
     const versionB = await publishVersion(submissionId, HASH_2);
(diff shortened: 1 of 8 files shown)


dylanrogowsky-oxd:
Redundant with PR-507

PR #508, opened 2026-07-07
SIMSBIOHUB-1085: Remove recursive reachability query when evaluating expression

mauberti-bc:
## Links to Jira Tickets

- https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1085

## Description of Changes

- Updated expression search so results are always limited to the requested feature type.
- Changed predicate matching to use related features as evidence for selecting results, rather than returning every feature connected to that evidence.
- Added direct matching for predicates on the same feature type, so a feature can match on its own properties without needing a self-reference in the closure table.
- No longer uses `submission_feature_feature` (the `content` property of a feature) when finding related features
- Use the closure table only when the predicate is on a different feature type than the one being returned.
- Prevented unrelated sibling records from being returned just because they share a parent, dataset, submission, or other relationship with a matching feature.

## Examples

- Example: searching telemetry by animal species
  - If the user searches for `telemetry` where the related `animal.species = caribou`, the query should return the matching telemetry records.
  - It should not return the animal record itself, the dataset, or unrelated telemetry records from other animals.

- Example: telemetry matching on its own property
  - If the user searches for `telemetry` where `telemetry.temperature > 20`, the telemetry record can match directly on its own property.
  - It does not need a self-referencing row in the closure table to be returned.

- Example: preventing sibling leakage
  - If one animal in a dataset matches the search criteria, the query should not return telemetry for every other animal in that same dataset.
  - Only telemetry records connected to the matching animal should be returned.

- Example: “big fish”
  - If the user searches for `fish` where a related observation or sample has `subcount_comment = big fish`, the query should return the relevant fish features.
  - It should not return every feature reachable from that observation or sample, and it should not pull in unrelated fish just because they share a parent dataset.

- Example: secured matches
  - The normal search returns only records the user can access.
  - The unfiltered check can still detect that additional secured records match the same search, so the UI can say there are hidden secured results.

## Testing Notes

- Search with an expression and confirm every returned result is the requested anchor feature type.
- Confirm results match the expression criteria, including predicates on related feature types.
- Confirm related predicates return only the relevant anchor features, not unrelated siblings from the same parent, dataset, or submission.
- Confirm same-type predicates work without requiring a self-referencing closure row.
- Confirm secured matching results are hidden from the normal result set but still counted for `has_more_secured_features`.

image

### DIFF
--- a/api/src/__integration__/db/expression-evaluation.integration.ts
+++ b/api/src/__integration__/db/expression-evaluation.integration.ts
@@ -46,7 +46,6 @@
 
 import { expect } from 'chai';
 import SQL from 'sql-template-strings';
-import { MAX_SEARCH_GRAPH_DEPTH } from '../../constants/expression';
 import { defaultPoolConfig, getAPIUserDBConnection, IDBConnection, initDBPool } from '../../database/db';
 import {
   NormalizedExpressionTreeExpression,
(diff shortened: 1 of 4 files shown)


dwhogan (approved):
👍

PR #512, opened 2026-07-10
SIMSBIOHUB-1066: Evaluate Taxonomy Filters using Taxonomic Hierarchy

dylanrogowsky-oxd:
## Links to Jira Tickets

- [SIMSBIOHUB-1066](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1066)
- [SIMSBIOHUB-1059](https://apps.nrs.gov.bc.ca/int/jira/browse/SIMSBIOHUB-1059) - Dependency

## Description of Changes

Makes the taxonomy hierarchy expression predicates evaluate over the first-class `taxon.parent_taxon_id` link rather than parsing raw ITIS JSON (`itis_data->>'parentTSN'`). The hierarchy storage, ingestion, and `parent_taxon_id` resolution come from the dependency branch (SIMSBIOHUB-1059); this PR is scoped to the expression-evaluation layer and touches two files:

- **Fix the `ParentOf` operator's generated SQL.** The depth-1 limit was emitted as a second `WHERE` clause (`... WHERE taxon_id = <candidate> WHERE depth = 1`), which Postgres rejects as a syntax error, so `ParentOf` filters failed at runtime. The limit is now appended as `AND depth = 1`, correctly matching only the immediate parent (`target.parent_taxon_id = candidate.taxon_id`). `AscendsFrom` is unaffected (it uses no depth limit).
- **Add focused unit tests** for all four hierarchy operators (`ChildOf`, `ParentOf`, `DescendsFrom`, `AscendsFrom`) in `expression-evaluation.test.ts`. They build each operator's SQL offline (no DB) and assert it walks `parent_taxon_id`, never references `itis_data->>'parentTSN'`, and encodes each operator's relationship: ChildOf = single-hop parent link; ParentOf = depth-limited parent walk; AscendsFrom = full ancestor walk; DescendsFrom = recursive walk up from the candidate. The ParentOf test also guards against the double-`WHERE` regression.

## Testing Notes

- **Unit tests (offline, no DB):** `cd api && npm run test -- --grep applyTaxonExpressionOperator` — 5 passing; the full `expression-evaluation` suite is green (24 passing). `eslint` is clean on both changed files.
- **Integration coverage:** behavioral tests for all four operators over a seeded A→B→C lineage live in `api/src/__integration__/db/expression-evaluation.integration.ts`. Running them requires the `parent_taxon_id` / `parent_itis_tsn` migration from the dependency branch to be applied to the test DB; with this fix the `ParentOf` case goes from failing (syntax error) to passing.
- **Manual DB check:** verified against Postgres that the fixed `ParentOf(child)` returns only the immediate parent, `AscendsFrom(child)` returns the full ancestor chain, and the pre-fix SQL reproduces `ERROR: syntax error at or near "WHERE"`.


api/src/services/expression-predicate-semantic-validator.ts
```diff
@@ -275,6 +278,58 @@ export class ExpressionPredicateSemanticValidator extends DBService {
...
+   * A numeric value is treated as a `taxon_id` directly. A string value is treated as an exact scientific name and
+   * resolved against the local taxon cache (case-insensitive). Resolution assumes taxon ingestion has already inserted
+   * the relevant lineage rows, so an unknown name is rejected rather than fetched from ITIS.
+   *
+   * @param {ExpressionTreePredicate} predicate - Public taxon predicate leaf.
+   * @return {Promise<number>} Resolved taxon id.
+   * @private
+   * @memberof ExpressionPredicateSemanticValidator
+   */
+  private async resolveTaxonValue(predicate: ExpressionTreePredicate): Promise<number> {
```
mauberti-bc, line 293: Suggestion: avoid separate numeric and string paths in `resolveTaxonValue`. Pass the raw `ExpressionTreePredicate.value` to a new `TaxonomyRepository.findTaxon` method using both filters. The repository should match numeric values against `itis_tsn` and otherwise perform an exact, case-insensitive scientific-name lookup.

```ts
private async resolveTaxonValue(predicate: ExpressionTreePredicate): Promise<number> {
  const records = await this.taxonomyRepository.findTaxon({
    itis_tsn: predicate.value,
    itis_scientific_name: predicate.value
  });

  if (records.length !== 1) {
    throw new ApiValidationError(records.length ? 'Taxon is ambiguous' : 'Taxon not found');
  }

  return records[0].taxon_id;
}
```

`findTaxon` should follow existing repository/filter conventions, with `FindTaxonFilters` defined in `models/taxon.ts`:

```ts
type FindTaxonFilters = {
  itis_tsn?: number;
  itis_scientific_name?: string;
};
```

The query should use OR semantics, guard the integer cast, and only return active taxa:

```ts
async findTaxon(filters: FindTaxonFilters): Promise<TaxonRecord[]> {
  const value = filters.itis_tsn ?? filters.itis_scientific_name;

  if (value === undefined || value === null) {
    return [];
  }

  const queryBuilder = getKnex()
    .queryBuilder()
    .select(taxonRecordColumns)
    .from('taxon')
    .whereNull('record_end_date')
    .where((query) => {
      query
        .whereRaw(
          `itis_tsn = CASE
            WHEN ?::text ~ '^[0-9]+$' THEN ?::integer
            ELSE NULL
          END`,
          [value, value]
        )
        .orWhereRaw('LOWER(itis_scientific_name) = LOWER(?::text)', [value]);
    });

  const response = await this.connection.knex(queryBuilder, TaxonRecord);

  return response.rows;
}
```

app/src/utils/expression.ts
```diff
@@ -1216,10 +1217,16 @@ const serializePredicateValue = (draft: BuilderPredicateDraft): unknown => {
     return datetimeObjectToString(draft.value);
   }
 
-  if (draft.type === 'number' || draft.type === 'taxon' || draft.type === 'code') {
+  if (draft.type === 'number' || draft.type === 'code') {
     return Number(draft.value);
   }
 
+  if (draft.type === 'taxon') {
```
mauberti-bc, line 1223: This can just become 
```
if (draft.type === 'taxon') {
  return draft.value;
}
```
after changing `resolveTaxonValue`

api/src/services/expression-predicate-semantic-validator.ts
mauberti-bc, line 293: Consider adding database indexes for the revised `resolveTaxonValue` lookup via `findTaxon`. The existing active `itis_tsn` index covers TSN equality, but `LOWER(itis_scientific_name) = LOWER(?)` cannot use the current case-sensitive scientific-name index. Add a partial functional index on `LOWER(itis_scientific_name)` where `record_end_date IS NULL`?

api/src/services/expression-predicate-semantic-validator.ts
```diff
@@ -187,7 +190,7 @@ export class ExpressionPredicateSemanticValidator extends DBService {
       case FEATURE_PROPERTY_TYPE.DATETIME:
         return this.buildTimestampPredicate(predicate);
       case FEATURE_PROPERTY_TYPE.TAXON:
-        return { type: 'taxon', operator: predicate.operator, value: this.requirePositiveIntegerValue(predicate) };
+        return { type: 'taxon', operator: predicate.operator, value: await this.resolveTaxonValue(predicate) };
```
mauberti-bc, line 194: Note for clarification: The client-facing `ExpressionTreePredicate.value` should remain the submitted TSN or scientific name. `resolveTaxonValue` converts that value into a `taxon_id`, which belongs only in `NormalizedExpressionTreePredicate.internal_predicate.value`. Both identifiers are numbers, so the distinction is structural rather than enforced by TypeScript: public `value` is a TSN, while internal taxon `value` is a `taxon_id`.

api/src/services/expression-predicate-semantic-validator.ts
dylanrogowsky-oxd: Done. `resolveTaxonValue` passes the raw predicate value straight to `findTaxon` and returns the single match, with `FindTaxonFilters` in `models/taxon.ts`.

One deviation from the snippet: the TSN cast is `?::text::bigint`, not `?::integer`. `node-pg` sends bound params untyped, so with `?::integer` Postgres infers the param as numeric and fails at bind time on a scientific name (`invalid input syntax for type integer: "Cervidae"`) before the `CASE` guard runs. Casting through `::text` first keeps it text-typed so the guard evaluates. `::bigint` so an all-digit value larger than `int4` returns not-found instead of erroring.

app/src/utils/expression.ts
dylanrogowsky-oxd: Done - it's `return draft.value;` now. Comment updated to say the numeric value is a TSN, not a `taxon_id`.

api/src/services/expression-predicate-semantic-validator.ts
dylanrogowsky-oxd: Added in `20260714120000_add_taxon_scientific_name_lower_index.ts` - partial functional index `taxon_scientific_name_lower_idx` on `LOWER(itis_scientific_name)` where `record_end_date IS NULL`. The TSN branch is already covered by `taxon_nuk3`.

api/src/repositories/taxonomy-repository.ts
```diff
@@ -52,6 +58,51 @@ export class TaxonomyRepository extends BaseRepository {
...
+      .whereNull('record_end_date')
+      .where((query) => {
+        query
+          // The value is cast through ::text first so the bound parameter is inferred as text: a scientific-name
+          // string (e.g. 'Cervidae') would otherwise be bound directly to a numeric parameter and fail at bind time
+          // ("invalid input syntax for type bigint"), before the CASE guard runs. ::bigint (not ::integer) then avoids
+          // an int4 overflow for an all-digit value larger than int4 — it simply never equals an int4 itis_tsn.
+          .whereRaw(
+            `itis_tsn = CASE
+              WHEN ?::text ~ '^[0-9]+$' THEN ?::text::bigint
```
mauberti-bc: I don't think we should see `bigint` anywhere, because `taxon.itis_tsn` is a regular 6-digit integer? It should be typed as an integer in the database.

